Issue #6, opened 2016-12-12
Add reflection to types to know when they are generic


krzysztofzablocki:
We need to add ability to recognize generic types, as often we want to avoid those when doing code generation. 

E.g. we'll want to skip `allValues` generation for Enums that are generic.

krzysztofzablocki:
@ilyapuchka I started working on this and wondering whether we should treat the generic part as part of type name or introduce another variable for that component? e.g. `Foo<Element: Equatable>` would you put whole thing as name or just Foo and then add another `var for Element: Equatable`?

Or maybe for now we should just limit generic support to being able to check whether type `isGeneric`, and that will suffice for most needs?

ilyapuchka:
I would say that for now we can have just `isGeneric`. Because how should we handle (if we should at all) generics with complex constraints with `where` and so one?

krzysztofzablocki:
yeah that's my thoughts, I'm not even sure if code-generation in general should even go so far as to try to deal with this, it feels too error prone, I'm going to add `isGeneric` tomorrow so that people can skip those classes ;)

PR #9, opened 2016-12-12
Improve Eqaulity.stencil.swift

ilyapuchka:
For now I just fixed some typos in current version of template to make it work on Insanity own model types.

But I'd like to improve that template to make it more generic:

1. We may add `if lhs === rhs { return true }` in the second line of the `isEqual` for classes.
2. For structs instead of generating `isEqual` template should generate `Equitable` conformance.
3. For enums with associated values we may generate `Equatable` conformance based on switch like the following.

```swift
swift (lhs, rhs):
case let (.{{ case }}(lhsValue), .{{ case}}.(rhsValue)): return lhsValue == rhsValue
```
But that will probably require some additional checks (like in #6 for instance) and maybe a separate issue/pr will be better.


ilyapuchka:
Aslo having https://github.com/kylef/Stencil/pull/32 would be cool. But in meantime maybe we should  patch Stencil the same way as @AliSoftware patched it in SwiftGen?

krzysztofzablocki:
This template is for Insanity itself, We can add a generic one but might be worth doing that in top root folder under `ExampleTemplates` ?

In this pr you just tweaked spaces right? I used this template to code generate the boilerplate code for the tool so it worked, I think the spaces were inserted either by Xcode or I've to correct Swiftlint ignore entry

ilyapuchka:
Yes, there were few problems with spaces and new lines. With that I was able to generate the same code as what is currently checked in.

Regarding where to create a more generic template I was going to create a separate issue asking about your plans on how to distribute custom templates. This one used internally by the tool can be different and "private", but I guess there will be more useful templates coming and they will drive new features in generator. Let's discuss that there =)

PR #11, opened 2016-12-12
Add header with version to generated output file

nubbel:
Thank you for this awesome tool! 🎉
This PR adds a header to each generated file that includes the version of Insanity used to generate it.

Let me know what you think about it.


krzysztofzablocki:
I like the idea, can we tweak the header to the following?
```
// Generated using Insanity \(version) — https://github.com/krzysztofzablocki/Insanity
// DO NOT EDIT
```

nubbel:
@krzysztofzablocki Thanks for the quick response!
Of course, will do so now.

nubbel:
Done!

Btw. the rationale behind that change is to make it easier to reproduce the output, which does not only depend on the input and template file, but also on the version of Insanity.
Additionally the warning in the second line should prevent unintentional edits to the generated files.
Should have stated that earlier…

krzysztofzablocki:
Absolutely, it's a good idea and thank you very much for contributing 👍  
I'll merge this in the morning once CI passes (after midnight here)

nubbel:
Yep, same here 🇩🇪. 
Good night! 🌙😴💤

PR #12, opened 2016-12-12
Implemented based, inheriting and implementing types collections

### LINKED ISSUE #8: Inherits vs Implements
Currently `implements` is just an alias for `inherits`. But wouldn't it align more with language concepts of implementing some protocol and inheriting from some type if `inherits` will only contain inherited type (which can be only one and only for classes) and `implements` will contain all implemented protocols? Or is there any particular reason why you decided not to distinguish these types? 

Also there must be cases for templates when it's needed to distinguish inherited type and implemented protocols, though I can not come up with an example right now.

The second thing is that maybe it will make sense not to put in `inherited` types a raw type name for enums, as it is not a case of inheritance per se. And there is already `rawType` for that.

What do you think @krzysztofzablocki ? Will be glad to make a pr for that.

ilyapuchka:
Resolves #8

### DIFF
--- a/Insanity/Generating/Generator.swift
+++ b/Insanity/Generating/Generator.swift
@@ -11,8 +11,9 @@ import Stencil
 /// types.structs <- lists all structs
 /// types.enums <- lists all enums
 /// types.protocols <- lists all protocols (that were defined in the project)
-/// types.inheriting.BaseClassOrProtocol <- lists all types inheriting from given BaseClass or implementing given Protocol
-/// types.implementing.BaseClassOrProtocol <- convience alias for inheriting ^
+/// types.inheriting.BaseClass <- lists all classes inheriting from known BaseClass
+/// types.implementing.BaseProtocol <- lists all types implementing known BaseProtocol
+/// types.based.BaseClassOrProtocol <- lists all types inheriting from given BaseClass or implementing given Protocol
 private class TypesReflectionBox: NSObject {
     private let types: [Type]
 
@@ -44,7 +45,7 @@ private class TypesReflectionBox: NSObject {
         return self.types.filter { $0 is Enum }
     }()
 
-    lazy var inheriting: [String: [Type]] = {
+    lazy var based: [String: [Type]] = {
         var content = [String: [Type]]()
         self.all.forEach { type in
             type.inheritedTypes.forEach { name in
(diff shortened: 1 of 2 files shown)


Insanity/Generating/Generator.swift
```diff
@@ -44,10 +45,28 @@ private class TypesReflectionBox: NSObject {
...
+        }
+        return content
+    }()
+
     lazy var inheriting: [String: [Type]] = {
+        let classesNames = self.classes.map({ $0.name })
+        
         var content = [String: [Type]]()
         self.all.forEach { type in
+            guard classesNames.contains(type.name) else { return }
```
krzysztofzablocki: if you use
```swift
self.classes.forEach
```
then you can get rid of this guard

InsanityTests/GeneratorSpec.swift
```diff
@@ -46,11 +47,27 @@ class GeneratorSpec: QuickSpec {
...
+                expect(generate("Found {% ifnot types.implementing.Foo %}0{% endif %} types")).to(equal("Found 0 types"))
+                expect(generate("Found {% ifnot types.implementing.NSObject %}0{% endif %} types")).to(equal("Found 0 types"))
+                expect(generate("Found {% ifnot types.implementing.Bar %}0{% endif %} types")).to(equal("Found 0 types"))
             }
 
-            it("feeds types.inheriting specific protocol") {
-                expect(generate("Found {{ types.inheriting.Decodable.count }} types")).to(equal("Found 1 types"))
+            it("feeds types.inheriting specific class") {
+                expect(generate("Found {{ types.inheriting.Foo.count }} types")).to(equal("Found 1 types"))
+                expect(generate("Found {% ifnot types.inheriting.NSObject.count %}0{% endif %} types")).to(equal("Found 0 types"))
```
krzysztofzablocki: this ifnot should not use `.count` but rather the NSObject for consistency with other checks

InsanityTests/GeneratorSpec.swift
```diff
@@ -46,11 +47,27 @@ class GeneratorSpec: QuickSpec {
...
+                expect(generate("Found {% ifnot types.inheriting.Bar %}0{% endif %} types")).to(equal("Found 0 types"))
+                expect(generate("Found {% ifnot types.inheriting.KnownProtocol %}0{% endif %} types")).to(equal("Found 0 types"))
+                expect(generate("Found {% ifnot types.inheriting.Decodable %}0{% endif %} types")).to(equal("Found 0 types"))
+            }
+
+            it("feeds types.based specific type or protocol") {
+                expect(generate("Found {{ types.based.Foo.count }} types")).to(equal("Found 1 types"))
```
krzysztofzablocki: could you re-order those 3 containers to have the same the order of types?, that way it's easier to read through

krzysztofzablocki:
Awesome work, thank you 👍

PR #17, opened 2016-12-13
Fixed description template used internally and updated generated code

ilyapuchka:
The same as #9 but this pr fixes description template and also I updated generated code with the latest version of Sourcery.


krzysztofzablocki:
👍 thanks

PR #21, opened 2016-12-14
Improve logic around enum rawType

### LINKED ISSUE #18: Correctly detecting enum rawValue type
Currently enums assume that the first type after `:` is a rawValue type. But that will fail in a couple of scenarios. There are several different ways to define valid enum:

1. Enum explicitly implementing `RawRepresentable`:

```swift
enum MyEnum: RawRepresentable { //not necessary first or present directly
  typealias RawValue: Int //not necessary present
  var rawValue: Int { ... } //type can be RawValue
  init?(rawValue: Int) { ... } //type can be RawValue
}
```

2. Implicit rawValue enum

```swift
enum StringEnum: String {
  case a, b, c
}

enum NumericEnum: Int { //or any other numeric type
  case one, two, three
}
```

3. Enum with no rawValue at all

```swift
enum JustEnum {
 case a, b, c
}
```

I can suggest the following logic:

1. if there are no inherited types - enum does not have raw value
2. if there are inherited types:
  2.a if there is `RawRepresentable` at any position in inherited types:
    2.a.i check type of rawValue property, or
    2.a.ii check type of rawValue constructor argument, or
    2.a.iii fail otherwise (as SourceKitten does not provide info about typealiases) or scan type body content for `typealias`
  2.b if there is no `RawReperesentable` in inherited types get the first type as rawValue type if there is any

There are of course more scenarios that will still not work (like enum that implements protocol derived from RawRepresentable), but we can always improve in the future.

krzysztofzablocki:
I refactored some general structure to ease future parsing a bit, take a look @ilyapuchka
Since this is built on top of your PR, it closes #19

Thank you for contributing, and let me know if I missed something or you disagree with anything I did.

Fixes #18

### DIFF
--- a/Sourcery/Parsing/Parser.swift
+++ b/Sourcery/Parsing/Parser.swift
@@ -148,13 +148,11 @@ final class Parser {
                 containingType.staticVariables += [variable]
             } else {
                 containingType.variables += [variable]
-            }
 
-        if let enumeration = containingType as? Enum {
-            if let updatedRawType = parseEnumRawType(enumeration: enumeration, from: variable) {
-                enumeration.rawType = updatedRawType
+                if let enumeration = containingType as? Enum, let updatedRawType = parseEnumRawType(enumeration: enumeration, from: variable) {
+                    enumeration.rawType = updatedRawType
+                }
             }
-        }
 
         case let (_, childType as Type):
             containingType.containedTypes += [childType]
--- a/Sourcery/Parsing/Parser.swift
+++ b/Sourcery/Parsing/Parser.swift
@@ -68,9 +68,11 @@ final class Parser {
             case .extensionStruct:
                 type = Struct(name: name, accessLevel: access, isExtension: true, inheritedTypes: inheritedTypes)
             case .enum:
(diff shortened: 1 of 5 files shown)


Sourcery/Parsing/Parser.swift
```diff
@@ -129,6 +149,13 @@ final class Parser {
             } else {
                 containingType.variables += [variable]
             }
+
+        if let enumeration = containingType as? Enum {
```
ilyapuchka: If I understand correctly `processContainedType` will be called for any variable defined inside the type, so we might do the same work again and again. My initial intention was also to put it here, but then I moved it. I think at least this block can be moved to `else` brunch of previous `if` as we know for sure that `rawValue` is not a static variable.
Apart from that looks good! 👍

krzysztofzablocki:
I do think it could be useful to have a callback when a type has finished building and all it's internals are set, we can move that variable logic there once it's added, I'll be refactoring our test suite probably on Saturday to simplify constructing test data as I've pretty nice pattern for that :)

krzysztofzablocki:
Travis queue times are so looong :)

PR #25, opened 2016-12-15
Added kind property to Type

ilyapuchka:
Not sure about how useful this might be but it might be just an alternative to iterating separately `type.classes`, `types.structs`,`types.enums`.


krzysztofzablocki:
Might be useful, Could you add it to readme listing?

krzysztofzablocki:
Actually no worries, I'll add it since I need to update readme for annotation docs anyway ;)

PR #26, opened 2016-12-15
Variables types reflection

ilyapuchka:


Sourcery/Generating/Generator.swift
```diff
@@ -19,6 +19,13 @@ private class TypesReflectionBox: NSObject {
 
     init(types: [Type]) {
         self.types = types
+        
+        for type in types {
```
krzysztofzablocki: pass in `typesByName` then you can get rid of inner loop for filtering types and it will be faster :)

ilyapuchka:
@krzysztofzablocki I think I'm pretty close here, can you please review?

Sourcery/Parsing/Parser.swift
```diff
@@ -95,14 +95,37 @@ final class Parser {
...
+                guard let keyword = contents.bridge().substringWithByteRange(start: token.offset, length: token.length) else { return false }
+                return keyword == "typealias"
+            }
+            .map({ (index, _) in (tokens[index + 1], tokens[index + 2]) })
+            .forEach({ (aliasToken, typeToken) in
+                guard
+                    let alias = contents.bridge().substringWithByteRange(start: aliasToken.offset, length: aliasToken.length),
+                    let type = contents.bridge().substringWithByteRange(start: typeToken.offset, length: typeToken.length) else { return }
+                
+                typealiases[alias] = type
```
krzysztofzablocki: What will happen if we have private type aliases? those wouldn't be accessible to anything we'd ever generate

Sourcery/Parsing/Parser.swift
```diff
@@ -207,28 +230,42 @@ final class Parser {
...
 
         types
             .filter { $0.isExtension == false }
             .forEach { unique[$0.name] = $0 }
 
         types.forEach { type in
             guard let current = unique[type.name] else {
-                let inheritanceClause = type.inheritedTypes.isEmpty ? "" : ": \(type.inheritedTypes)"
-                if verbose { print("\(logPrefix)Ignoring \"extension \(type.name)\(inheritanceClause)\" because we don't have original type definition information") }
+                unique[type.name] = type
```
krzysztofzablocki: when we enter this guard expression, that means that we are an extension of a type that we didn't scan. 

Curious on what's the use-case for having the extension meta-info available here? 

(Also I think we should leave warning, just modify wording so people are aware of the limits if they use verbose logging)

SourceryTests/Models/VariableSpec.swift
```diff
@@ -25,6 +25,16 @@ class VariableSpec: QuickSpec {
 
             it("reports optional false") {
                 expect(sut?.isOptional).to(beFalse())
+                sut?.typeName = "Int?"
```
krzysztofzablocki: it's better to use context and separate tests scenarios as it makes understanding regressions and code easier. So here I'd change this into:
```swift
context("given an optional type") {
  it("recognizes optional) {
    expect(sut?.isOptional).to(beFalse())
  }
  it("returns non-optional type for unwrappedTypeName") {
    expect(sut?.unwrappedTypeName).to(equal("Int"))
  }
}
```

and another context for non-optional type

SourceryTests/ParserSpec.swift
```diff
@@ -178,6 +185,41 @@ class ParserSpec: QuickSpec {
...
+                context("given typealias") {
+                    it("extracts typealiases properly") {
+                        expect(sut?.parseContents("typealias FooAlias = Foo; class Foo {}", existingTypes: []).1)
+                            .to(equal(
+                                ["FooAlias": "Foo"]
+                            ))
+                    }
+                    
+                    it("replaces variable alias type with actual type") {
+                        let expectedVariable = Variable(name: "foo", type: "FooAlias")
```
krzysztofzablocki: it's useful to use AAA separation in testing http://wiki.c2.com/?ArrangeActAssert, it improved readability

Sourcery/Parsing/Parser.swift
krzysztofzablocki: Are you not worried that this might cause inconsistencies?

ilyapuchka: I don't know to be honest. It all depends on use cases for templates. I'll revert it for now and if there will be such feature request we can put it back.

krzysztofzablocki: Sounds reasonable and this is an easy thing to add back if we find use cases for it, implementation is looking solid :)

Sourcery/Parsing/Parser.swift
ilyapuchka: For now type aliases will not be accessible in templates, I simply use them here to find out the correct type information for variables and to extend proper class in case there is an extension on type alias.

krzysztofzablocki: Ok makes sense

ilyapuchka:
@krzysztofzablocki I addressed you review comments, feels like it's ready to be merged

ilyapuchka:
I also adjusted description and equality templates, but if you want I can make a separate pr for that

krzysztofzablocki:
Looks good, thanks 👍

PR #29, opened 2016-12-15
Shell escape path names

bejar37:
This was failing for me trying to build on my Jenkins machine, when the path has a space in it. This fixes that.


krzysztofzablocki:
:+1: thank you for contributing!

PR #36, opened 2016-12-16
Enabled extensions of unknown types

### LINKED ISSUE #35: `types.implementing` with an extension?
Hi,
I'm using a phantom protocol to guide my code generation, and I noticed something interesting.

I have a template that looks like this:

```
{% for types.implementing.MyProtocol %}
// Generate some stuff here...
{% endfor %}
```

This works fine when I declare conformance like this:
```swift
struct IntBag: MyProtocol {
   let int: Int
}
```

but not like this:
```swift
struct IntBag {
   let int: Int
}

extension IntBag: MyProtocol { }
```

Is there a particular reason why the second example isn't picked up?

ilyapuchka:
Resolves #35

### DIFF
--- a/Sourcery/Generating/Generator.swift
+++ b/Sourcery/Generating/Generator.swift
@@ -25,7 +25,7 @@ private class TypesReflectionBox: NSObject {
     lazy var classes: [Type] = {
         return self.types.filter { type in
             let isNotClass = type is Struct || type is Enum || type is Protocol
-            return !isNotClass
+            return !isNotClass && !type.isExtension
         }
     }()
 
--- a/Sourcery/Models/Type.swift
+++ b/Sourcery/Models/Type.swift
@@ -9,7 +9,7 @@ import Foundation
 class Type: NSObject {
     internal var isExtension: Bool
 
-    var kind: String { return "class" }
+    var kind: String { return isExtension ? "extension" : "class" }
     var accessLevel: AccessLevel
 
     /// Name in global scope 
--- a/Sourcery/Parsing/Parser.swift
+++ b/Sourcery/Parsing/Parser.swift
@@ -243,8 +243,9 @@ final class Parser {
 
         types.forEach { type in
             guard let current = unique[type.name] else {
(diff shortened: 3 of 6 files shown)


Sourcery/Generating/Generator.swift
```diff
@@ -25,7 +25,7 @@ private class TypesReflectionBox: NSObject {
     lazy var classes: [Type] = {
         return self.types.filter { type in
             let isNotClass = type is Struct || type is Enum || type is Protocol
-            return !isNotClass
+            return !isNotClass && !type.isExtension
```
krzysztofzablocki: what's the logic around this check? Wouldn't this cause inconsistency with other kind reflections? rest of changes looks good

ilyapuchka: My reasoning for that is that we don't know what was extended, was it a objective-c class or 3rd party struct. So we can not say for sure that that extension belongs to classes collection. But it will be still accessible through `types.all` collection. And with `kind` property it will be  possible to distinguish different kinds of types if needed. 
Btw maybe we also need to return "extension" instead of "class" in `kind` property, what do you think?

krzysztofzablocki: ah right I was thinking this won't be consistent but we have fake types for all other kind of entries, so this is the only one we need to filter :+1: 

It will also be accessible from `types.based.ProtocolName` right? 
also :+1: on adding `extension` kind

ilyapuchka: Yes, `types.based` will also work, they only be filtered form `types.inheriting` which makes sense as we can only add protocol conformance with extensions.

krzysztofzablocki: @ilyapuchka while you are at it, can you add 3 simple generator tests to prove that scenario for accessing this kind of data is working as expected, and so we can avoid regressions? I'll merge this and release a new version once that's done

ilyapuchka: I've already added extension type instance in generator test types and updated tests for `type.based` and `type.implements`. Or do you mean some kind of other tests?

krzysztofzablocki: ah I think I missed it, merging, thanks!

PR #37, opened 2016-12-16
fixed detecting enum rawType and inherited types

### LINKED ISSUE #34: Can't see enums implementing a protocol?
I'm trying to write a template to deserialize enums, and trying to generate code for all enums that implement a phantom protocol, `AutoEnumDecodable`. However, it seems like `types.implementing.Protocol` doesn't work for enum types?
```
{% for type in types.implementing.AutoEnumDecodable %}
extension {{type.name}}: JsonDecodable {
    static func parse(decoder: CupertinoDecoder) {
        //...
    }
}
{% endfor %}
```
Enums that implement `AutoEnumEncodable` aren't generating anything.
BTW, I'm really enjoying using this library so far!

ilyapuchka:
Fixes #34

### DIFF
--- a/SourceryTests/ParserSpec.swift
+++ b/SourceryTests/ParserSpec.swift
@@ -249,41 +249,58 @@ class ParserSpec: QuickSpec {
                                         Enum(name: "Foo", accessLevel: .internal, isExtension: false, inheritedTypes: [], cases: [], variables: [Variable(name: "x", type: "Int", accessLevel: (.internal, .none), isComputed: true)])
                                 ]))
                     }
+                    
+                    context("given enum without rawType") {
+                        it("extracts inherited types properly") {
+                            expect(parse("enum Foo: SomeProtocol { case optionA }; protocol SomeProtocol {}"))
+                                .to(equal([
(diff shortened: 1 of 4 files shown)


SourceryTests/ParserSpec.swift
```diff
@@ -249,41 +249,58 @@ class ParserSpec: QuickSpec {
...
+                    }
 
                     context("given enum containing rawType") {
 
                         it("extracts enums without RawRepresentable") {
-                            let expectedEnum = Enum(name: "Foo", accessLevel: .internal, isExtension: false, inheritedTypes: [], cases: [Enum.Case(name: "optionA")])
-                            expectedEnum.rawType = "String"
-
-                            expect(parse("enum Foo: String { case optionA }")).to(equal([expectedEnum]))
+                            expect(parse("enum Foo: String, SomeProtocol { case optionA }; protocol SomeProtocol {}"))
```
krzysztofzablocki: nothing to change here but when doing tdd/bdd and refactoring code it's more reliable to do so while going from green to green and changing tests after (in separate commit) because if public interface didn't change then tests aren't in need of changes and it's making reasoning about changes not breaking anything easier.

krzysztofzablocki:
thanks :)

PR #39, opened 2016-12-17
Add isGeneric support

### LINKED ISSUE #6: Add reflection to types to know when they are generic

krzysztofzablocki:
@ilyapuchka Since you have been so helpful, I've invited you to the project as a collaborator, hope that's ok :) 
Check this PR out

Closes #6

### DIFF
--- a/Sourcery/Models/Type.swift
+++ b/Sourcery/Models/Type.swift
@@ -18,10 +18,8 @@ class Type: NSObject {
         return "\(parentName).\(localName)"
     }
 
-    /// Is this type generic? regardless whether it's this type directly or one of its parent types
-    var isGeneric: Bool {
-        return (parent?.isGeneric ?? false) || hasGenericComponent
-    }
+    /// Is this type generic?
+    var isGeneric: Bool
 
     /// Name in parent scope
     var localName: String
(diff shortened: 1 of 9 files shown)


Sourcery/Models/Type.swift
```diff
@@ -14,10 +14,15 @@ class Type: NSObject {
...
     /// Name in global scope 
     var name: String {
-        guard let parentName = parentName else { return localName }
+        guard let parentName = parent?.name else { return localName }
         return "\(parentName).\(localName)"
     }
 
+    /// Is this type generic? regardless whether it's this type directly or one of its parent types
+    var isGeneric: Bool {
+        return (parent?.isGeneric ?? false) || hasGenericComponent
```
krzysztofzablocki: Before Swift 3 they couldn't, now they can

ilyapuchka: Really? 😯 I got a compilation error with this code:
```
struct MyStruct<Element> {
    struct Inner { } //Type 'Inner' can not be nested in generic type 'MyStruct'
}
```

krzysztofzablocki: Do you think we should remove this check to improve performance, or doesn't matter?

krzysztofzablocki: ok, I'll get rid of it for now 👍

PR #44, opened 2016-12-17
Fixed extending type with extensions that contains static properties

ilyapuchka:
Instead of keeping `staticVariables` as a separate array I think it makes sense to work with them the same way as with computed/stored variables, just filtering by `isStatic` flag. Then it fixes the issue automatically instead of needing to do `self.staticVariables += type.staticVariables`


Sourcery/Models/Type.swift
```diff
@@ -24,23 +24,25 @@ class Type: NSObject {
...
 
+    /// All type static variables
+    var staticVariables: [Variable] {
+        return variables.filter({ $0.isStatic })
+    }
+
     /// Only computed instance variables
     var computedVariables: [Variable] {
-        return variables.filter { $0.isComputed }
+        return variables.filter { $0.isComputed && !$0.isStatic }
```
ilyapuchka: Something not related to this pr. I was wondering why do we need a stored `isComputed` flag? Can't we just check for write access to be not `none`? It looks like if the property is not computed SourceKitten always returns setter access level, it does not return it only for computed properties.

krzysztofzablocki: We need to, my first version did that but that's a simplification that's not going to work in more advanced cases eg. you can have let that doesn't have write, you have bodies that don't necessarily mean computed like didSet etc

Issue #42, opened 2016-12-17
Add `implements` and `inherits` properties to type


ilyapuchka:
As now we have a separation between `implementing` and `inheriting` types collections in generator we may as well add corresponding properties to `Type` model.
Then `inheritedTypes` (should be renamed to `based` probably?) becomes analogous to `types.based`, `implements` to `types.implementing` and `inherits` to `types.inheriting`.

With that instead of `variable.type.inheritedTypes.MyProtocol` it will be possible to do `variable.type.implements.MyProtocol` which just reads better. Also it will be more consistent with mentioned separation of inheriting and implementing types in generator.

krzysztofzablocki:
I'll add it 👍

krzysztofzablocki:
This will be needed for the following to work properly:
```swift
protocol AutoEquatable {}

struct AStruct: Equatable {
  public static func == (l: AStruct, r: AStruct) -> Bool {
    return true
  }
}

struct Book: AutoEquatable {
  let aStruct: AStruct
}
```

Given template like
```swift
{% for type in types.implementing.AutoEquatable %}
  {% for variable in type.storedVariables %}
    {% if variable.type.inheritedTypes.AutoEquatable %}
      {{ variable.type }}
    {% endif %}
  {% endfor %}
{% endfor %}
```

It should not generate code

ilyapuchka:
Ah, so if `inheritedTypes` is just an array of strings we can not use it in `if type.inheritedTypes.SomeType`? Then looks like those collections should be "dummy" dictionaries like `["TypeName": "TypeName"]`?

krzysztofzablocki:
well right now we don't store real types in inheritedTypes, and we should as we want to be able to query they implement something so I'll add it

ilyapuchka:
We may store them like `[TypeName: Type]`, but I'm not sure we will get something from storing actual type references, because what we need is just to check if the type name is inside the implementing types collection and not to access the metadata of those types. At least I don't see it useful right now.

krzysztofzablocki:
Implemneted in 0.4.1

Issue #41, opened 2016-12-17
Type `inheritedTypes` should flatten types hierarchy


ilyapuchka:
Currently we only put into `inheritedTypes` types that we directly inherit from or implement. Shouldn't we flatten types hierarchy so that if type that implements a protocol (or inherits from class) that inherits from another protocol we can check in the template if type implements both of them?

For example with such code:
```swift
protocol AutoEquatable: Equatable {}
class BaseClass: AutoEquatable {}
class SubClass: BaseClass {}
```

in the template `{% types.implementing.Equatable %}`will not give us any type when it should give us both `BaseClass` and `SubClass`. `{% types.implementing.AutoEquatable %}` will only give us `BaseClass`, not `SubClass`.

Implementation should be trivial I think.

krzysztofzablocki:
:+1: I'll modify it

krzysztofzablocki:
Implemented in 0.4.1, @bolismauro Can you give it a try?

PR #46, opened 2016-12-17
Type implements, inherits and flat mapping 

krzysztofzablocki:
This should solve #42 and #41

### DIFF
--- a/Sourcery/Generating/Generator.swift
+++ b/Sourcery/Generating/Generator.swift
@@ -105,6 +105,20 @@ enum Generator {
         var typesByName = [String: Type]()
         types.forEach { typesByName[$0.name] = $0 }
 
+        types.forEach { type in
+            type.based.keys.forEach { name in
+                switch typesByName[name]?.kind {
+                case "class"?:
+                    type.inherits[name] = name
+                case "protocol"?:
+                    type.implements[name] = name
+                default:
+                    break
+                }
+            }
+
+        }
+
         let context: [String: Any]? = [
             "types": TypesReflectionBox(types: types),
             "type": typesByName
--- a/Sourcery/Generating/Generator.swift
+++ b/Sourcery/Generating/Generator.swift
@@ -48,7 +48,7 @@ private class TypesReflectionBox: NSObject {
     lazy var based: [String: [Type]] = {
         var content = [String: [Type]]()
         self.all.forEach { type in
-            type.inheritedTypes.forEach { name in
+            type.based.keys.forEach { name in
                 var list = content[name] ?? [Type]()
                 list.append(type)
                 content[name] = list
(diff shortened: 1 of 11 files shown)


Sourcery/Generating/Generator.swift
```diff
@@ -101,10 +93,39 @@ extension Type {
 }
 
 enum Generator {
+    private static func updateTypeRelationship(for type: Type, typesByName: [String: Type], processed: inout [String: Bool]) {
+        type.based.keys.forEach { name in
+            guard let baseType = typesByName[name] else { return }
```
ilyapuchka: I guess with that such test case in generator tests will fail:
`expect(generate("{% if type.Bar.based.NSObject %} TRUE {% endif %}")).to(equal(" TRUE "))`
as `NSObject` is not known type so is not in `typesByName`

Sourcery/Generating/Generator.swift
```diff
@@ -101,10 +93,39 @@ extension Type {
...
+            if processed[name] != true {
+                processed[name] = true
+                updateTypeRelationship(for: baseType, typesByName: typesByName, processed: &processed)
+            }
+
+            baseType.based.keys.forEach {  type.based[$0] = $0 }
+
+            switch baseType.kind {
+            case "class":
```
ilyapuchka: I would maybe use `is Protocol` and others instead of relying on strings

Sourcery/Generating/Generator.swift
ilyapuchka: I would say to make it work it should be`guard let baseType = typesByName[name] else { type.based[name] = name; return }`

krzysztofzablocki: I think that's configured already, it's iteration over based after all so that's a no op?

Sourcery/Generating/Generator.swift
krzysztofzablocki: That's is a good point, I'll update tomorrow

Sourcery/Generating/Generator.swift
ilyapuchka: I just run that test case - it fails even with the fix I proposed.

ilyapuchka: Ah, I tested against `inherits`, with `base` it works. I think it's good to merge 👍

SourceryTests/GeneratorSpec.swift
```diff
@@ -108,6 +110,31 @@ class GeneratorSpec: QuickSpec {
                 it("can access variable type information") {
                     expect(generate("{% for variable in type.Complex.variables %}{{ variable.type.name }}{% endfor %}")).to(equal("FooBar"))
                 }
+
+                it("generates proper response for type.inherits") {
+                    expect(generate("{% if type.Bar.inherits.ProjectClass %} TRUE {% endif %}")).to(equal(" TRUE "))
+                    expect(generate("{% if type.Bar.inherits.Decodable %} TRUE {% endif %}")).toNot(equal(" TRUE "))
```
ilyapuchka: Shouldn't `Decodable` be present only in `based`? It's unknown type, right?

krzysztofzablocki:
Merged by hand

PR #50, opened 2016-12-17
Properly detect local and global typealiases

ilyapuchka:


Sourcery/Parsing/Parser.swift
```diff
@@ -453,18 +456,57 @@ extension Parser {
...
 
     fileprivate func isGeneric(source: [String: SourceKitRepresentable]) -> Bool {
-        guard let substring = extract(.nameSuffix, from: source), substring.trimmingCharacters(in: .whitespacesAndNewlines).hasPrefix("<") == true else { return false }
+        guard let substring = extract(.nameSuffix, from: source, contents: contents), substring.trimmingCharacters(in: .whitespacesAndNewlines).hasPrefix("<") == true else { return false }
         return true
     }
 
-    fileprivate func extract(_ substringIdentifier: SubstringIdentifier, from source: [String: SourceKitRepresentable]) -> String? {
-        let substring = substringIdentifier.range(for: source).flatMap { self.contents.substringWithByteRange(start: Int($0.offset), length: Int($0.length)) }
+    fileprivate func extract(_ substringIdentifier: SubstringIdentifier, from source: [String: SourceKitRepresentable], contents: String) -> String? {
```
krzysztofzablocki: why always pass contents in? add a convenience variant that uses this function but passes in self.contents. That way you only use a subset when explicitly needed

Sourcery/Parsing/Parser.swift
```diff
@@ -453,18 +456,57 @@ extension Parser {
 
         return annotations
     }
+    
+    fileprivate func parseTypealiases(from source: [String: SourceKitRepresentable], containingType: Type?, processed: [[String: SourceKitRepresentable]]) -> [Typealias] {
+        var contents = self.contents
+        
+        //replace all processed substructures with whitespaces so that we don't process their typealiases again
+        for substructure in processed {
+            if let offset = substructure[SwiftDocKey.offset.rawValue] as? Int64,
```
krzysztofzablocki:
```swift
if let data = SubstringIdentifier.key.range(for: source).flatMap({ (substring: contents.substringWithByteRange(start: Int($0.offset), length: Int($0.length)), range: $0) }) {
  let replacement = String(repeating: " ", count: data.range.length)
  contents = contents.bridge().replacingCharacters(in: data.range, with: replacement)
}
```

Sourcery/Parsing/Parser.swift
```diff
@@ -453,18 +456,57 @@ extension Parser {
 
         return annotations
     }
+    
+    fileprivate func parseTypealiases(from source: [String: SourceKitRepresentable], containingType: Type?, processed: [[String: SourceKitRepresentable]]) -> [Typealias] {
+        var contents = self.contents
+        
+        //replace all processed substructures with whitespaces so that we don't process their typealiases again
+        for substructure in processed {
```
krzysztofzablocki: if we are doing that, shouldn't we clear processed then? won't you repeat the same clearing out logic every time and for data that was already replaced?

Sourcery/Parsing/Parser.swift
ilyapuchka: I'm not quite sure that this will work as `replacingCharacters` needs `Range<Self.Index>` and `SubstringIdentifier.key.range(for: source)` still returns ints 🤔 We might though change it to `replacingOccurrences` but I think it's not better from performance side (we already have a range, why to search for it again)

Maybe something like this:

```swift
if let range = SubstringIdentifier.key.range(for: source)
  .flatMap({ contents.byteRangeToNSRange(start: Int($0.offset), length: Int($0.length)) }) {
            
    let replacement = String(repeating: " ", count: range.length)
    contents = contents.bridge().replacingCharacters(in: range, with: replacement)
}
```

Sourcery/Parsing/Parser.swift
ilyapuchka: That's actually already happening as in `walkType` we create new `processed` array again (line 189). This serves to track processed inner types of each other type. We also keep one separate `processed` for global level types.

krzysztofzablocki: should we rename those variables to make it more clear then?

ilyapuchka:
I think I'm done here and hope most of the cases will be properly handled now. But I could miss some of edge cases. We can add them later as improvements if we encounter any issues. There are also few TODO's I left for some of very rare edge cases, hope you are ok with that @krzysztofzablocki

krzysztofzablocki:
could you rebase the project on top of develop and add changelog info about what supported and what not?

ilyapuchka:
Do you mind if I squash everything? I'm afraid rebasing like it is now will make me to solve lots of merge conflicts.

krzysztofzablocki:
not a problem

krzysztofzablocki:
uh I messed up something with Danger, will fix :)

krzysztofzablocki:
I'll remove that merge commit by hand later, wanted to see danger work ;)

Sourcery/Parsing/Parser.swift
```diff
@@ -523,18 +553,86 @@ extension Parser {
...
+                    continue
+                }
+                
+                guard let alias = extract(tokens[index + 1], contents: contents),
+                    let type = extract(tokens[index + 2], contents: contents) else {
+                        continue
+                }
+                
+                //get all subsequent type identifiers
+                var subtypes = [type]
```
krzysztofzablocki: I refactored some code and have changes as I'm adding support for one typealias to reference another, but I'm hitting issue with this method(didn't touch it), this test should pass right?:
```swift
fit("extracts typealiases of other typealiases") {
    expect(sut?.parseContents("typealias Foo = Int; typealias Bar = Foo").typealiases)
        .to(contain([
            Typealias(aliasName: "Foo", typeName: "Int"),
            Typealias(aliasName: "Bar", typeName: "Foo"),
            ]))
}
```

krzysztofzablocki: this is what I'm getting on this branch (without any of my changes):
```swift
got <[Typealias: aliasName = Foo, typeName = Int.Foo, parentName = nil, name = Foo, , Typealias: aliasName = Bar, typeName = Foo, parentName = nil, name = Bar, ]>

```

krzysztofzablocki: it seems like the Int is getting threated as subtype in this method

ilyapuchka: yep, there is `else { break }` missing in the for loop

krzysztofzablocki: I don't see it, how this should look like?

krzysztofzablocki: cool that's what I thought ;)

krzysztofzablocki:
Closing this as your changes are part of #61

PR #53, opened 2016-12-18
Refactors + Fix flattening logic

krzysztofzablocki:
Some refactorings but also fixing the implements / inherits flattening: since if an entity is a protocol then it can't inherit but can implement, when it's a class then it can do both, so we can join inherits and implements and only do check for manual assignment (because it's recursive)

### DIFF
--- a/Sourcery/CodeGenerated/Description.generated.swift
+++ b/Sourcery/CodeGenerated/Description.generated.swift
@@ -26,7 +26,7 @@ extension Enum.Case.AssociatedValue {
     override var description: String {
         var string = "\(type(of: self)): "
         string += "name = \(name), "
-        string += "type = \(type), "
+        string += "type = \(typeName), "
         return string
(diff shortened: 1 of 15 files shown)


Sourcery/CodeGenerated/Description.generated.swift
```diff
@@ -26,7 +26,7 @@ extension Enum.Case.AssociatedValue {
     override var description: String {
         var string = "\(type(of: self)): "
         string += "name = \(name), "
-        string += "type = \(type), "
+        string += "type = \(typeName), "
```
ilyapuchka: Should be `string += "typeName = \(typeName), "`

krzysztofzablocki: This is what happens when people manually try to refactor code, instead of using Sourcery! :)

PR #54, opened 2016-12-18
Add section annotations

### LINKED ISSUE #49: Support sections for annotations
Taking as an example `AutoEquatable` protocol it can be tedious to add /// sourcery: skipEquatable for each of the standard library types. Will it be possible to support sections of annotations the same ways as `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`, for example like this?

```
// sourcery: skipAnnotations begin

extension String: AutoEquatable {}
extension Int: AutoEquatable {}
...
// sourcery: skipAnnotations end
```

krzysztofzablocki:
You can now annotate sections now, this closes #49

### DIFF
--- a/Sourcery/Models/Variable.swift
+++ b/Sourcery/Models/Variable.swift
@@ -56,11 +56,12 @@ class Variable: NSObject {
     /// sourcery: skipEquality, skipDescription
     internal var __parserData: Any?
 
-    init(name: String,
-         typeName: String,
+    init(name: String = "",
+         typeName: String = "",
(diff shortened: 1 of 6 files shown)


Sourcery/Models/Variable.swift
```diff
@@ -56,17 +56,19 @@ class Variable: NSObject {
     /// sourcery: skipEquality, skipDescription
     internal var __parserData: Any?
 
-    init(name: String,
-         typeName: String,
+    init(name: String = "",
```
ilyapuchka: why to use default values here?

PR #55, opened 2016-12-19
Add upperFirst Stencil filter to template rendering environment.

bejar37:
We don't have a canonical way for users to add Stencil filters to Sourcery without modifying Sourcery itself. Would be open to looking into implementing a nicer way to do this.

Regardless, I have a hunch that this particular filter will be useful to many Sourcery users. The `capitalize` filter in Stencil is for human readable text, not for manipulating source identifiers like in Sourcery. I wanted to make a transformation from lowerCamelCase to UpperCamelCase on an enum case name, but that's currently not possible with stencil. See commit comment for more details.

### DIFF
--- a/Sourcery/Sourcery.swift
+++ b/Sourcery/Sourcery.swift
@@ -12,8 +12,25 @@ import SwiftTryCatch
 internal class SourceryTemplate: Template {
     private(set) var sourcePath: Path = ""
     convenience init(path: Path) throws {
-        self.init(templateString: try path.read())
+        self.init(templateString: try path.read(), environment: SourceryTemplate.sourceryEnvironment())
         sourcePath = path
+    }
+
+    convenience init(templateString: String) {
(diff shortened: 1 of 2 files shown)


krzysztofzablocki:
In plans to extract some common code from SwiftGen into a framework that both tools would share, reference issue here https://github.com/AliSoftware/SwiftGen/issues/240

bejar37:
> Thanks for contributing!

Anytime! I've been having a lot of fun building a json serialization tool on top of Sourcery over the past couple of days, I hope to make more upstream contributions in the future.

bejar37:
BTW @krzysztofzablocki do you have an idea of when you'll cut a release with this merge?

bejar37:
It would be nice to get it sooner but if it's a lot of trouble I'll just wait.

krzysztofzablocki:
no worries, I've pushed `0.4.3` that should have your changes

PR #58, opened 2016-12-19
Adding Support for SwiftPM

vknabel:
This PR adds support for macOS-only SwiftPM support.
Currently building is only supported by running `xcrun -sdk macosx swift build`.
This implementation currently uses some forks instead of the real projects in order to add (currently) missing version tags.

Currently missing:
- [x] Support for `swift build`
- [x] README.md updates
- [x] CircleCI and TravisCI updates
- [ ] Support for `swift test`


krzysztofzablocki:
thanks for contributing!
What do you think would be the best approach to follow up on those checkbox items?

vknabel:
I will open some follow up PRs. This way others can start to experiment with Homebrew support. 

Did you use Xcode's Swift? Mine has been installed with swiftenv.

vknabel:
@krzysztofzablocki Adding support for `swift test` would take a bit more time as there are compile errors. Shall we merge this PR? I would open another PR later.

Once it's merge-time: in the CHANGELOG.md I used `Future Release` as a placeholder. Before you merge it I may change it to the next version.

krzysztofzablocki:
Sounds good to me, you can just put `# Master` as release version, we gather changes that are in master and then when I do next release I change it to future version :)

Any chance we could change the product name in SPM version to be lowercase `sourcery` rather than uppercase?

krzysztofzablocki:
We should update the readme entry to lowercase as well, thank you so much for your help with this :)

krzysztofzablocki:
I think I pressed the merge too fast instead of waiting on CI, any idea why `error: these referenced modules could not be found: sourcery` would happen ? should we lowercase the package name as well?

vknabel:
@krzysztofzablocki Git has sometimes problems when renaming files from `Uppercase` to `Lowercase`

PR #61, opened 2016-12-20
Typealiases and Enum cases

### LINKED ISSUE #38: Type aliases should be scoped to the containing type
With current implementation type aliases with the same name defined in different types will override each other. That breaks the usual case of defining type alias to conform to protocol.

```swift

struct Foo {}
enum FooEnum: RawRepresentable {
  typealias RawValue = Foo
  var rawValue: RawValue { ... }
}

struct Bar {}
enum BarEnum: RawRepresentable {
  typealias RawValue = Bar
  var rawValue: RawValue { ... }
}
```

In this case type of `rawValue` property of `FooEnum` will be detected incorrectly as `Bar` as `RawValue = Bar` will override `RawValue = Foo`

### LINKED ISSUE #59: Annotations should be available for Enum.Case.
Hi everyone!

As of version 0.4.3, annotations are available for types and variables, but not for cases of enum's.

I think it would be very nice to have access to annotations for enum cases. 

For example, I'm trying to wrap localized strings in some custom `MyLocalizable` structs and starting from
````swift
enum Localizable {
    enum Category {
        /// sourcery comment = "Some comment"
        case someKey = "Default value"
    }
}
````
I would like to generate
````swift
struct Localizable {
    struct Category {
        static let someKey = MyLocalizable(key: "Localizable.Category.someKey", default: "Default value", comment: "Some comment")
    }
}
````
Please note that the enum is not part of the project. It is used only by Sourcery to generate code.
`MyLocalizable` has a `string` var that returns the localized string and I can then simply use `Localizable.Category.someKey.string`. where I need to. The Localizable.strings files can easily be generated from Sourcery too.

With the current version, the best I can do is to use a struct instead of an enum:
````swift
struct Localizable {
    struct Category {
        /// sourcery comment = "Some comment"
        /// sourcery default = "Default value"
        let someKey: String
    }
}
````
Here, I need to use another comment for the default since one doesn't have the rawValue anymore. Furthermore, I need to specify the type of key, even if I'm not using it after. This is not as optimal as the enum case.

An alternative solution would be to give access to literal values of variables when they have some. For example, for the variable `let foo = "bar"` could maybe have a literalValue property returning `"bar"`.

### PR DESCRIPTION
**Pull request by krzysztofzablocki**
I added referencing type aliases of other typealiases and updating inherited types on top of your changes. Fixes #38 

Also squeezed in enum cases, fixes #59

### CODE REVIEW

File: Sourcery/Parsing/Parser.swift [hunk 1]
```diff
@@ -297,9 +297,22 @@ final class Parser {
             })
         }
 
-        //TODO: replace typealiases of typealiases
-        //typealias Foo = Int; typealias Bar = Foo
-        //TODO: replace typealiases in inherited types
+        //! if a typealias leads to another typealias, follow through and replace with final type
+        typealiasesByNames.forEach { _, alias in
+
+            var aliasNamesToReplace = [alias.name]
```
ilyapuchka: shouldn't it be empty at the beginning? And I'm not sure I can get the following loop.

krzysztofzablocki: easier logic this way, it will reassign to the same place if there is no alias redirection, which is neglible.

What this loop does:

1. It looks whether an alias is redirecting to another typealias, and it follows through until it finds final alias, all while gathering all typealiases on the way. 
2. Replaces all intermediate typealiase names with the final typealias

Basically if you had `A -> B -> C -> D` after this loop you'd have `A -> D` `B -> D` `C -> D`

ilyapuchka: ok, now got it, makes sence

ilyapuchka:
Good to merge, I think 👍

krzysztofzablocki:
Thanks :)

PR #66, opened 2016-12-21
Checking for processed ranges instead of replacing content with spaces when processing typealiases

### LINKED ISSUE #65: Figure out more reliable approach to parsing typealiases
I've created a workaround for the crash we were having parsing the typealiases, [take a look here](https://github.com/krzysztofzablocki/Sourcery/commit/9e2f8f548a5d467eeff75a0e1ad29a17e100613a). 

I'm still not sure that this is reliable approach so if we can find something that doesn't rely on modifying the string contents, that should be a better way of doing it.

ilyapuchka:
Resolves #65

I guess it should be much more reliable now, at least no need for weird operations with strings.

### DIFF
--- a/Sourcery/Parsing/Parser.swift
+++ b/Sourcery/Parsing/Parser.swift
@@ -26,6 +26,8 @@ private extension Type {
     }
 }
 
+typealias SourceRange = (offset: Int64, length: Int64)
+
 fileprivate enum SubstringIdentifier {
     case body
     case key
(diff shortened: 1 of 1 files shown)


krzysztofzablocki:
I run it against NYT codebase and the performance is significantly affected: 
current master takes: `8s`
this version takes: `77s`

you can see performance by just adding this in main:
```swift
let time = CACurrentMediaTime()
defer { print("Time: \(CACurrentMediaTime() - time)") }
```

Maybe we should leave current implementation as is for now? we can come back to this if we get crash reports from people

ilyapuchka:
@krzysztofzablocki I'll try to see how that can be improved

ilyapuchka:
@krzysztofzablocki I've made some changes, can you please try again?
P.S. we surely need to implement a set of performance tests

krzysztofzablocki:
Time: 36.8333016719989

It's better but still way slower than your initial approach with my tweak, shall we close this for now? If we encounter crashes or wrong functionality we can go back to this, but no point in wasting our time if it works properly now

ilyapuchka:
Sure, I'll close it

krzysztofzablocki:
your initial idea was the best 👍

PR #68, opened 2016-12-22
Implement Diff matchers

krzysztofzablocki:
When working on tests I often find myself having to scan a lot of logs just to see whats different, this is annoying. 

I've implemented diffing for tests.

From this:
before

to this:

after

### DIFF
--- a/Sourcery/CodeGenerated/Diffable.generated.swift
+++ b/Sourcery/CodeGenerated/Diffable.generated.swift
@@ -1,0 +1,95 @@
+// Generated using Sourcery 0.4.7 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+
+extension Enum {
+     override func diffAgainst(_ object: Any?) -> DiffableResult {
+        var results = DiffableResult()
+
+        guard let rhs = object as? Enum else {
+          results.append("Incorrect type, expected Enum, received: \(type(of: object))")
+          return results
+        }
+        if self.cases != rhs.cases { results.append("Incorrect cases, expected \(self.cases), received: \(rhs.cases)") }
+        if self.rawType != rhs.rawType { results.append("Incorrect rawType, expected \(self.rawType), received: \(rhs.rawType)") }
+
+        results.append(contentsOf: super.diffAgainst(object))
+        return results
+    }
+}
+
+extension Protocol {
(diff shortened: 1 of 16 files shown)


Sourcery/CodeGenerated/Diffable.generated.swift
```diff
@@ -0,0 +1,103 @@
...
+        var results = DiffableResult()
+        guard let rhs = object as? Enum else {
+          results.append("Incorrect type, expected Enum, received: \(type(of: object))")
+          return results
+        }
+        if self.cases != rhs.cases { results.append("Incorrect cases, expected \(self.cases), received: \(rhs.cases)") }
+        if self.rawType != rhs.rawType { results.append("Incorrect rawType, expected \(self.rawType), received: \(rhs.rawType)") }
+
+        results.append(contentsOf: super.diffAgainst(object))
+        return results
```
ilyapuchka: I think we need to call `super` here

Sourcery/Models/Diffable.swift
```diff
@@ -0,0 +1,23 @@
...
+
+    /// Returns `DiffableResult` for the given objects.
+    ///
+    /// - Parameter object: Object to diff against.
+    /// - Returns: Diffable results.
+    func diffAgainst(_ object: Any?) -> DiffableResult
+}
+
+/// Phantom protocol for code generation
+protocol AutoDiffable {}
```
ilyapuchka: How do you plan to use that? Looks like just having `Diffable` will be enough, no?

ilyapuchka:
Looks very cool and useful indeed! What else are you planning to do here to make it not `wip`?

Sourcery/Models/Diffable.swift
krzysztofzablocki: In our case, but I'd like to write template a bit more generic than that as we could put it as example then

Sourcery/CodeGenerated/Diffable.generated.swift
krzysztofzablocki: yeah I kind of needs to add `super` access to `Type` so we can do things like `{% if type.super.implements.Diffable %}`

krzysztofzablocki:
access to supertype and maybe allow recursive diff for properties that are composed

krzysztofzablocki:
@ilyapuchka ready for review :)

PR #72, opened 2016-12-22
Extracting methods

### LINKED ISSUE #24: Add methods reflection
To be able to use Sourcery to generate code for such things like test doubles or dependency injection container it's vital to be able to access information about methods defined in class (or even on a protocol?).

To start method can be modeled like this:

```swift
class Method: NSObject {
  class Parameter: NSObject {
     var argumentLabel: String
     var parameterName: String
     var type: String
  }
  var parameters: [Parameter]
  var shortName: String //does not include external parameters names
  var fullName: String
  var returnType: String
  var accessLevel: AccessLevel
  var isStatic: Bool
  var isClass: Bool
}
```

Then in template it might be used something like this (taking [Dobby](https://github.com/trivago/Dobby) as an example):

```
{% for type in types.classes %}
    {% for method in type.methods %}
    override func {{ method.shortName }}({% for parameter in method.parameters %}{% if parameter.argumentLabel %}{{ parameter. argumentLabel }} {% endif %}{{ parameter.parameterName}}: {{ parameter.type }}{% endfor %}) -> {{ method.returnType }} {
      /* record a method call */
    }
  }
{% endfor %}
```

ilyapuchka:
Resolves #24


Sourcery/Models/Method.swift
```diff
@@ -0,0 +1,71 @@
...
+
+        init(typeName: String, argumentLabel: String? = nil, parameterName: String) {
+            self.typeName = typeName
+            self.argumentLabel = argumentLabel ?? parameterName
+            self.parameterName = parameterName
+        }
+    }
+    
+    private(set) var parameters: [Parameter]
+    var shortName: String
```
krzysztofzablocki: we should document those things as it's unclear to me what each means

Sourcery/Models/Method.swift
```diff
@@ -0,0 +1,71 @@
...
+        
+        self.fullName = fullName
+        self.shortName = fullName.range(of: "(").map({ fullName.substring(to: $0.lowerBound) }) ?? fullName
+        self.parameters = parameters
+        self.returnTypeName = returnTypeName
+        self.accessLevel = accessLevel
+        self.isStatic = isStatic
+        self.isClass = isClass
+        
+        if let argumentLabels = fullName.range(of: "(")
```
krzysztofzablocki: wouldn't this logic be better off in Parser?

Sourcery/Parsing/Parser.swift
```diff
@@ -252,30 +259,37 @@ final class Parser {
...
-
-        switch (containingType, type) {
-        case let (_, variable as Variable):
-            containingType.variables += [variable]
-            if !variable.isStatic {
-                if let enumeration = containingType as? Enum,
-                    let updatedRawType = parseEnumRawType(enumeration: enumeration, from: variable) {
-
-                    enumeration.rawType = updatedRawType
+        if let containingType = containing as? Type {
```
krzysztofzablocki: keep the guard or use if function and return at the end, if else statements are not very readable and requires us to read everything, the sooner we can break flow the easier it is to process

Sourcery/Models/Method.swift
ilyapuchka: do you mean to move that to `uniqueTypes`?

Sourcery/Models/Method.swift
```diff
@@ -0,0 +1,71 @@
+import Foundation
+
+class AMethod: NSObject {
```
ilyapuchka: I'm not a fan of `AMethod` name, but using `Method` causes ambiguities. Can you suggest any better name?

Sourcery/Models/Method.swift
krzysztofzablocki: wasn't thinking where it makes most sense, anything preventing us from just doing this in the place where we are creating this entity? I just think this looks like parsing domain and not initializer, `uniqueType` is for final step of processing so if we can just put it where we parse methods instead of waiting until unique types thats better

Sourcery/Models/Method.swift
krzysztofzablocki: if we can't use `Method` maybe `TypeMethod` ?

ilyapuchka:
@krzysztofzablocki ready for final review

Sourcery/Models/Type.swift
```diff
@@ -70,12 +78,12 @@ class Type: NSObject, AutoDiffable {
     /// sourcery: skipDescription
     var based = [String: String]()
 
-    /// contains all types implementing known BaseProtocol
+    /// contains all types implementing known BaseClass
```
krzysztofzablocki: `/// contains all types inheriting from known BaseClass`

Sourcery/Parsing/Parser.swift
```diff
@@ -7,25 +7,27 @@ import Foundation
 import SourceKittenFramework
 import PathKit
 
-private extension Variable {
-    /// Source structure used by the parser
-    var __underlyingSource: [String: SourceKitRepresentable] {
-        return (__parserData as? [String: SourceKitRepresentable]) ?? [:]
-    }
+protocol Parsable {
```
krzysztofzablocki: :+1:

Sourcery/Parsing/Parser.swift
```diff
@@ -7,25 +7,27 @@ import Foundation
...
+private extension Parsable {
     /// Source structure used by the parser
     var __underlyingSource: [String: SourceKitRepresentable] {
         return (__parserData as? [String: SourceKitRepresentable]) ?? [:]
     }
-
+    
     /// sets underlying source
-    func setSource(source: [String: SourceKitRepresentable]) {
+    mutating func setSource(_ source: [String: SourceKitRepresentable]) {
```
krzysztofzablocki: can you add `:class` to Parsable definition and then remove this `mutating`?

Sourcery/Parsing/Parser.swift
```diff
@@ -173,7 +177,7 @@ final class Parser {
     internal func parseTypes(_ source: [String: SourceKitRepresentable], existingTypes: [Type] = [], processed: inout [[String: SourceKitRepresentable]]) -> [Type] {
         var types = existingTypes
         walkTypes(source: source, processed: &processed) { kind, name, access, inheritedTypes, source in
-            let type: Type
+            var type: Type
```
krzysztofzablocki: I don't think this needs to be var?

Sourcery/Parsing/Parser.swift
```diff
@@ -252,102 +259,79 @@ final class Parser {
...
-        case let (_, variable as Variable):
-            containingType.variables += [variable]
-            if !variable.isStatic {
-                if let enumeration = containingType as? Enum,
-                    let updatedRawType = parseEnumRawType(enumeration: enumeration, from: variable) {
-
-                    enumeration.rawType = updatedRawType
-                }
+        
+        if let containingMethod = containing as? Method {
```
krzysztofzablocki: I'd reverse the order, put this on top of function since it's shorter and then the structure is a little easier to read

Sourcery/Parsing/Parser.swift
```diff
@@ -364,16 +348,52 @@ final class Parser {
...
                     variable.type = unique[actualTypeName]
                 } else {
                     variable.type = unique[variable.unwrappedTypeName]
                 }
             }
-        }
+            // find actual methods parameters types and their argument labels
+            for method in type.methods {
+                let argumentLabels: [String]
+                if let labels = method.fullName.range(of: "(")
```
krzysztofzablocki: what will happen if you have spaces in function definition?

Sourcery/Parsing/Parser.swift
```diff
@@ -364,16 +348,52 @@ final class Parser {
...
+                    .map({ method.fullName.substring(from: $0.upperBound) })?
+                    .trimmingCharacters(in: CharacterSet(charactersIn: ")"))
```
krzysztofzablocki: what if the argument is a closure and has multitple ()?

Sourcery/Parsing/Parser.swift
```diff
@@ -364,16 +348,52 @@ final class Parser {
...
+                    .components(separatedBy: ":")
+                    .dropLast() {
+                    argumentLabels = Array(labels)
+                } else {
+                    argumentLabels = []
+                }
 
-        for (_, type) in unique {
+                for (index, parameter) in method.parameters.enumerated() {
+                    if argumentLabels.count > index {
```
krzysztofzablocki: `index < argumentLabels.count` is easier to read

Sourcery/Parsing/Parser.swift
```diff
@@ -364,16 +348,52 @@ final class Parser {
...
+                    }
+                    
+                    if let actualTypeName = typeName(for: parameter.unwrappedTypeName, containingType: type, typealiases: typealiases) {
+                        parameter.type = unique[actualTypeName]
+                    } else {
+                        parameter.type = unique[parameter.unwrappedTypeName]
+                    }
+                }
+                
+                if method.returnTypeName != "Void" && method.returnType == nil {
```
krzysztofzablocki: is it possible for `method.returnType` to be set before this logic?

Sourcery/Parsing/Parser.swift
```diff
@@ -431,12 +502,50 @@ extension Parser {
             computed = false
         }
 
-        let variable = Variable(name: name, typeName: type, accessLevel: (read: accesibility, write: writeAccessibility), isComputed: computed, isStatic: isStatic)
+        var variable = Variable(name: name, typeName: type, accessLevel: (read: accesibility, write: writeAccessibility), isComputed: computed, isStatic: isStatic)
```
krzysztofzablocki: unnecesary var

SourceryTests/ParserSpec.swift
```diff
@@ -119,13 +119,128 @@ class ParserSpec: QuickSpec {
...
+                                Type(name: "Foo", methods: [
+                                    Method(fullName: "foo(bar:foo:)", parameters: [
+                                        Method.Parameter(name: "bar", typeName: "Int"),
+                                        Method.Parameter(name: "foo", typeName: "String")
+                                        ], returnTypeName: "Void")
+                                    ])
+                            ]))
+                        }
+                        
+                        it("extracts method with parameter with two names") {
```
krzysztofzablocki: add tests for more complex methods:
- method with a closure (or two) that has some arguments and return type
- method with extra spaces between arguments

Sourcery/Parsing/Parser.swift
ilyapuchka: I think it's easier to do it here, so that we already have a final type in the `returnType`, instead of for example extension if this method is contained in extension and then needing to post process it again.

SourceryTests/ParserSpec.swift
ilyapuchka: I guess we will need a separate improvement for dealing with closures, like having `isClosure` property, list of closure parameters types, escaping and so on.

Sourcery/Parsing/Parser.swift
ilyapuchka: There are no parameter types if `fullName`, so the only parentheses there are for the method parameters

Sourcery/Parsing/Parser.swift
ilyapuchka: SourceKit already handles stripping all the spaces and `fullName` does not have any

Sourcery/CodeGenerated/Diffable.generated.swift
```diff
@@ -1,127 +1,150 @@
...
+
 extension Enum {
-     override func diffAgainst(_ object: Any?) -> DiffableResult {
+    override func diffAgainst(_ object: Any?) -> DiffableResult {
         let results = DiffableResult()
-
         guard let rhs = object as? Enum else {
-          results.append("Incorrect type <expected: Enum, received: \(type(of: object))>")
-          return results
+            results.append("Incorrect type <expected: Enum, received: \(type(of: object))>")
```
ilyapuchka: Should I do any additional setup for it? After installing it it started to complain about all lowercased cases

ilyapuchka: Nevermind, brew installed old version with latests (0.15) everything is ok

SourceryTests/ParserSpec.swift
krzysztofzablocki: I think for now we can just treat whole closure as a parameter, without extra handling. Just so it can be used as other parameters

Sourcery/Parsing/Parser.swift
krzysztofzablocki: I was asking because you have if check whether is set already and though it might be not necessary

ilyapuchka: Ah, yes, exactly, I think I just forgot to remove it.

krzysztofzablocki:
I'll merge this by hand

krzysztofzablocki:
merged in master, will make a new release in a moment 👍

PR #76, opened 2016-12-25
Add type for Enum.Case.AssociatedValue

inamiy:
Hi @krzysztofzablocki !
I want to examine what protocol is used in `Enum.Case.AssociatedValue`'s type,
but currently `typeName` is only supported, and I think we need `type` accessor too,
just as `Variable` has.

### Stencil example

```
{% for inherited in case.associatedValues.0.type.inheritedTypes %}
{% if inherited == "SomeProtocol" %}
// do something
{% endif %}
{% endfor %}
```

### DIFF
--- a/Sourcery/Parsing/Parser.swift
+++ b/Sourcery/Parsing/Parser.swift
@@ -377,7 +377,11 @@ final class Parser {
             if let enumeration = type as? Enum, enumeration.rawType == nil {
                 enumeration.cases.forEach { enumCase in
                     enumCase.associatedValues.forEach { associatedValue in
-                        associatedValue.type = unique[associatedValue.typeName]
+                        if let actualTypeName = typeName(for: associatedValue.unwrappedTypeName) {
+                            associatedValue.type = unique[actualTypeName]
+                        } else {
+                            associatedValue.type = unique[associatedValue.unwrappedTypeName]
+                        }
                     }
                 }
 
--- a/Sourcery/Parsing/Parser.swift
+++ b/Sourcery/Parsing/Parser.swift
@@ -375,6 +375,12 @@ final class Parser {
 
         for (_, type) in unique {
             if let enumeration = type as? Enum, enumeration.rawType == nil {
+                enumeration.cases.forEach { enumCase in
+                    enumCase.associatedValues.forEach { associatedValue in
+                        associatedValue.type = unique[associatedValue.typeName]
+                    }
+                }
+
                 guard let rawTypeName = enumeration.inheritedTypes.first else { continue }
                 if let rawTypeCandidate = unique[rawTypeName] {
                     if !(rawTypeCandidate is Protocol) {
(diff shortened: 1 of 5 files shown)


Sourcery/Parsing/Parser.swift
```diff
@@ -375,6 +375,12 @@ final class Parser {
 
         for (_, type) in unique {
             if let enumeration = type as? Enum, enumeration.rawType == nil {
+                enumeration.cases.forEach { enumCase in
+                    enumCase.associatedValues.forEach { associatedValue in
+                        associatedValue.type = unique[associatedValue.typeName]
```
krzysztofzablocki: you should resolve typealiases here, you can copy the snippet that variable uses just above this for loop, (I'll be refactoring those repetitions separately).

krzysztofzablocki:
ps. once this type is merged you can improve your stencil:
```swift
{% if case.associatedValues.0.type.implements.SomeProtocol %}
// do something
{% endif %}
```

Sourcery/Parsing/Parser.swift
inamiy: Done in 0d82f3c.

I also added `Enum.Case.AssociatedValue.unwrappedTypeName` (same as `Variable`) for safely parsing optional type.
This could be abstracted in future update 😉

inamiy:
0d82f3c Tests are added!

SourceryTests/ParserSpec.swift
```diff
@@ -496,6 +496,74 @@ class ParserSpec: QuickSpec {
...
+                            let enumeration = types[2] as? Enum
+                            let enumCases = enumeration?.cases
+                            expect(enumCases?.count) == 1
+
+                            let associatedValues = enumCases?.first?.associatedValues
+                            expect(associatedValues?.count) == 1
+
+                            let associatedValue = associatedValues?.first
+                            expect(associatedValue?.name) == "key"
+                            expect(associatedValue?.typeName) == "Bar"
```
krzysztofzablocki: You can use equality against whole type like we do in other tests, this has benefits because we diff data and it will show us better errors if test fail

inamiy: Done in a3957ac.
I also updated `Equality.generated.swift` for better `AssociatedValue` equality check.

krzysztofzablocki:
I can tweak this when I'll be merging this pr (either later today or tomorrow). Thanks for contributing! ;-)

inamiy:
Yeah, please! 🙏

krzysztofzablocki:
ok, your changes are in master and will be available in next release :) thank you!

inamiy:
Thanks for merge!

By the way, here's my (tricky!) project that now uses Sourcery if you are interested.
https://github.com/inamiy/VTree/pull/3

So far, Sourcery works like a charm even in my edge case!
Thanks for the library, and please keep up this great work 💪

krzysztofzablocki:
That's cool, reminds me of Redux approach;) 
BTW my general recommendation for type markup is using empty protocol since its harder to make a mistake and you can use types.implementing.AutoVType, annotations are better for tweaking particular variables etc

inamiy:
Right, I will note that :)

PR #79, opened 2016-12-25
Simplify typename logic and dry out code

krzysztofzablocki:


Sourcery/Models/Method.swift
```diff
@@ -67,23 +56,14 @@ class Method: NSObject, AutoDiffable {
...
     var unwrappedReturnTypeName: String {
-        guard isOptionalReturnType else { return returnTypeName }
-        guard !isFailableInitializer else { return returnTypeName }
-
-        if returnTypeName.hasSuffix("?") {
-            return String(returnTypeName.characters.dropLast())
-        } else {
-            return String(returnTypeName.characters.dropFirst("Optional<".characters.count).dropLast())
-        }
+        guard !isFailableInitializer else { return returnTypeName.unwrappedTypeName }
```
ilyapuchka: I guess there is no more need in this `guard`

Sourcery/Parsing/Parser.swift
```diff
@@ -374,21 +376,14 @@ final class Parser {
...
+                    parameter.type = resolveType(parameter.unwrappedTypeName, type, typealiases)
                 }
 
-                if method.returnTypeName != "Void" {
-                    if let actualTypeName = typeName(for: method.unwrappedReturnTypeName, containingType: type, typealiases: typealiases) {
-                        method.returnType = unique[actualTypeName]
-                    } else {
-                        method.returnType = unique[method.unwrappedReturnTypeName]
-                    }
+                if method.returnTypeName.name != "Void" {
```
ilyapuchka: I just remembered that we should also check for `()`, can you add this check here too? Or maybe better will be to incapsulate that into a property on `TypeName` like `isVoid` and use it here?

Sourcery/Models/Method.swift
```diff
@@ -13,32 +13,21 @@ class Method: NSObject, AutoDiffable {
         let name: String
 
         /// Parameter type name
-        let typeName: String
+        let typeName: TypeName
```
ilyapuchka: How about adding a common protocol for `Variable`, `Method.Parameter`, `AssociatedValue` like this:

```swift
protocol Typed {
  var type: Type? { get set }
  var typeName: TypeName { get }
  var isOptional: Bool { get }
  var unwrappedTypeName: String { get }
}
extension Typed {
  var isOptional: Bool { return typeName.isOptional }
  var unwrappedTypeName: String { return typeName.unwrappedTypeName }
}
```

Sourcery/Parsing/Parser.swift
krzysztofzablocki: :+1:

Sourcery/Models/Method.swift
krzysztofzablocki: great idea :)

krzysztofzablocki:
@ilyapuchka ready for final review :)

PR #80, opened 2016-12-26
Processing additional cli arguments

### LINKED ISSUE #14: User should be able to specify additional template parameters on the command line.
There is a great sample of similar templating engine (http://atomist.com) in this talk (although for Elm):
https://www.youtube.com/watch?v=jJ4e6cIBgYM

Also default values for unspecified arguments could be configured using yaml config file. This way we could for example write:

`sourcery MyViewController.swift add_table_view_cell MyCellType --fixed_height=100`

ilyapuchka:
Resolves #14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Sourcery CHANGELOG
 
 ---
+
+## Master
+
+- You can now pass arbitrary values to templates with `--args` argument.
+
 ## 0.4.9
 
 ### New Features
--- a/README.md
+++ b/README.md
@@ -350,13 +350,14 @@ You can clone it from the repo and just run `Sourcery.xcworkspace`.
 ## Usage
 Sourcery is a command line tool `sourcery`:
 ```
-$ ./sourcery <source> <templates> <output>
+$ ./sourcery <source> <templates> <output> [--args arg1=value,arg2]
 ```
 Arguments:
 
 - source - Path to a source swift files.
 - templates - Path to templates. File or Directory.
 - output - Path to output. File or Directory.
+- args - Additional arguments to pass to templates. Each argument can have explicit value or will have implicit `true` value. Arguments should be separated with `,` without spaces.
 
 Options:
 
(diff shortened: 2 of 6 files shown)


krzysztofzablocki:
What's the reason for the whitespace limitation?

ilyapuchka:
Commander that is used to parse command line attributes. It breaks them by spaces. There is though an alternative approach. At first I just pulled all remaining positional arguments in a dictionary. But that will require them to appear only in the end of the command and without `--args`, like this:

`sourcery <paths> arg1=value arg2 arg3=value`

krzysztofzablocki:
Ok thanks, I'll take a look today at tweaking this a little to use the same parsing code that we use for annotation line parsing so we have exactly same behaviour in both

krzysztofzablocki:
moved to #82

AliSoftware:
I think @djbe added the ability to specify a flag multiple times in Stencil via a PR?
So you could do `--args x 1 --args y 2`?

djbe:
No it was in Commander, and it was for repeating parameters like:
`--import Something --import SomethingElse`
Which resulted in a `import` array variable with as values `["Something", "SomethingElse"]`.

I think a better command line syntax would be:
`--arg x=1 --arg y=2`
Where you just parse the contents of the `arg` array.

AliSoftware:
Ah yeah of course I meant Commander ^^
Ok so this is definitely possible now 😉

ilyapuchka:
Hm, didn't know about that option. @krzysztofzablocki should we use this feature instead of what we have implemented?

djbe:
I'm planning on adding this functionality into StencilSwiftKit, so we can unify this for both codebases. Something that receives an array of strings such as:
```
[
  "foo.bar=test1",
  "foo.baz=test2",
  "AppModule=MyApp",
  "items=a",
  "items=b"
]
```
And converts it into a stencil context dictionary:
```
[
  "AppModule": "MyApp",
  "foo": [
    "bar": "test1",
    "baz": "test2",
  ],
  "items": [
    "a",
    "b"
  ]
]
```
To which you could then add whatever variables your parser provides.

krzysztofzablocki:
I'm not entirely sure this isnt overkill

PR #83, opened 2016-12-26
Open access level

### LINKED ISSUE #73: Detect open variables, methods and classes
As in Swift not everything is automatically overridable there should be a way to check for that in templates to not try to override these methods/variables or subclass non-open classes. Can be implemented by adding `isOpen` property to methods, properties and classes. 

@krzysztofzablocki should we also add a dedicated type `Class` and move there properties like `supertype`, `inherits` and this new property as they are not applicable for structs, enums and protocols?

ilyapuchka:
Resolves #73

### DIFF
--- a/Sourcery/Generating/Generator.swift
+++ b/Sourcery/Generating/Generator.swift
@@ -99,10 +99,10 @@ enum Generator {
             }
 
             baseType.based.keys.forEach {  type.based[$0] = $0 }
+            baseType.inherits.keys.forEach {  type.inherits[$0] = $0 }
             baseType.implements.keys.forEach {  type.implements[$0] = $0 }
 
-            if let type = type as? Class, let baseType = baseType as? Class {
-                baseType.inherits.keys.forEach {  type.inherits[$0] = $0 }
+            if baseType is Class {
                 type.inherits[name] = name
             } else if baseType is Protocol {
                 type.implements[name] = name
--- a/Sourcery/Generating/Generator.swift
+++ b/Sourcery/Generating/Generator.swift
@@ -6,13 +6,6 @@
 import Foundation
 import Stencil
 
-private extension Type {
-    var isClass: Bool {
-        let isNotClass = self is Struct || self is Enum || self is Protocol
-        return !isNotClass && !isExtension
-    }
-}
-
 private class TypesReflectionBox: NSObject {
     private let types: [Type]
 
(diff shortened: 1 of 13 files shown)


Sourcery/Models/Class.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+// sourcery: skipDescription, skipEquatable
+class Class: Type {
+    override var kind: String { return "class" }
+
+    /// contains all types inheriting from known BaseClass
+    // sourcery: skipEquality
+    var inherits = [String: String]()
```
krzysztofzablocki: as much as I like having `Class` I've been thinking about this and I think we should include this properties in the base `Type` and just be empty or everyone will have to duplicate their templates because accessing supertype, inherits would crash the KVC templating engine. 

Usually template for classes / structs differs minimal, so have the `Class` type since it's useful but put those properties back in Type

ilyapuchka: Done

Sourcery/Generating/Generator.swift
```diff
@@ -106,10 +99,10 @@ enum Generator {
             }
 
             baseType.based.keys.forEach {  type.based[$0] = $0 }
-            baseType.inherits.keys.forEach {  type.inherits[$0] = $0 }
             baseType.implements.keys.forEach {  type.implements[$0] = $0 }
 
-            if baseType.isClass {
+            if let type = type as? Class, let baseType = baseType as? Class {
+                baseType.inherits.keys.forEach {  type.inherits[$0] = $0 }
```
krzysztofzablocki: I wouldn't be putting this expression here, it's impossible for swift to allow wrong logic here, that's why I put inherits and implements above this if statement

Sourcery/Templates/Diffable.stencil
```diff
@@ -2,12 +2,12 @@
 extension {{ type.name }}{% if not type.supertype.implements.AutoDiffable %}: Diffable{% endif %} {
     {% if type.based.Type %}override {% endif %}func diffAgainst(_ object: Any?) -> DiffableResult {
         let results = DiffableResult()
-        {% if type.storedVariables.count %}guard let rhs = object as? {{ type.name }} else {
+        {% for variable in type.storedVariables %}{% if not variable.annotations.skipEquality %}{% if forloop.first %}guard let rhs = object as? {{ type.name }} else {
```
krzysztofzablocki: I think this is less readable, also I think we can just do guard regardless, because even without variables if type mismatches then we should list it

ilyapuchka: Then it will produce a warning for unused property, that's why I changed that in the first place.

Sourcery/Generating/Generator.swift
ilyapuchka: Reverted

krzysztofzablocki:
thanks 👍 I'll merge this once CI passes

Issue #85, opened 2016-12-27
Wrong RawType on enums


txaiwieser:
I came across this today, don't know if its a bug, bug it was not what I was expecting to be the RawType

```swift
{% for enum in types.enums %}{% if enum.rawType %}
RawType => {{ enum.rawType }}
{% endif %}{% endfor %}


enum Test1 {
	case one
}
// Does not generate code // ok!



enum Test2: String {
	case two
}
// Does generate code
RawType => String // ok!



enum Test3: Int {
	case three
}
// Does generate code
RawType => Int // ok!



enum Test4: Equatable {
	case four
}
// Does generate code
RawType => Equatable // it this suppouse to be the RawType? I was expecting to be nil

```

Thanks 👍

krzysztofzablocki:
I think it happens because we don't know what Equatable is, since it's not part of your project. /cc @ilyapuchka

ilyapuchka:
Yes, that's the reason, but I will look at it again trying to improve the behaviour, so that it could be worked around by adding `Equatable` conformance in extension. Currently only using dummy protocol that extends `Equatable` will work, but extension approach sounds better.

krzysztofzablocki:
maybe we should whitelist primitive types that can become RawType automatically? I think any complex type would require the protocol conformance that we are already looking for automaticaly?

ilyapuchka:
We can, but that will be a pretty long list =)

krzysztofzablocki:
@txaiwieser you will be able to apply that workaround @ilyapuchka implemented in 0.5.0, we'll do release once my PR is merged

PR #86, opened 2016-12-27
Enabled workaround for enums raw types

ilyapuchka:
Enables workaround mentioned in #85  - `enum Foo { }; extension Foo: Equatable {}` will not infer `Equatable` as `rawType`

### DIFF
--- a/Sourcery/Models/Enum.swift
+++ b/Sourcery/Models/Enum.swift
@@ -51,6 +51,7 @@ class Enum: Type {
     internal(set) var rawType: String? {
         didSet {
             if let rawType = rawType {
+                hasRawType = true
                 if let index = inheritedTypes.index(of: rawType) {
                     inheritedTypes.remove(at: index)
                 }
@@ -60,6 +61,8 @@ class Enum: Type {
             }
         }
     }
+
+    private(set) var hasRawType: Bool
 
     /// sourcery: skipEquality
     /// sourcery: skipDescription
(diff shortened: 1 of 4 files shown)


Sourcery/Parsing/Parser.swift
```diff
@@ -390,6 +390,10 @@ final class Parser {
 
             //find enums rawValue types
             if let enumeration = type as? Enum, enumeration.rawType == nil {
+                guard enumeration.hasRawType != false else {
```
krzysztofzablocki: `guard enumeration.hasRawType else { continue }`

Sourcery/Models/Enum.swift
```diff
@@ -51,6 +51,7 @@ class Enum: Type {
     internal(set) var rawType: String? {
         didSet {
             if let rawType = rawType {
+                hasRawType = true
```
krzysztofzablocki: I'm a little confused, isn't this the same as `rawType == nil && inheritedTypes.isEmpty` ?

ilyapuchka: Not exactly, as we can extend enum what can add inherited types, but they can not be raw types in this case.

ilyapuchka: I've changed a bit the logic around `hasRawType`, now we set it right in the `init`, as we already should have directly inherited types list there.

krzysztofzablocki:
can you add entry in changelog under bugfixes so that danger is happy ? :)

ilyapuchka:
Changelog entry added

PR #88, opened 2016-12-28
Add allVariables and introduce bunch of filters

### LINKED ISSUE #84: Should we flatten variables?
If a type inherits from another type, or implements given protocol, should the variables be available in the given type itself?
e.g. 
```swift
protocol Something {
  var typeName: String { get }
}

struct Foo: Something {
}
```

1. Should we list typeName in Foo? if so should we add alternative variable so people can choose between inherited or direct variables?

2. Also same question can be asked in regards to annotations, if the protocol specified type level annotation, should we flatten that into Foo ?

@ilyapuchka what do you think?

krzysztofzablocki:
- `allVariables` that contains flattened list
- `static`, `instance`, `computed`, `stored` filters for Variable
- `enum`, `class`, `struct`, `protocol` filters for Type

This closes #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,10 +174,9 @@ For each type you can access following properties:
 - `kind` <- convience accessor that will contain one of `enum`, `class`, `struct`, `protocol`, it will also provide `extension` for types that are unknown to us(e.g. 3rd party or objc), but had extension in the project
 - `isGeneric` <- info whether the type is generic
 - `localName` <- name within parent scope
-- `staticVariables` <- list of static variables
-- `variables` <- list of instance variables
-- `computedVariables` <- list of computed instance variables
-- `storedVariables` <- list of computed stored variables
+- `variables` <- list of all variables defined in this type, excluding variables from protocols or inheritance
+  - if you want to access all available variables, including those from inherited / protocol, then use `allVariables`
+  - if you want to accces computed, stored, instance, or static variables, you can do so using our [custom filters](#custom-stencil-tags-and-filter) on both `variables` and `allVariables`
 - `methods` <- list of all methods
 - `initializers` <- list of all initializers
 - `inherits.BaseClass` => info whether type inherits from known base class
@@ -249,6 +248,8 @@ For each type you can access following properties:
 - `{% if name|contains: "Foo" %}` - check if `name` contains arbitrary substring
 - `{% if name|hasPrefix: "Foo" %}`- check if `name` starts with arbitrary substring
 - `{% if name|hasSuffix: "Foo" %}`- check if `name` ends with arbitrary substring
+- `static`, `instance`, `computed`, `stored` - can be used on Variable[s] as filter e.g. `{% for var in variables|instance %}`
+- `enum`, `class`, `struct`, `protocol` - can be used for Type[s] as filter
 
 ### Using Source Annotations
 
--- a/README.md
+++ b/README.md
@@ -250,7 +250,8 @@ For each type you can access following properties:
 - `{% if name|hasSuffix: "Foo" %}`- check if `name` ends with arbitrary substring
 - `static`, `instance`, `computed`, `stored` - can be used on Variable[s] as filter e.g. `{% for var in variables|instance %}`
 - `enum`, `class`, `struct`, `protocol` - can be used for Type[s] as filter
-
+- `count` - can be used to get count of filtered array
+-
 ### Using Source Annotations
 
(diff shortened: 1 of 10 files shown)


krzysztofzablocki:
I'm wondering what naming is better `{% for var in variables|static %}` or `{% for var in variables|isStatic %}` ? @vknabel @ilyapuchka

Sourcery/SourceryTemplate.swift
```diff
@@ -44,3 +64,24 @@ private func stringContentFilter(_ name: String, _ filter: @escaping (String) ->
         return filter(s)(arg)
     }
 }
+
+private struct Filter<T> {
+    static func make(_ filter: @escaping (T) -> Bool) -> (Any?) throws -> Any? {
+        return { (any) throws -> Any? in
+            switch any {
+            case let type as T:
+                if filter(type) {
```
ilyapuchka: just `return filter(type)`?

Sourcery/SourceryTemplate.swift
```diff
@@ -44,3 +64,24 @@ private func stringContentFilter(_ name: String, _ filter: @escaping (String) ->
...
+                    return true
+                } else {
+                    return false
+                }
+
+            case let array as NSArray:
+                return array.flatMap { $0 as? T }.filter(filter)
+
+            default:
+                return nil
```
ilyapuchka: Should we throw an error in this case?

Sourcery/SourceryTemplate.swift
```diff
@@ -44,3 +64,24 @@ private func stringContentFilter(_ name: String, _ filter: @escaping (String) ->
         return filter(s)(arg)
     }
 }
+
+private struct Filter<T> {
```
ilyapuchka: Would be nice to apply the same pattern for string filters

krzysztofzablocki: they use extra arguments, where this just filters, we can when we add more filters but out of scope of this pr

Sourcery/SourceryTemplate.swift
krzysztofzablocki: we don't in string cases, only if someone tries to pass wrong arguments, I changed it to return any as thats what string does

Sourcery/SourceryTemplate.swift
```diff
@@ -20,18 +20,20 @@ internal class SourceryTemplate: Template {
...
 
         ext.registerFilter("enum", filter: Filter<Type>.make({ $0 is Enum }))
-        ext.registerFilter("class", filter: Filter<Type>.make({ $0 is Class }))
         ext.registerFilter("struct", filter: Filter<Type>.make({ $0 is Struct }))
         ext.registerFilter("protocol", filter: Filter<Type>.make({ $0 is Protocol }))
 
         ext.registerFilter("count", filter: count)
 
+        ext.registerFilter("initializer", filter: Filter<Method>.make({ $0.isInitializer }))
+        ext.registerFilter("class", filter: FilterOr<Type, Method>.make({ $0 is Class }, other: { $0.isClass }))
```
ilyapuchka: Btw can't variables be `class` too?

krzysztofzablocki: They can but we don't support that right now, we only have isStatic on them

Sourcery/SourceryTemplate.swift
```diff
@@ -44,3 +66,44 @@ private func stringContentFilter(_ name: String, _ filter: @escaping (String) ->
...
+                return array.flatMap { $0 as? T }.filter(filter)
+
+            default:
+                return any
+            }
+        }
+    }
+}
+
+private struct FilterOr<T, Y> {
```
ilyapuchka: I wonder if using `Either<A, B>` would work here and would be more scalable, for example later if we add `isClass` property for variables. But we can try that later =)

README.md
```diff
@@ -249,7 +248,10 @@ For each type you can access following properties:
 - `{% if name|contains: "Foo" %}` - check if `name` contains arbitrary substring
 - `{% if name|hasPrefix: "Foo" %}`- check if `name` starts with arbitrary substring
 - `{% if name|hasSuffix: "Foo" %}`- check if `name` ends with arbitrary substring
-
+- `static`, `instance`, `computed`, `stored` - can be used on Variable[s] as filter e.g. `{% for var in variables|instance %}`
```
ilyapuchka: should we add a note about applying filters for methods?

krzysztofzablocki: already added

PR #91, opened 2016-12-31
Support variables with default value initializer

krzysztofzablocki:


Sourcery/Parsing/FileParser.swift
```diff
@@ -234,8 +234,25 @@ extension FileParser {
 
     internal func parseVariable(_ source: [String: SourceKitRepresentable], isStatic: Bool = false) -> Variable? {
         guard let (name, _, accesibility) = parseTypeRequirements(source),
-            accesibility != .private && accesibility != .fileprivate,
-            let type = source[SwiftDocKey.typeName.rawValue] as? String else { return nil }
+            accesibility != .private && accesibility != .fileprivate else { return nil }
+
+        var maybeType: String? = source[SwiftDocKey.typeName.rawValue] as? String
```
ilyapuchka: Can we do the same not by manually parsing string, but using structure output? In case of `var some = String()` it will give us

```
{
      "key.kind" : "source.lang.swift.expr.call",
      "key.offset" : 12,
      "key.nameoffset" : 12,
      "key.namelength" : 6,
      "key.bodyoffset" : 19,
      "key.bodylength" : 0,
      "key.length" : 8,
      "key.name" : "String"
    }
```
The same will be for arrays or dictionaries when we use `[String]()` or `[String: Int]()`. For `Foo(some: Int)`, it will report just `Foo` in `key.name`, in case of `Foo.init(...)`, it will report `Foo.init`, then we just trim `init` part. Btw there also could be any static method calls, like `Foo.makeBar()`, then we would need to get return type of that method...

krzysztofzablocki: Maybe I'm missing something but that's not the output from `Structure(source)`, it doesn't contain those type information.
We won't support `Foo.makeBar()` for now, this would be pretty hard to support reliably as you can have that in separate file and we are dealing with a single `FileParser`

krzysztofzablocki: as I see what you mean, calling it on the substring after removing `=`. 
I'm not sure if that's a good call for performance reasons, it will spin up sourcekit just for that expression?

ilyapuchka: It will be in the substructure already, right after the entry for the variable declaration. Try `sourcekitten structure --text "var some = String()"` The problem is that it returns it as a separate substructure component, what I didn't notice at first. So maybe it will be not that easy to use, though it looks like we can just check if there is expression call right after variable declaration... Manual parsing is fine, was just thinking if we can avoid doing the job that SourceKit already kind of did for us. Like for typealiases I would prefer to use it's output instead of manually walking through tokens or lines, but for some reason SourceKitten does not provide it =(

krzysztofzablocki: I think that might be good future improvement but it would require bigger changes as we'd need to correlate source with specific Structure.substructure, maybe think of that when we have another scenario where we could leverage Structure?

PR #92, opened 2017-01-02
Parsing tuples

### LINKED ISSUE #48: Handle tuples as variable types
As it's possible to compare tuples of `Equatable` types, but it's not possible to extend tuple types with protocols, we should support them to be able to use them in equatable template. For example it could be used like `{ % if variable.isTuple and variable.tuple.allTypes.implements.AutoEquatable % }`. Also we may support tuple elements names.

ilyapuchka:
Resolves #48

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@
 
 ### New Features
 - Variables with default initializer are now supported, e.g. `var variable = Type(...)`
+- Added support for tuple types and `tuple` filter for variables
+- Added `actualTypeName` for `TypeName` that is typealias
+- Added `implements`, `inherits` and `based` filters
+
+### Fixed
+- Fixed parsing typealiases for tuples and closure types
+- Fixed parsing associated values of generic types
 
 ### Internal Changes
 - Performed significant refactoring and simplified mutations in parsers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Variables with default initializer are now supported, e.g. `var variable = Type(...)`
 - Added support for special escaped names in enum cases e.g. `default` or `for`
 - Added support for tuple types and `tuple` filter for variables
+- Enum associated values now have `localName` and `exteranlName` properties.
 - Added `actualTypeName` for `TypeName` that is typealias
 - Added `implements`, `inherits` and `based` filters
 
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ Available types:
 <details><summary>**Enum.Case.AssociatedValue**. Properties:</summary>
 
 - `localName` <- name to use to construct value, i.e. `value` in `Foo.foo(value: ...)`
-- `exteranlName` <- name to use when binding value, i.e. `value` or `other` in `enum Foo { case foo(value: ..., other: ... )}`. Will use index as a fallback
+- `externalName` <- name to use when binding value, i.e. `value` or `other` in `enum Foo { case foo(value: ..., other: ... )}`. Will use index as a fallback
 - `typeName` <- name of type of associated value (*TypeName*)
 - `unwrappedTypeName` <- shorthand for `typeName.unwrappedTypeName`
 - `isOptional` <- shorthand for `typeName.isOptional`
@@ -287,8 +287,11 @@ Available types:
 <details><summary>**TupleType.Element**. Properties:</summary>
 
 - `name` <- element name
-- `typeName` <- element type name (*TypeName*)
 - `type` <- type of element, if known
+- `typeName` <- element type name (*TypeName*)
+- `unwrappedTypeName` <- shorthand for `typeName.unwrappedTypeName`
+- `isOptional` <- shorthand for `typeName.isOptional`
+- `isTuple` <- shorthand for `typeName.isTuple`
 
 </details>
 
--- a/README.md
+++ b/README.md
@@ -203,19 +203,22 @@ For each type you can access following properties:
 **Enum.Case.AssociatedValue** provides:
 
 - `name` <- name
-- `typeName` <- name of type of associated value
-- `unwrappedTypeName` <- returns name of the type, unwrapping the optional e.g. for variable with type `Int?` this would return `Int`
-- `isOptional` <- whether is optional
+- `typeName` <- name of type of associated value (*TypeName*)
+- `unwrappedTypeName` <- shorthand for `typeName.unwrappedTypeName`
+- `isOptional` <- shorthand for `typeName.isOptional`
+- `isTuple` <- shorthand for `typeName.isTuple`
 
 **Variable** provides:
 
 - `name` <- Name
 - `type` <- type of the variable, if known
-- `typeName` <- returns name of the type, including things like optional markup
-- `unwrappedTypeName` <- returns name of the type, unwrapping the optional e.g. for variable with type `Int?` this would return `Int`
-- `isOptional` <- whether is optional
+- `typeName` <- returns name of the type (*TypeName*)
+- `unwrappedTypeName` <- shorthand for `typeName.unwrappedTypeName`
+- `isOptional` <- shorthand for `typeName.isOptional`
 - `isComputed` <- whether is computed
 - `isStatic` <- whether is static variable
+- `isTuple` <- shorthand for `typeName.isTuple`
+- `typeName.tuple` <- returns information about tuple type
 - `readAccess` <- what is the protection access for reading?
 - `writeAccess` <- what is the protection access for writing?
 - `annotations` <- dictionary with configured [annotations](#source-annotations)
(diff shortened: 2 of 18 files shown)


Sourcery/Models/TypeName.swift
```diff
@@ -26,27 +26,35 @@ protocol Typed {
...
     var unwrappedTypeName: String { return typeName.unwrappedTypeName }
+    var actualTypeName: TypeName? { return typeName.actualTypeName }
+    var isTuple: Bool { return typeName.isTuple }
 }
 
 // sourcery: skipDescription
 final class TypeName: NSObject, AutoDiffable {
     let name: String
 
+    /// Actual type name if given type name is type alias
```
krzysztofzablocki: so how is this going to happen now? I think that before this PR TypeName was never used for typealiases

Sourcery/Models/TypeName.swift
```diff
@@ -59,13 +67,120 @@ final class TypeName: NSObject, AutoDiffable {
...
+        init(name: String, typeName: String, type: Type? = nil) {
+            self.name = name
+            self.typeName = TypeName(typeName)
+            self.type = type
+        }
+    }
+
+    lazy private(set) var elements: [Element] = {
+        let trimmedBracketsName = String(self.name.characters.dropFirst().dropLast())
+        return trimmedBracketsName
```
krzysztofzablocki: I really think Parsing should happen in the `FileParser` or one of associated files rather than model objects

Sourcery/Parsing/FileParser.swift
```diff
@@ -344,20 +344,21 @@ extension FileParser {
...
-                print("\(logPrefix)parseEnumAssociatedValues: Unknown enum case body format \(body)")
-                return nil
-            }
+        let items = body.commaSeparated()
+        return items
+            .map({ $0.trimmingCharacters(in: .whitespaces) })
+            .enumerated()
+            .map {
+                let nameAndType = $1.colonSeparated().map({ $0.trimmingCharacters(in: .whitespaces) })
+                let defaultName: String? = $0 == 0 && items.count == 1 ? nil : "\($0)"
```
krzysztofzablocki: why not use switch? and can you explain what is that ternary logic suppose to do here?

Sourcery/Parsing/FileParser.swift
```diff
@@ -344,20 +344,21 @@ extension FileParser {
...
+
+                guard nameAndType.count == 2 else {
+                    return Enum.Case.AssociatedValue(name: defaultName, typeName: $1)
```
krzysztofzablocki: what use case is this correct in?

Sourcery/StencilTemplate.swift
```diff
@@ -52,6 +57,18 @@ final class StencilTemplate: Stencil.Template, Template {
     }
 }
 
+extension Stencil.Extension {
+
+    func registerFilterWithArguments<A>(_ name: String, filter: @escaping (Any?, A) throws -> Any?) {
+        registerFilter(name) { (any, args) throws -> Any? in
+            guard args.count == 1, let arg = args.first as? A else {
+                throw TemplateSyntaxError("'\(name)' filter takes a single string argument")
```
krzysztofzablocki: you should use the A type not string in this message

Sourcery/StencilTemplate.swift
```diff
@@ -29,13 +29,18 @@ final class StencilTemplate: Stencil.Template, Template {
...
+        ext.registerFilter("upperFirst", filter: Filter<String>.make({ $0.upperFirst() }))
+        ext.registerFilterWithArguments("contains", filter: Filter<String>.make({ $0.contains($1) }))
+        ext.registerFilterWithArguments("hasPrefix", filter: Filter<String>.make({ $0.hasPrefix($1) }))
+        ext.registerFilterWithArguments("hasSuffix", filter: Filter<String>.make({ $0.hasSuffix($1) }))
 
         ext.registerFilter("computed", filter: Filter<Variable>.make({ $0.isComputed && !$0.isStatic }))
         ext.registerFilter("stored", filter: Filter<Variable>.make({ !$0.isComputed && !$0.isStatic }))
+        ext.registerFilter("tuple", filter: Filter<Variable>.make({ $0.isTuple }))
+
+        ext.registerFilterWithArguments("based", filter: FilterOr<Type, Typed>.make({ $0.based[$1] != nil }, other: { $0.type?.based[$1] != nil }))
```
krzysztofzablocki: 👍

Sourcery/Models/TypeName.swift
ilyapuchka: Before we had only `typeName` and `type`, and type will be set as actual type even when type is an alias, but only if it is an alias for known type. It will not work in case of types that are not `Type` at all, like closures and tuples. So when resolving type we now not just set the `type` property, but also the name of the actual type. Nothing else changes.
I think it makes sense to allow access to both type names so user can decide what to use - what is used in code, or what will be used in runtime.

Also in future we should improve how we handle typealiases for optional types, i.e. `typealias Handler = (()->())?`. Not saying that this is a "good" practice, but it's just valid Swift, so maybe we should support that too, or just agree to ignore that case for now.

Sourcery/Models/TypeName.swift
ilyapuchka: Yeah, I will try to move that to Parser. Actually I even tried to make `Enum.Case.AssociatedValue` an alias for `TupleType.Element` as in fact they are tuples, but have to workaround the fact that tuple of single item is not considered a tuple in this pr.

Sourcery/Parsing/FileParser.swift
ilyapuchka: In case of single associated value it will not have a name. Even though we can set it, but we can not access it when we bind it. In this case the name will be nil, otherwise if there is no name, or it is `_` we use index.

Btw I think we should change it, because though we can not access single associated value by name when binding, we must use it when creating an enum:

```swift
enum Some { case value(v: Int) }
if case let .value(a) = Some.value(v: 1) { }
```

It looks like it should be handled like method parameters names (argument label + parameter name)

Sourcery/Parsing/FileParser.swift
ilyapuchka: `case value(first: Int, Int)`, it's the case of the second associated value, `nameAndType` will have count 1 and well contain just `Int`. `defaultName` will be "1" in this case.

Sourcery/Parsing/FileParser.swift
krzysztofzablocki: do you feel comfortable tweaking that in this PR or should we do it later on separately?

Sourcery/Parsing/FileParser.swift
krzysztofzablocki: thanks for explaining :)

Sourcery/Parsing/FileParser.swift
ilyapuchka: I'll try to do it here, if it will take too much of an effort I let you know, but should be not hard to fix

krzysztofzablocki:
Anything I can help you with in this pr ? I want to get this one and #95 merged before I do PR for the #96 / #97

ilyapuchka:
I'll finish this tonight

Ilya Puchka

Отправлено с iPhone

Sourcery/Parsing/FileParser.swift
ilyapuchka: @krzysztofzablocki any ideas what names better to use for that? `localName` and `externalName`?

To better illustrate the issue here are few cases:

```swift
enum Some {
  case option0(Int)
  case option1(v: Int)
  case option2(v: Int, Int)
}

let opt1 = Some.option0(1) //no local name
let opt2 = Some.option1(v: 1) //has local name
let opt3 = Some.option2(v: 1, 2) //first has local name, second has no local name

switch some {
  case let .option0(a): print(a) // has no external name
  case let .option1(a): print(a) // has no external name
  case let .option2(a): print(a.first); print(a.1) //first has external name, second has index name
}
```

krzysztofzablocki: would we have 3 different properties? `local` `external` and `index` ?

ilyapuchka: yes, it also makes sense to have `indexName` as named elements are also accessible by their index

krzysztofzablocki: ok, if external name automatically gets the indexName in case nothing else is available (callsite convenience) this looks reasonable 👍

krzysztofzablocki:
ready for review?

ilyapuchka:
Yes

CHANGELOG.md
```diff
@@ -6,9 +6,15 @@
 ### New Features
 - Variables with default initializer are now supported, e.g. `var variable = Type(...)`
 - Added support for special escaped names in enum cases e.g. `default` or `for`
+- Added support for tuple types and `tuple` filter for variables
+- Enum associated values now have `localName` and `exteranlName` properties.
```
krzysztofzablocki: typo `externalName`

README.md
```diff
@@ -187,73 +195,113 @@ For each type you can access following properties:
...
+- `isTuple` <- whether given type is a tuple
+- `tuple.elements` <- if given type is a tuple returns its elements information (*TupleType.Element*)
+
+</details>
+
+<details><summary>**TupleType.Element**. Properties:</summary>
+
+- `name` <- element name
+- `typeName` <- element type name (*TypeName*)
+- `type` <- type of element, if known
```
krzysztofzablocki: missing the shorthands for optional and unwrapped

Sourcery/Models/TypeName.swift
```diff
@@ -54,13 +56,52 @@ final class TypeName: NSObject, AutoDiffable {
...
+        } else {
+            return name.isValidTupleName()
+        }
+    }
+
+    var tuple: TupleType?
+
     override var description: String {
-        return name
+        if let actualTypeName = actualTypeName {
```
krzysztofzablocki: I don't think this should be doing anything like `aka`, this isn't debugDescription, the reason I've made it return `name` is because it's convienent to allow users to render it instead of having to reach through to name. So it should what you'd like to generate on callsite

Sourcery/Parsing/Utils/Extensions.swift
```diff
@@ -0,0 +1,57 @@
...
+    func commaSeparated() -> [String] {
+        return components(separatedBy: ",", excludingDelimiterBetween: ("<(", ")>"))
+    }
+
+    /// Returns components separated with colon respecting nested types
+    func colonSeparated() -> [String] {
+        return components(separatedBy: ":", excludingDelimiterBetween: ("<[(", ")]>"))
+    }
+
+    fileprivate func components(separatedBy delimiter: Character, excludingDelimiterBetween between: (open: String, close: String)) -> [String] {
```
krzysztofzablocki: awesome, I think I can use this to solve one bug I noticed :)

ilyapuchka:
Fix your comments

PR #94, opened 2017-01-02
Update README

bolismauro:
Some properties listed in `Variable` are in fact nested inside `TypeName`.

I'm not sure whether is something transitory or it is supposed to stay. Anyway this PR updates the README with the nesting.

(I think that `typeName` is not the most appropriated name for that variable. )

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,14 +211,18 @@ For each type you can access following properties:
 
 - `name` <- Name
 - `type` <- type of the variable, if known
-- `typeName` <- returns name of the type, including things like optional markup
-- `unwrappedTypeName` <- returns name of the type, unwrapping the optional e.g. for variable with type `Int?` this would return `Int`
-- `isOptional` <- whether is optional
+- `typeName` <- returns information about the type
 - `isComputed` <- whether is computed
 - `isStatic` <- whether is static variable
 - `readAccess` <- what is the protection access for reading?
 - `writeAccess` <- what is the protection access for writing?
 - `annotations` <- dictionary with configured [annotations](#source-annotations)
+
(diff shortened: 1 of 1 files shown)


krzysztofzablocki:
yeah I agree, it's better if we avoid repetitions in the docs e.g. one that happened in #92

Do you have suggestions for `typeName` alternative naming? /cc @ilyapuchka

ilyapuchka:
Actually I have similar changes to README in #92 but instead of removing info that isOptional and so on can be accessed directly on the variable and other types I mention that it is a shorthand to typeName property.

bolismauro:
I think it is fine too :)
In general I'd expect to find that information at the `variable` level and not tested.
I can't find a proper name because I can't find a rationale that separates information that is in `typeName` from the ones that is not.

I guess it is because in `typeName` there are all the properties that can be computed from the name

krzysztofzablocki:
so what do you guys think we should do here in regards to changes in #92, @bolismauro @ilyapuchka which doc approach you prefer?

bolismauro:
I'd keep changes in #92
The main reason is because `typeName.isOptional` is very weird to see. As I wrote, it is not clear why a variable named "typeName" should have a value "isOptional". The shorthand is much easier to use and to understand

ilyapuchka:
What about hiding these details about properties in <detail>?

bolismauro:
From a user perspective I think it is better (less is more, having the same thing in multiple points may be confusing)

krzysztofzablocki:
ok let's do that, ok to close this pr @bolismauro?

bolismauro:
Sure 👍

PR #95, opened 2017-01-02
Swift templates proof of concept.

kzaher:
Hi,

this PR adds a proof of concept for Swift generator.

It was the simplest implementation I could come up with for now. It works by serializing metadata from sourcery in binary form and deserializing it in compiled template binary.

This approach can be drastically improved in future:
- injecting compiled code into sourcery so no serialization would be needed
- adding a caching layer so swift template can be compiled only once
- adding proper source mapping 
- adding some more swift template examples, etc ...

The ugliest part was that I've had to manually copy generated serialization code to original source files. This code can be removed in future and it's just temporary. If Sourcery will be packed as framework in future, then we can remove that serialization code entirely.

I didn't want to update documentation because it would probably be more suitable for project owners to update it where they feel appropriate.


Sourcery/CodeGenerated/Coding.generated.swift
```diff
@@ -0,0 +1,354 @@
+// Generated using Sourcery 0.4.9 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+
+import Foundation
+
+extension NSCoder {
+
+    @nonobjc func decode(forKey: String) -> Int {
+        // swiftlint:disable force_cast
```
krzysztofzablocki: the force_cast seem unnecesary?

Sourcery/Generating/Template/Swift/GenerationContext.swift
```diff
@@ -0,0 +1,169 @@
...
+        self.arguments = aDecoder.decode(forKey: "arguments")
+    }
+
+    func encode(with aCoder: NSCoder) {
+        aCoder.encode(self.types, forKey: "types")
+        aCoder.encode(self.typeByName, forKey: "typeByName")
+        aCoder.encode(self.arguments, forKey: "arguments")
+    }
+
+    override func isEqual(_ object: Any?) -> Bool {
```
krzysztofzablocki: can't we just generate this one?

Sourcery/Generating/Template/Swift/SwiftTemplate.swift
```diff
@@ -0,0 +1,199 @@
...
+
+    return sourceFile.joined(separator: "")
+}
+
+struct ProcessResult {
+    let output: String
+    let error: String
+}
+
+private extension Path {
```
krzysztofzablocki: we could create a simple extension that both tests and sourcery can reuse, the code is the same outside of the path component which we can parametrize

Sourcery/Generating/Template/Swift/SwiftTemplate.swift
```diff
@@ -0,0 +1,199 @@
...
+    case unmatchedOpening(path: Path, line: Int)
+    case compilationError(path: Path, error: String)
+}
+
+enum Delimiters {
+    static let open = "<%"
+    static let close = "%>"
+}
+
+func generateSwiftCode(templateContent _templateContent: String, path: Path) throws -> String {
```
krzysztofzablocki: should we mark them private? also any reason this isn't part of SwiftTemplate class?

Sourcery/Models/Method.swift
```diff
@@ -100,4 +119,34 @@ final class Method: NSObject, AutoDiffable {
...
+        self.selectorName = aDecoder.decode(forKey: "selectorName")
+        self.parameters = aDecoder.decode(forKey: "parameters")
+        self.returnTypeName = aDecoder.decode(forKey: "returnTypeName")
+        self.returnType = aDecoder.decode(forKey: "returnType")
+        self.accessLevel = aDecoder.decode(forKey: "accessLevel")
+        self.isStatic = aDecoder.decodeBool(forKey: "isStatic")
+        self.isClass = aDecoder.decodeBool(forKey: "isClass")
+        self.isFailableInitializer = aDecoder.decodeBool(forKey: "isFailableInitializer")
+        self.annotations = aDecoder.decode(forKey: "annotations")
+        self.__parserData = aDecoder.decode(forKey: "__parserData")
```
krzysztofzablocki: you should mark __parserData not codable, no point in serializing that, it's only used in parsing stage

Sourcery/Models/Type.swift
```diff
@@ -193,4 +193,55 @@ class Type: NSObject, AutoDiffable {
...
+        aCoder.encode(self.inheritedTypes, forKey: "inheritedTypes")
+        aCoder.encode(self.containedTypes, forKey: "containedTypes")
+        aCoder.encode(self.parentName, forKey: "parentName")
+        aCoder.encode(self.parent, forKey: "parent")
+        aCoder.encode(self.supertype, forKey: "supertype")
+        aCoder.encode(self.__parserData, forKey: "__parserData")
+    }
+}
+
+extension Type {
```
krzysztofzablocki: this can be removed, we now have dedicated `Class` object

krzysztofzablocki:
@ilyapuchka take a look as this one is important

ilyapuchka:
To be honest I don't think I get it how it works except on a very general level =) Will need to look at it closer and run it manually. Will do this weekend.

krzysztofzablocki:
First we parse data as usual, then we persist the data to disk.
We create a temporary Swift file that contains the `.swifttemplate` contents along with some setup boilerplate code that's responsible for loading that parsed data back from disk and making it available to template code. We compile that final file using `swiftc` (it included all our model symbols) and execute it to render the final content

CHANGELOG.md
```diff
@@ -3,6 +3,22 @@
 ---
 ## 0.5.1
 
+- Swift Templates are now supported
+
+```
+<% for type in types.classes { %>
+    extension <%= type.name %>: Equatable {}
```
ilyapuchka: What about using string interpolation style?

Sourcery/Generating/Template/Swift/GenerationContext.swift
```diff
@@ -0,0 +1,179 @@
...
+    /// Lists all known classes in the project
+    /// sourcery: skipEquality
+    lazy var classes: [Class] = {
+        return self.types.classes
+    }()
+
+    /// lists all known types, excluding protocols
+    /// sourcery: skipEquality
+    lazy var all: [Type] = {
+        return self.types.all
```
ilyapuchka: In current version we exclude protocols, I guess should do the same here?

Sourcery/Generating/Template/Swift/SwiftTemplate.swift
```diff
@@ -0,0 +1,192 @@
...
+    let sourcePath: Path
+
+    init(path: Path) throws {
+        self.sourcePath = path
+    }
+
+    fileprivate static func generateSwiftCode(templateContent _templateContent: String, path: Path) throws -> String {
+        let templateContent = "<%%>" + _templateContent
+
+        let components = templateContent.components(separatedBy: Delimiters.open)
```
ilyapuchka: If delimiter is used for example in a string literal, do we escape it somehow?

Sourcery/Generating/Template/Swift/SwiftTemplate.swift
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        return sourceFile.joined(separator: "")
+    }
+
+    func render(types: [Type], arguments: [String: NSObject]) throws -> String {
+        let context = GenerationContext(types: types, arguments: arguments)
+        let swiftCode = try SwiftTemplate.generateSwiftCode(templateContent: try sourcePath.read(), path: sourcePath)
+
+        let compilationDir = Path.cleanTemporaryDir(name: "build")
```
ilyapuchka: Will this clean build folder of the target that uses Sourcery? Wouldn't that lead to Xcode recompiling it every time completely?

Sourcery/Generating/Template/Swift/SwiftTemplate.swift
```diff
@@ -0,0 +1,192 @@
...
+
+        if !compilationResult.error.isEmpty {
+            #if DEBUG
+                let command = "/usr/bin/swiftc " + arguments.map { "\"\($0)\"" }.joined(separator: " ")
+                print(command)
+            #endif
+            throw SwiftTemplateParsingError.compilationError(path: binaryFile, error: compilationResult.error)
+        }
+
+        let result = try Process.runCommand(path: binaryFile.description,
```
ilyapuchka: So if I get it correctly we encode context with all types information, then we pass the path to this archive to another swift file in which we insert code that has `<%= %>` replaced with prints and that decodes the archive and runs the printing code we inserted in it. And then we get the output of all these prints as a result. Right?
Do we somehow handle the issue with empty lines when removing control code (https://github.com/kylef/Stencil/pull/32) or we don't have it at all?

Sourcery/CodeGenerated/Coding.generated.swift
```diff
@@ -0,0 +1,397 @@
...
+        guard let object = self.decodeObject(forKey: forKey) else {
+            return nil
+        }
+
+        return object as? E
+    }
+
+}
+
+/*
```
ilyapuchka: Why all that commented out? It does not work? Do we miss something to be able to generate that code properly?

Sourcery/Generating/Template/Swift/SwiftTemplate.swift
krzysztofzablocki: no, we are using NSTemporaryDirectory for each run, completely unrelated to sourcery or users folders

Sourcery/CodeGenerated/Coding.generated.swift
krzysztofzablocki: @kzaher uses a script to copy this into each class until we have inline code generation as part of sourcery

CHANGELOG.md
kzaher: Hi @ilyapuchka ,

We need to use something obscure to have the least possible impact on Swift grammar and for it visually stand out from regular swift for ease of human template parsing.

There are two most common families of template grammars. One that use `{{}}`, and one that use `<%%>`.

`{{}}` is more suitable for languages that don't have `{}` as a part of their grammar (HTML).

I've found that `<%` is typically used for template engines that generate C family of languages so I've just used what was proven and common to me.

Sourcery/Generating/Template/Swift/SwiftTemplate.swift
kzaher: We don't

Sourcery/Generating/Template/Swift/SwiftTemplate.swift
kzaher: I'm unsure what is exactly that referenced issue. I'm assuming it doesn't have that issue.

Sourcery/Generating/Template/Swift/GenerationContext.swift
kzaher: This is fixed, I've made `types.all` perform the filtering.

Sourcery/Generating/Template/Swift/SwiftTemplate.swift
ilyapuchka: Yeah, I guess as instead of removing control code we use print we don't have that exact issue, but then how is indentation handled?

kzaher: Hi @ilyapuchka ,

there is no magic behavior. It should be simple to deduce how your result template will look like according to white spaces in template since all non code pars will be escaped and printed.

krzysztofzablocki:
@kzaher can you rebase on top of master (2 conflicts)? I'll merge it once that's done

krzysztofzablocki:
merged 👍

PR #101, opened 2017-01-04
Add support for special escaped enum cases

krzysztofzablocki:
This will allow us to parse `default` `for` etc cases


ilyapuchka:
Should we also do the same for any variable, parameter name, tuple element etc.?

krzysztofzablocki:
We should but I wasn't sure how exactly we need to handle it, we need to think about whether we need to store both names or just unwrapped

PR #106, opened 2017-01-07
Parallelize parsing

krzysztofzablocki:


krzysztofzablocki:
@ilyapuchka @kzaher @bolismauro could you run this pr version against your code base few times and see if you notice any issues/crashes ? there shouldn't be any but I'd rather get more test before merging

bolismauro:
Everything seems fine here

If you are interested: the generation time is ~40% lower with respect to "master"

krzysztofzablocki:
@bolismauro pushed one commit, it should be even faster now

PR #110, opened 2017-01-08
Parsing attributes

### LINKED ISSUE #103: Better support for escaping function types
In some use-cases, it would be useful to get a typeName of a function that includes the `@escaping` attribute, and one that doesn't include it.

For instance, when assigning an escaping function parameter to a variable outside of that function. The parameter `typeName` has to be `@escaping (String)-> Void`, but in variables and tuples, functions are implicitly escaping, and thus don't require the `@escaping` annotation.

```
var onComplete: (String)-> Void

func someFunc(onComplete: @escaping (String)-> Void) {
  self.onComplete = onComplete
}
```

The `typeName` for these two is different, in the function parameter it's `@escaping (String)-> Void`, and in the variable it's just `(String)-> Void`.

It would be nice to expose `nonEscapingTypeName` in `Parameter`.

Another way to handle this would be by adding a custom `replaceSubrange` filter for Stencil strings, although that would be a separate issue.

ilyapuchka:
Resolves #103

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,7 @@ Available types:
 - `based.BaseClassOrProtocol` => info whether type implements or inherits from `BaseClassOrProtocol` (all type names encountered, even those that Sourcery didn't scan)
 - `containedTypes` <- list of types contained within this type
 - `parentName` <- list of parent type (for contained ones)
+- `attributes` <- type attributes, i.e. `type.attributes.objc`
 - `annotations` <- dictionary with configured [annotations](#source-annotations)
 
 </details>
@@ -242,6 +243,7 @@ Available types:
 - `typeName.tuple` <- returns information about tuple type
 - `readAccess` <- what is the protection access for reading?
 - `writeAccess` <- what is the protection access for writing?
+- `attributes` <- variable attributes, i.e. `var.attributes.NSManaged`
 - `annotations` <- dictionary with configured [annotations](#source-annotations)
 
 </details>
@@ -261,6 +263,7 @@ Available types:
 - `isClass` <- whether method is class (can be overriden by subclasses)
 - `isInitializer` <- whether method is an initializer
 - `isFailableInitializer` <- whether method is failable initializer
+- `attributes` <- method attributes, i.e. `method.attributes.discardableResult`
 - `annotations` <- dictionary with configured [annotations](#source-annotations)
 
 </details>
@@ -276,6 +279,7 @@ Available types:
 - `isOptional` <- shorthand for `typeName.isOptional`
 - `isImplicitlyUnwrappedOptional` <- shorthand for `typeName. isImplicitlyUnwrappedOptional `
 - `isTuple` <- shorthand for `typeName.isTuple`
+- `typeAttributes` <- parameter's type attributes, shorthand for `typeName.attributes`, i.e. `param.typeAttributes.escaping`
 
 </details>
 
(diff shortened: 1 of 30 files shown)


README.md
```diff
@@ -289,6 +293,7 @@ Available types:
 - `isVoid` <- whether type is Void (`Void` or `()`)
 - `isTuple` <- whether given type is a tuple
 - `tuple.elements` <- if given type is a tuple returns its elements information (*TupleType.Element*)
+- `attributes` <- type attributes, i.e. `typeName.attributes.escaping`, or `typeName.escaping`
```
krzysztofzablocki: I think we are better off wrapping it in `attributes` rather than exposing it via shorthand. I'd rather avoid the possibility of nameclash

Sourcery/Parsing/Composer.swift
```diff
@@ -59,7 +59,7 @@ struct Composer {
 
         let resolveType = { (typeName: TypeName, containingType: Type?) -> Type? in
             let name = self.typeName(for: typeName.unwrappedTypeName, containingType: containingType, typealiases: typealiases)
-            typeName.actualTypeName = name.flatMap(TypeName.init)
+            typeName.actualTypeName = name.flatMap({ TypeName($0) })
```
krzysztofzablocki: why change this? partial application is nice

Sourcery/Parsing/FileParser.swift
```diff
@@ -453,17 +489,93 @@ extension FileParser {
...
-    fileprivate func isGeneric(source: [String: SourceKitRepresentable]) -> Bool {
-        guard let substring = extract(.nameSuffix, from: source), substring.trimmingCharacters(in: .whitespacesAndNewlines).hasPrefix("<") == true else { return false }
-        return true
+}
+
+// MARK: - Attributes
+extension FileParser {
+
+    internal func parseDeclarationAttributes(_ source: [String: SourceKitRepresentable]) -> [String: Attribute] {
+        guard let prefix = extract(Substring.keyPrefix, from: source) else { return [:] }
```
krzysztofzablocki: just `.keyPrefix`

Sourcery/Templates/TypedSpec.stencil
```diff
@@ -0,0 +1,35 @@
+import Quick
+import Nimble
+@testable import Sourcery
+
+class TypedSpec: QuickSpec {
```
krzysztofzablocki: nice :+1:

Sourcery/Parsing/Composer.swift
ilyapuchka: It does not seem to work with parameters with default values =(

README.md
ilyapuchka: So only remove KVC shorthand for TypeName, but keep for other types?

ilyapuchka: So only remove KVC shorthand for TypeName, but keep for other types?

krzysztofzablocki: I think we should always allow only accessing those via `attributes` rather than doing the trick of `variable.escaping`, it's not a big deal to write it in the template and it's more consistent and avoids potential clashes

Sourcery/Parsing/Composer.swift
krzysztofzablocki: oh interesting, this is fine the way you have it but didn't expect it :)

README.md
ilyapuchka: The reason why I added that is that most of the attributes feel like boolean flags on attributed instances. But yes, skipping `attribute` is not a huge win here.

PR #112, opened 2017-01-10
Improve inferring types of variables with initial value

### LINKED ISSUE #111: Improve default value type inferrence
If I have a type like this

```swift
struct B {
 let d = true
}
```

there is no way in the stencils to know that `d` exists (at least, I haven't found it).

Now, of course it is not possible to know that `d` is a `Bool`, since Sourcery currently doesn't have information about things that it doesn't scan. I think that `d` should be listed though. Maybe with an empty type?
Otherwise it is impossible to add errors in the code.

Use case: I have this stencil for auto equatable, and if developers forget to use the proper syntax, there is not easy way (besides review the generated code) to check that all the variables have been correctly handled. My idea is to add either a comment, or something that make the code fail to compile. It is a safer apporach

ilyapuchka:
Resolves #111 

Here I've tried approach described in https://github.com/jpsim/SourceKitten/issues/240 with passing "-j4" argument to `SwiftDocs`. This way we get information for inferred literal types. 

But there are some limitations. As we are getting docs for a single file it does not infer literals (i.e. array) that uses type defined in another file (instead it has `<<error type>>`). So for example if we have `var some = ["a": TypeName("a")]` property in `Type` it will fail to infer type correctly. We also process it manually not correctly, so the type will be `[\"\": TypeName`.

I think we can try to solve that by compiling all the files, but I will need to investigate how `sourcekitten` does that as in `SourceKittenFramework` there are only methods to get docs for single file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### New Features
 - Added a new filter `replace`. Usage: `{{ name|replace:"substring","replacement" }}` - replaces occurrences of `substring` with `replacement` in `name` (case sensitive)
+- Improved support for inferring types of variables with initial values
 - Sourcery will now use parallel parsing, expect more than 2x as fast execution.
 - Sourcery will now cache source artifacts, in many scenarios it will lead to order of magnitude faster processing.
 - e.g. on big codebase of over 300 swift files:
--- a/Sourcery/Generating/Template/Stencil/StencilTemplate.swift
+++ b/Sourcery/Generating/Template/Stencil/StencilTemplate.swift
@@ -142,7 +142,7 @@ extension String {
 
     fileprivate func upperFirst() -> String {
         let first = String(characters.prefix(1)).capitalized
-        let other = String(characters.dropFirst())
+        let other = dropFirst()
         return first + other
     }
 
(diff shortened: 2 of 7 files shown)


krzysztofzablocki:
I'm not sure if we should attempt this, I don't want us to parse everything together because it will prevent us from parallelization and caching. I don't want this project to have performance implication or regular swift complication (which is awful). 

Users can define type on variable definition if they need to, it's not a big ask for all the benefits we get

krzysztofzablocki:
How is the performance test affected by including docs parsing?

ilyapuchka:
Absolutely agree and I'm also not completely sold on should we use that at all, as of course performance decreases as we have basically to additionally compile each file separately. I will run measure tests later to post results, but from what I noticed from running against Sourcery models itself it's now 1,9 s and was around 1 s before.
If we don't do that I would still like to fix the parsing of literal types and as you mentioned make `TypeName` optional (or we can reuse `<<error type>>`) so that we don't skip such variables.

krzysztofzablocki:
Agreed on fixing the parsing of literal type, and I think having <<uknown type, please add type attribution to variable>> would be best as it would make uncompilable code in situation where user tried to use it and also give clear info what to do

ilyapuchka:
Agree. I think we can also include declaration in the unknown type message (if it's very easy to get it with what we have now), to provide more context:
`<<unknown type, please add type attribution to variable 'let some = true'>>`

krzysztofzablocki:
Can you rebase on top of master? I'll try to review this today after work

PR #114, opened 2017-01-10
Add filter named `replace` to substitute one substring with another

sibljon:
According to the contribution guideline, I was supposed to open an Issue to discuss this first, but since this is something I need in my own project and I needed it now, it seemed simplest to just discuss in this Pull Request. That being said, feel free to not accept this, of course! I'm also happy to make changes, as desired.

### DIFF
--- a/Sourcery/Generating/Template/Stencil/StencilTemplate.swift
+++ b/Sourcery/Generating/Template/Stencil/StencilTemplate.swift
@@ -30,6 +30,10 @@ final class StencilTemplate: Stencil.Template, Template {
     private static func sourceryEnvironment() -> Stencil.Environment {
         let ext = Stencil.Extension()
         ext.registerFilter("upperFirst", filter: Filter<String>.make({ $0.upperFirst() }))
+        ext.registerFilterWithTwoArguments("replace", filter: { (source: String, substring: String, replacement: String) -> Any? in
+            return source.replacingOccurrences(of: substring, with: replacement)
+        })
+        
         ext.registerBoolFilterWithArguments("contains", filter: { (s1: String, s2) in s1.contains(s2) })
         ext.registerBoolFilterWithArguments("hasPrefix", filter: { (s1: String, s2) in s1.hasPrefix(s2) })
         ext.registerBoolFilterWithArguments("hasSuffix", filter: { (s1: String, s2) in s1.hasSuffix(s2) })
@@ -85,6 +89,16 @@ extension Annotated {
 }
 
 extension Stencil.Extension {
+    
+    func registerFilterWithTwoArguments<T, A, B>(_ name: String, filter: @escaping (T, A, B) throws -> Any?) {
+        registerFilter(name) { (any, args) throws -> Any? in
+            guard let type = any as? T else { return any }
+            guard args.count == 3, let argA = args[0] as? A, let argB = args[2] as? B else {
+                throw TemplateSyntaxError("'\(name)' filter takes two arguments: \(A.self) and \(B.self)")
+            }
+            return try filter(type, argA, argB)
+        }
+    }
 
     func registerFilterWithArguments<A>(_ name: String, filter: @escaping (Any?, A) throws -> Any?) {
         registerFilter(name) { (any, args) throws -> Any? in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Sourcery CHANGELOG
 
 ---
+## Master
+
+### New Features
+- Added a new filter `replace`. Usage: `{{ name|replace:"substring","replacement }}` - replaces occurrences of `substring` with `replacement` in `name` (case sensitive)
+
+### Internal Changes
+- N/A
+
 ## 0.5.2
 
 ### New Features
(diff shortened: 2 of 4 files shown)


Sourcery/Generating/Template/Stencil/StencilTemplate.swift
```diff
@@ -30,6 +30,13 @@ final class StencilTemplate: Stencil.Template, Template {
     private static func sourceryEnvironment() -> Stencil.Environment {
         let ext = Stencil.Extension()
         ext.registerFilter("upperFirst", filter: Filter<String>.make({ $0.upperFirst() }))
+        ext.registerFilterWithTwoArguments("replace", filter: { (s1: Any?, s2: Any, s3: Any) -> Any? in
+            if let s1 = s1 as? String, let s2 = s2 as? String, let s3 = s3 as? String {
+                return s1.replacingOccurrences(of: s2, with: s3)
+            }
+            return s1
```
sibljon: I wasn't sure what's the best way to fail if the arguments are of the wrong type (not Strings), so I defaulted to just passing through the value.

Sourcery/Generating/Template/Stencil/StencilTemplate.swift
```diff
@@ -85,6 +92,15 @@ extension Annotated {
...
 extension Stencil.Extension {
+    
+    func registerFilterWithTwoArguments<A, B>(_ name: String, filter: @escaping (Any?, A, B) throws -> Any?) {
+        registerFilter(name) { (any, args) throws -> Any? in
+            guard args.count == 3, let argA = args[0] as? A, let argB = args[2] as? B else {
+                throw TemplateSyntaxError("'\(name)' filter takes two arguments: \(A.self) and \(B.self)")
+            }
+            return try filter(any, argA, argB)
+        }
+    }
```
sibljon: I'm also happy to make changes, as desired. I tried to follow suit by wrapping this in a `registerStringFilterWithTwoArguments`, but I couldn't quite get Swift compiler errors, so I left it as-is. I also wasn't sure what's the best way to fail

Sourcery/Generating/Template/Stencil/StencilTemplate.swift
ilyapuchka: Basically we should throw error if parameters do not validate. Like in `registerFilterWithArguments`.

Sourcery/Generating/Template/Stencil/StencilTemplate.swift
krzysztofzablocki: You can simplify your code:
```swift
func registerFilterWithTwoArguments<T, A, B>(_ name: String, filter: @escaping (T, A, B) throws -> Any?) {
        registerFilter(name) { (any, args) throws -> Any? in
            guard let type = any as? T else { return any } 
            guard args.count == 3, let argA = args[0] as? A, let argB = args[2] as? B else {
                throw TemplateSyntaxError("'\(name)' filter takes two arguments: \(A.self) and \(B.self)")
            }
            return try filter(type, argA, argB)
        }
   }
```

Then at your call site you can get rid of any guess work
```swift
 ext.registerFilterWithTwoArguments("replace", filter: { (source: String, substring: String, replacement: String) -> Any? in
              return source.replacingOccurrences(of: substring, with: replacement)
         })
```

ilyapuchka: Can we reuse `registerFilterWithArguments` for that, using tuple as arguments type?

Sourcery/Generating/Template/Stencil/StencilTemplate.swift
krzysztofzablocki: but only parameters, we don't throw on the value type (and careful with optionality)

Sourcery/Generating/Template/Stencil/StencilTemplate.swift
krzysztofzablocki: if the cast will work that's even better 👍

sibljon:
> Thanks for contribution @sibljon !
> I think we might also add filters like appendPrefix/appendSuffix on top of that. What do you think @krzysztofzablocki

My pleasure!

As a matter of fact, that's also a use case of mine. Is there any advantage to having appendPrefix/appendSuffix as filters vs. `{{ foo }}SomeSuffix` ?

Also, it seems adding the ability to preserve camel casing would be important, though I'm not exactly sure how that would work. Is there a way to compound two filters together currently, or rather, to use the result of one filter as an input to another?

ilyapuchka:
@sibljon ah, nevermind about "appendPrefix"/"appendSuffix", it was stupid, just having it like you pointed is more than enough 😅

Speaking about combining filters yes, you can combine them, we do that for example when combining `implements`/`based`/etc with `count` filter

Sourcery/Generating/Template/Stencil/StencilTemplate.swift
sibljon: Good idea @krzysztofzablocki I just amended my commit to 

> Can we reuse registerFilterWithArguments for that, using tuple as arguments type?

Also a good idea, @ilyapuchka. However, I just gave this a shot and came up short :-/

krzysztofzablocki:
thanks for contributing 👍

PR #117, opened 2017-01-11
[WIP] Caching parsing artifacts

krzysztofzablocki:
This enables caching code artifacts.

Should we store cache in derived data? I think that would require users to pass that data in from their build phase

cc @bejar37 @kzaher @bolismauro take a look, I'd also appreciate some performance comparison if you can :)


krzysztofzablocki:
btw. @ilyapuchka I had to flatten the classes because NSCoding didn't play well with nested types

bolismauro:
@krzysztofzablocki about 83% faster with respect to master
How many files NY app has (I've read it takes less than 1 sec)

krzysztofzablocki:
@bolismauro a bit over 300 files

bejar37:
I'm seeing similar improvements to @bolismauro, 👍 !!

PR #121, opened 2017-01-13
feat: support throwing methods

### LINKED ISSUE #119: returnTypeName is Void for methods with throw
I've got a factory method on my object that is not being detected as a method returning a value. All other parameters I've tried to use fine but `returnTypeName` only shows me `Void`.

It's starting to feel like a bug on my side, since Sourcery obviously has unit tests for this. I'm using Swift templates if that matters

ilyapuchka:
Resolves #119


ilyapuchka:
Shut, forgot the case for throwing methods with no return value, it may even crash now.

krzysztofzablocki:
uh, do you think you'll have time today ? If not I can take a look after work

ilyapuchka:
already on it =)

PR #124, opened 2017-01-14
Preserve inherited types order and minor internal templates improvements

ilyapuchka:


Sourcery/Templates/Diffable.stencil
```diff
@@ -2,12 +2,13 @@
 extension {{ type.name }}{% if not type.supertype.implements.AutoDiffable %}: Diffable{% endif %} {
     {% if type.based.Type %}override {% endif %}func diffAgainst(_ object: Any?) -> DiffableResult {
         let results = DiffableResult()
-        {% for variable in type.storedVariables %}{% if not variable.annotations.skipEquality %}{% if forloop.first %}guard let rhs = object as? {{ type.name }} else {
+        guard object is {{ type.name }} else {
```
ilyapuchka: @krzysztofzablocki I remember you saying that it will be still good to check for type, even if there are no variables to compare. Without filters in `if` tag I could not come up with a better approach to avoid warnings for unused `rhs` (what was the reason for previous change in this template).

Sourcery/Templates/Description.stencil
```diff
@@ -1,9 +1,9 @@
-{% for type in types.classes %} {% if type.variables.count %}{% if not type.annotations.skipDescription %}
+{% for type in types.classes|!annotated:"skipDescription" %} {% if type.variables.count %}
 extension {{ type.name }} {
     override var description: String {
         var string = {% if type.inheritedTypes.first == "NSObject" %}"\(type(of: self)): "{% else %}super.description{% endif %}
-        {% for variable in type.variables|!annotated:"skipDescription" %}string += "{{ variable.name }} = \({{ variable.name }}){% if not forloop.last %}, {% endif %}"
+        {% for variable in type.variables|!annotated:"skipDescription" %}string += "{{ variable.name }} = \(self.{{ variable.name }}){% if not forloop.last %}, {% endif %}"
```
ilyapuchka: Need `self.` here because we now have ``` `throws` ``` variable but back ticks are not in the name. Might be a bug in variable parsing or something we can improve. Though using `self` is a valid workaround I think.

Sourcery/Models/Type.swift
```diff
@@ -245,6 +251,8 @@ class Type: NSObject, AutoDiffable, Annotated, NSCoding {
 }
 
 extension Type {
+
+    // sourcery: skipDescription
```
ilyapuchka: Don't think if this property is used anywhere, so maybe we don't need it at all.

Sourcery/Models/Type.swift
```diff
@@ -195,7 +195,13 @@ class Type: NSObject, AutoDiffable, Annotated, NSCoding {
         type.annotations.forEach { self.annotations[$0.key] = $0.value }
         type.inherits.forEach { self.inherits[$0.key] = $0.value }
         type.implements.forEach { self.implements[$0.key] = $0.value }
-        self.inheritedTypes = Array(Set(self.inheritedTypes + type.inheritedTypes))
+
+        var inheritedTypes = self.inheritedTypes
+        for inherited in type.inheritedTypes {
```
krzysztofzablocki: Why not use `NSMutableOrderedSet` like I did with flatten ?

ilyapuchka: In fact I think we don't even need that as Swift will not allow duplicated in inherited types in the first place

ilyapuchka:
@krzysztofzablocki I've also added an improvement for handling void throwing methods in protocols

Sourcery/Templates/Diffable.stencil
ilyapuchka: Just noticed that you already updated Stencil to the commit with filters support for if tag, so my previous comment is not relevant any more 😀

Issue #45, opened 2016-12-17
Inline generated code


ilyapuchka:
After watching a talk mentioned in #14 I think it will be crucial to be able to provide an option to put generated code right into source files instead of separate files. The problem is that in separate files we can not access private variables, so either parser should just ignore them (as I mentioned in #38) or generated code will not compile anyway if it uses private variables. So we will not be able to use them for example for description and other stuff.  Or we must enforce user to not use private variables which is not very nice too.

To achieve that we can simply always put generated code in the end of the file and separate it from other code with annotation comment (like "// sourcery: generated"). Then user will be just required to put his own code before that section, so that we don't override it. Feels like a reasonable tradeoff.

We might need to add a command line argument to define in what mode Sourcery should work. Then user can choose what templates should be used to generate code inline and what in separate files.
There should be also a separate collection of types accessible in templates like `filetypes` that will contain only types defined in that file for "inline" mode.

krzysztofzablocki:
It's an interesting option to add, but I think we have much more things to add before we should consider implementing this.

When we start to mutate original Source then I start getting a little more worried about potential implications, FWIW. you wouldn't be able to use `private` code in that case anyway, only `fileprivate`.

Once we get to more stable feature set then we can explore this feature, does that make sense?

ilyapuchka:
Sure, not even for 1.0 maybe.

krzysztofzablocki:
I think I'll think about this next because it would be convenient for things like NSCoding etc

krzysztofzablocki:
How do we feel about this syntax for both code and template:

```swift
// sourcery:inline:Type.AutoCoding
// sourcery:end
```

It would:
- Remove the body of that generated pattern while parsing (to avoid chicken-egg problem)
- Find the same pattern in one of the template generated code
  - Replace the source code with generated code 
  - [Question] Comment out or Remove the Template fragment?

ilyapuchka:
What do you mean by "comment or remove the template fragment"? Do you mean these annotations for source code?

krzysztofzablocki:
I wonder what should happen inside the output swift file (because there might be more code than just the inline fragments, whether we comment it out, or remove it completely when we put it inside the project source code

ilyapuchka:
You mean having this in template:

```
something else

// sourcery:inline:Type.AutoCoding
...
// sourcery:end

something else
```

[...]

PR #125, opened 2017-01-14
LinuxMain template

### LINKED ISSUE #118: Add LinuxMain generation template
Automatically finding all methods of subclasses of XCTestCase that begin with `test` and generating a `LinuxMain.swift` file will reduce the possibility of error in many cases. This would be super useful as an example for Sourcery and also (IMO) a "killer app". I was wondering if anyone has already done this – otherwise, I may have one that I could contribute (it's very rough and fragile right now).

ilyapuchka:
Resolves #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,49 @@ to this:
 - `skipEquality` allows you to skip variable from being compared.
 </details>
 
+<details>
+<summary>I want to generate `LinuxMain.swift` for all my tests</summary>
+
+For all test cases generates `allTests` static variable and passes all of them as `XCTestCaseEntry` to `XCTMain`. Run with `--args testimports='import MyTests'` parameter to import test modules.
+
+#### [Stencil template](Templates/LinuxMain.stencil)
+
+#### Available annotations:
+
+- `disableTests` allows you to disable the whole test case.
+
+#### Example output:
+
+```swift
+import XCTest
+//testimports
+
+extension AutoInjectionTests {
+  static var allTests = [
+    ("testThatItResolvesAutoInjectedDependencies", testThatItResolvesAutoInjectedDependencies),
+    ...
+  ]
+}
+
+extension AutoWiringTests {
+  static var allTests = [
+    ("testThatItCanResolveWithAutoWiring", testThatItCanResolveWithAutoWiring),
+    ...
+  ]
+}
+
+...
+
+XCTMain([
+  testCase(AutoInjectionTests.allTests),
+  testCase(AutoWiringTests.allTests),
+  ...
+])
+
+```
+
+</details>
+
 ## Writing templates
 *Sourcery templates are powered by [Stencil](https://github.com/kylef/Stencil)*
 
--- a/Templates/LinuxMain.stencil
+++ b/Templates/LinuxMain.stencil
@@ -1,0 +1,14 @@
+import XCTest
+{{ argument.testimports }}
+{% for type in types.classes|based:"XCTestCase" %}
+{% if not type.annotations.disableTests %}extension {{ type.name }} {
+  static var allTests = [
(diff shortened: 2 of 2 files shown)


krzysztofzablocki:
Let's ping @vknabel and @swizzler, since they know this usecase better, looks good to me.

vknabel:
@ilyapuchka @krzysztofzablocki Sorry for the late response. In my case the stencil template got parsed wrong and therefore an error has been thrown. Maybe the Stencil versions declared by SwiftPM and CocoaPods got out of sync?

```
$ sourcery --version
0.5.3
$ sourcery Tests Templates/LinuxMain.swift.stencil Tests/LinuxMain.swift
Scanning sources...
Found 3 types.
Loading templates...
Loaded 1 templates.
Generating code...
2017-01-16 12:11:38.066 sourcery[37327:9309308] *** Terminating app due to uncaught exception 'NSUnknownKeyException', reason: '[<sourcery.Method
0x7fa0034144f0> valueForUndefinedKey:]: this class is not key value coding-compliant for the key shortName|hasPrefix:"stest".'
```

But everything else looks promising.

ilyapuchka:
@vknabel I think that's the issue, it should point to https://github.com/kylef/Stencil/commit/99efba56e942270780a452e5dcdfa1a839839cf4

krzysztofzablocki:
that's my bad, I'll update it in my next pr

vknabel:
@ilyapuchka I have released a new tag. (Sourcery still needs to use my fork for Stencil)
Parsing the template now works as expected, but I'll need to test it on linux first,

vknabel:
@ilyapuchka Works great! Shall be add some warnings when the methods/classes are not public? This would be an compile error on Linux.

ilyapuchka:
@vknabel we ignore everything private at this moment. With #45 it will change of course. If there are private methods accessible then it's most likely a bug.

vknabel:
@ilyapuchka by default they are internal, but the `LinuxMain.swift` is not inside the test target.

krzysztofzablocki:
We should add access level filter if we don't have it, it will come handy

ilyapuchka:
@vknabel  That explains why I was using `@testable` to import test module =) So I guess there are different options. And yes, filters will be handy.

ilyapuchka:
@vknabel actually I'm not sure about warnings, feels like it's better just to make generated code not to compile if user forgot to make test method public. If he does not want to include it in test he can always adjust template to use annotations for methods or to prefix those methods names with `_` or what ever. But that's more like a general best practice for templates that we should recommend or not. @krzysztofzablocki what do you think?

krzysztofzablocki:
why don't we just filter to only support public in the template? otherwise I'd agree with @ilyapuchka, if user misuses the template its better to make it not compile so they fix their setup

ilyapuchka:
@krzysztofzablocki we can, but I guess the whole purpose of such template is to never have any tests missing because you forgot to add it in `XCTMain`. So now you don't have to add methods manually but you still can forget to make them public. If we filter methods in template nothing will remind user about missing test. So one way is not to filter and make it fail to compile, another one is to filter but when filtering leave a comment in code.

krzysztofzablocki:
I'd go with fail to compile, Sourcery driving goal is to limit human errors

vknabel:
@krzysztofzablocki @ilyapuchka Sorry for the confusion. It's good that the build breaks when one test is not public. At least it breaks when testing on Linux. Initially I thought about additionally marking non-public tests with a comment `("testSomething", testSomething) // You need to mark this declaration as public`, but that would mean that warnings would eventually be committed, which is bad.

Instead I would just mention in the `<details>`, that the tests need to be public and hope that everyone reads it. 


But this template works great. 👍

PR #127, opened 2017-01-16
Fixes unknown tag 'endif' error for AutoHashable

vknabel:
The AutoHashable template had one `{% endif %}` tag too much, which results in the following error:

```
Scanning sources...
Found 6 types.
Loading templates...
Loaded 1 templates.
Generating code...
Unknown template tag 'endif'
```


krzysztofzablocki:
thanks!

PR #133, opened 2017-01-23
Inline generation support

### LINKED ISSUE #45: Inline generated code

krzysztofzablocki:
This closes #45 
I'm going to do refactor separately so it's easier to review.


Sourcery/CodeGenerated/TypedSpec.generated.swift
```diff
@@ -0,0 +1,142 @@
+// Generated using Sourcery 0.5.3 — https://github.com/krzysztofzablocki/Sourcery
```
krzysztofzablocki: I'll clean it up, no worries

krzysztofzablocki:
@ilyapuchka could you review this again? I was going to do bigger refactor, but I have some personal stuff I need to take care of so it won't happen for this release. I made some small refactoring changes and added ability to disable cache that @vknabel needs for brew

krzysztofzablocki:
Made a mistake, need to correct one thing because its causing loop ;)

bejar37:
FWIW, looks good to me. I'm excited about having `StencilSwiftKit`, should make writing templates a lot nicer! I do think that some of the features there belong in `Stencil` proper, so hopefully these features are merged into `Stencil` one day.

Sourcery/CodeGenerated/Coding.generated.swift
```diff
@@ -39,68 +39,56 @@ extension NSCoder {
 
 }
 
+extension AssociatedValue: NSCoding {}
```
ilyapuchka: Do we still need this commented out stuff since we have it inline now?

SourceryTests/Models/TypedSpec.generated.swift
```diff
@@ -25,11 +25,6 @@ class TypedSpec: QuickSpec {
                 expect(sut.value(forKeyPath: "isTuple") as? Bool).to(equal(true))
             }
 
-            it("can report closure type via KVC") {
```
ilyapuchka: Was this removed intentionally?

krzysztofzablocki: looks like its missing from generated code there, I just copied the content

krzysztofzablocki:
@ilyapuchka I think we should wait with re-enabling private and fileprivate entries, we need to figure out how to do it so that it doesnt break regular extension based templates, I'd do it in next version (not the one adding inline) after we had some discussion about it, what do you think?

bejar37:
> I think we should wait with re-enabling private and fileprivate entries

You're only thinking of enabling this for inline generation, right @krzysztofzablocki ?

krzysztofzablocki:
> You're only thinking of enabling this for inline generation, right @krzysztofzablocki ?
I think this would require us to instead of use text annotations in templates have dedicated Stencil Nodes that would enable / disable private entries, and it might be confusing to the users to get different properties under the same name.

Thus why I'd rather wait for us to discuss it more and see what makes sense :)

PR #134, opened 2017-01-23
Camel case / Snake case filters

bolismauro:
This PR adds two new filters to convert a camel case string to a snake case one and vice versa

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### New Features
 
 - Added `isClosure` property to `TypeName` to detect closure types
+- Added `snakeCaseToCamelCase` filter
+- Added `camelCaseToSnakeCase` filter
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


krzysztofzablocki:
Hey thanks for contributing :) 
We are going to integrate https://github.com/SwiftGen/StencilSwiftKit in upcoming release so worth checking if they don't already handle it, I can take a look when I finish some refactoring tweaks

bolismauro:
Alright :)
Do you have an eta per the next release?

krzysztofzablocki:
I should do it this week, probably weekend

krzysztofzablocki:
@bolismauro I think this is covered by https://github.com/SwiftGen/StencilSwiftKit, or if not lets add it there instead of here, next release after that will contain it

bolismauro:
They miss the camel case to snake (according to the documentation).
I'll send a PR

PR #137, opened 2017-01-27
Fix associated value names

bejar37:
Newlines weren't trimmed from associated value labels, this is a small fix for that.


krzysztofzablocki:
thanks 👍

PR #140, opened 2017-01-27
Update Built-In AutoHashable Template to Compile

grantjbutler:
For the current `AutoHashable.stencil` template that is distributed with Sourcery, given some [source code](https://gist.github.com/grantjbutler/d566019d7198643d2782346e6ee66b1c#file-source-swift), the [generated code](https://gist.github.com/grantjbutler/d566019d7198643d2782346e6ee66b1c#file-orig-generated-swift) will not compile and is semantically incorrect for the following reasons:

1. Implementations of `hashValue` on types call `combineHashes` with n arguments, whereas `combineHashes` takes in an array of integers.
2. Enumeration cases with no associated values call `combineHashes` directly with no arguments, resulting in all cases of an enumeration having the same hash value.
3. There is no checking that an associated value of an enum case implements `Hashable` and can then be used for calculating the hash value of the containing case.

This PR updates the template to address most of these issues<sup>[1]</sup>. Given the same source code, the [generated code](https://gist.github.com/grantjbutler/d566019d7198643d2782346e6ee66b1c#file-new-generated-swift) using this new template compiles and behaves as expected.

I could see about adding tests to warn about this in the future, if you think that's something worth adding. I can imagine two possibilities for a test:

1. Do a straight text comparison. Generate some code using the built-in template and store it on disk for use as a fixture. Then, for a test, regenerate the code and compare it to that fixture. If the string comparison fails, then the test fails.
2. Analyzing for valid Swift code. This could either be done by attempting to compile the source code that is generated by running Sourcery with the built-in templates, or possibly by using SourceKitten to try to analyze the source code (I'm assuming that, if the Swift code has some syntax error, then SourceKitten will surface the error and not complete the analysis of the code). If either operation fails (depending on which is implemented), then the test fails.

#### Footnotes

[1]: This PR addresses problems 1 and 2. It does not appear that problem 3 can be addressed as Sourcery currently does not return the necessary information for associated values to determine what protocols these associated values do or don't conform to. I suppose a simple solution to fix part of that would be to ignore any associated value where `isClosure` is `true`, but I don't know if that's the route you want to go down.

### DIFF
--- a/Templates/AutoHashable.stencil
+++ b/Templates/AutoHashable.stencil
@@ -1,6 +1,6 @@
 // swiftlint:disable file_length
 
-fileprivate func combineHashes(hashes: [Int]) -> Int {
+fileprivate func combineHashes(_ hashes: [Int]) -> Int {
     var combinedHash = 5381
 
     for itemHash in hashes {
@@ -16,7 +16,7 @@ fileprivate func combineHashes(hashes: [Int]) -> Int {
 extension {{ type.name }}{% if not type.kind == "protocol" %}: Hashable{% endif %} {
     {% if type.supertype.based.Hashable or type.supertype.implements.AutoHashable %} THIS WONT COMPILE, WE DONT SUPPORT INHERITANCE for AutoHashable {% endif %}
     {{ type.accessLevel }} var hashValue: Int {
-        return combineHashes({% for variable in type.variables %}{% if not variable.isStatic and not variable.annotations.skipHashing %}{% if not variable.isOptional %}{{ variable.name }}.hashValue{% else %}{{ variable.name }}?.hashValue ?? 0{% endif %}, {% endif %}{% endfor %}0)
+        return combineHashes([{% for variable in type.variables %}{% if not variable.isStatic and not variable.annotations.skipHashing %}{% if not variable.isOptional %}{{ variable.name }}.hashValue{% else %}{{ variable.name }}?.hashValue ?? 0{% endif %}, {% endif %}{% endfor %}0])
     }
 }
 {% endif %}{% endfor %}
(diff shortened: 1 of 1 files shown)


Templates/AutoHashable.stencil
```diff
@@ -16,7 +16,7 @@ fileprivate func combineHashes(hashes: [Int]) -> Int {
 extension {{ type.name }}{% if not type.kind == "protocol" %}: Hashable{% endif %} {
     {% if type.supertype.based.Hashable or type.supertype.implements.AutoHashable %} THIS WONT COMPILE, WE DONT SUPPORT INHERITANCE for AutoHashable {% endif %}
     {{ type.accessLevel }} var hashValue: Int {
-        return combineHashes({% for variable in type.variables %}{% if not variable.isStatic and not variable.annotations.skipHashing %}{% if not variable.isOptional %}{{ variable.name }}.hashValue{% else %}{{ variable.name }}?.hashValue ?? 0{% endif %}, {% endif %}{% endfor %}0)
+        return combineHashes([{% for variable in type.variables %}{% if not variable.isStatic and not variable.annotations.skipHashing %}{% if not variable.isOptional %}{{ variable.name }}.hashValue{% else %}{{ variable.name }}?.hashValue ?? 0{% endif %}, {% endif %}{% endfor %}0])
```
krzysztofzablocki: you can do `for variable in type.storedVariables` to simplify the if statement

krzysztofzablocki:
@grantjbutler Thanks for contributing 👍  
you are right, my mistake as I wrote the template in our project and we have more variants of combineHashes in NYT :)

> There is no checking that an associated value of an enum case implements Hashable and can then be used for calculating the hash value of the containing case.

FYI: You can check if type of associated value implements a protocol if its your project type, but since Hashable is not known to Sourcery (Apple framework not your project) you'd have to use `type.based.Hashable`. 

But I wouldn't because I think skipping some variables for hashing should be explicit, and its too easy to not realise a variable type might not be hashable when writing type, right now you'd get compile error and just be forced to either explicitly skip it or add `AutoHashable` conformance in another type, which is simple enough and prevents some human errors.

But 👍  on skipping those that are closures.

grantjbutler:
> FYI: You can check if type of associated value implements a protocol if its your project type, but since Hashable is not known to Sourcery (Apple framework not your project) you'd have to use `type.based.Hashable`.

Does that work for `AssociatedValue`? The documentation under the "Writing Templates" section of the ReadMe didn't explicitly state that `AssociatedValue` types have all the properties from `Type`, so I assumed that doing a check for protocol conformance wasn't possible.

> But I wouldn't because I think skipping some variables for hashing should be explicit

Sure. I wanted to add support for a `skipHashing` annotation as is done for the generated implementation of `Hashable` for non-enumeration types, but it looks like you can't add annotations for associated values. I guess you add the annotation to the `EnumCase` instead. However, I'm guessing that you can't have multiple declarations of the same annotation (to opt multiple associated values out of the implementation of `hashValue`) as it's not explicitly mentioned in the ReadMe.

> But 👍 on skipping those that are closures.
>
> For tests I think using simple fixtures as you proposed in 1 would be good enough to prevent regressions as we work on the tool and templates.

Cool. I'll get those changes up as separate PRs. Would you want them as two distinct PRs, or combined into one?

krzysztofzablocki:
> Does that work for AssociatedValue? The documentation under the "Writing Templates" section of the ReadMe didn't explicitly state that AssociatedValue types have all the properties from Type, so I assumed that doing a check for protocol conformance wasn't possible.

AssociatedValue has `type` property, but it won't work for types from outside your project like Apple / 3rd Party code, so it could only be used for specific filtering. Thus why I prefer to not attempt it and just enforce that all variables that aren't skipped are hashable, avoiding programmer mistakes.

> Sure. I wanted to add support for a skipHashing annotation as is done for the generated implementation of Hashable for non-enumeration types, but it looks like you can't add annotations for associated values. I guess you add the annotation to the EnumCase instead. However, I'm guessing that you can't have multiple declarations of the same annotation (to opt multiple associated values out of the implementation of hashValue) as it's not explicitly mentioned in the ReadMe.

Indeed, right now we don't support annotations per associated value, our annotation parser though would scan them if you separated enum case definition into multiline, making adding this feature pretty straightforward.

Seems to me like it would be fine to just have a single PR for those changes we discussed

PR #144, opened 2017-02-02
Fix custom stencil tag examples on README

joanromano:
Some small fixes which at least confused me when trying to use things like `contains` or `hasPrefix` straight from README examples and caused the generation to fail.


krzysztofzablocki:
allowing spaces in filters have been removed from stencil, it used to work :) thanks for contributing!

PR #154, opened 2017-02-11
Added notes about known and unknown types

ilyapuchka:


README.md
```diff
@@ -296,6 +296,17 @@ XCTMain([
 </details>
 
 ## Writing templates
+
```
krzysztofzablocki: Can you make an expandable summary / detail section and move this after the initial reflection info, meaning just after the line
`All of these properties return Type objects.`

README.md
```diff
@@ -315,6 +316,17 @@ There are multiple ways to access your types:
 
 All of these properties return `Type` objects.
 
+<details><summary>**What are _known_ and _unknown_ types**</summary>
```
ilyapuchka: Like this?

krzysztofzablocki: Yeah, looks good

PR #157, opened 2017-02-15
Fix AutoEquatable to only use stored variables, not all variables

khanlou:
cc @krzysztofzablocki @grantjbutler

### DIFF
--- a/Templates/AutoEquatable.stencil
+++ b/Templates/AutoEquatable.stencil
@@ -25,7 +25,7 @@ fileprivate func compareArrays<T>(lhs: [T], rhs: [T], compare: (_ lhs: T, _ rhs:
 {% if not type.kind == "protocol" %}extension {{ type.name }}: Equatable {} {% endif %}
 {% if type.supertype.based.Equatable or type.supertype.implements.AutoEquatable %} THIS WONT COMPILE, WE DONT SUPPORT INHERITANCE for AutoEquatable {% endif %}
 {{ type.accessLevel }} func == (lhs: {{ type.name }}, rhs: {{ type.name }}) -> Bool {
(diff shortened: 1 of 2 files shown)


Templates/AutoEquatable.stencil
```diff
@@ -25,7 +25,7 @@ fileprivate func compareArrays<T>(lhs: [T], rhs: [T], compare: (_ lhs: T, _ rhs:
...
 {{ type.accessLevel }} func == (lhs: {{ type.name }}, rhs: {{ type.name }}) -> Bool {
-    {% for variable in type.variables %}{% if not variable.isStatic and not variable.annotations.skipEquality %}guard {% if not variable.isOptional %}{% if not variable.annotations.arrayEquality %}lhs.{{ variable.name }} == rhs.{{ variable.name }}{% else %}compareArrays(lhs: lhs.{{ variable.name }}, rhs: rhs.{{ variable.name }}, compare: ==){% endif %}{% else %}compareOptionals(lhs: lhs.{{ variable.name }}, rhs: rhs.{{ variable.name }}, compare: ==){% endif %} else { return false }{% endif %}
+    {% for variable in type.storedVariables %}{% if not variable.isStatic and not variable.annotations.skipEquality %}guard {% if not variable.isOptional %}{% if not variable.annotations.arrayEquality %}lhs.{{ variable.name }} == rhs.{{ variable.name }}{% else %}compareArrays(lhs: lhs.{{ variable.name }}, rhs: rhs.{{ variable.name }}, compare: ==){% endif %}{% else %}compareOptionals(lhs: lhs.{{ variable.name }}, rhs: rhs.{{ variable.name }}, compare: ==){% endif %} else { return false }{% endif %}
```
grantjbutler: @krzysztofzablocki can correct me if I'm wrong, but I don't believe that `storedVariables` includes static variables, so we should be good to drop the `isStatic` check from the first condition after the beginning of the `for` loop.

krzysztofzablocki: indeed, I think I wrote the original template before I had `storedVariables` :)

krzysztofzablocki:
want to get rid of isStatic check as its not needed anymore?

PR #164, opened 2017-02-22
Added example for conditional on collections

ggrell:
The example shows an if-statement on the variables collection, filtered by an annotation.


krzysztofzablocki:
thanks for contributing 👍

PR #168, opened 2017-02-24
fix: escaping reserved variable names

ilyapuchka:
See https://github.com/SwiftGen/StencilSwiftKit/blob/master/Sources/Filters.swift#L15

The reason for this change is that when we try to generate property with the same name, or use it without `self.` we loose it's escaping (SourceKitten/SourceKite returns it unescaped), and generated code does not compile

### DIFF
--- a/Sourcery/Parsing/FileParser.swift
+++ b/Sourcery/Parsing/FileParser.swift
@@ -250,9 +250,30 @@ final class FileParser {
 // MARK: - Details parsing
 extension FileParser {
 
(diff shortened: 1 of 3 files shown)


Sourcery/Parsing/FileParser.swift
```diff
@@ -250,9 +250,30 @@ final class FileParser {
 // MARK: - Details parsing
 extension FileParser {
 
+    fileprivate static let reservedKeywords = ["associatedtype", "class", "deinit", "enum", "extension",
```
krzysztofzablocki: it's ok for now but we should just make the list available in the StencilSwiftKit

Issue #143, opened 2017-02-01
Support for Xcode Build Phase input files parameters


ilyapuchka:
Currently we only support a single path parameter, but it might be the case that scanned files are located in different paths and not all the files in their common root should be scanned. That's actually the case for Sourcery itself - `GenerationContext` class is not located in the same path as all other scanned types (models). With input files parameters support we can simply add them to all scanned files in the path.

See also #126 that might also solve issue in it's own way, but adding parameters support feels a simpler solution. Generally I think more options we provide - the better, so one does not dismiss the other.

krzysztofzablocki:
What if we introduced solution similar to SwiftLint yaml definition file? we'd be listing inputs and exclude paths there, this would simplify the configuration of the tool

ilyapuchka:
Though I'm not a great fan of yaml files this could be the best option, especially when we start to grow the number of configurations

PR #170, opened 2017-02-25
JS templates

ilyapuchka:
This is a Proof of Concept for JavaScript templates based on EJS template engine. In future would be cool to find a better way to include template engine sources, so that any other JS template engine can be used without adding it to the Sourcery. Here I used browserify to create a js bundle which is then included in the app bundle.

### DIFF
--- a/Sourcery.xcodeproj/project.pbxproj
+++ b/Sourcery.xcodeproj/project.pbxproj
@@ -10,7 +10,13 @@
 		0906124D1E2BC43500BA0937 /* FileParser + MethodsSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0906124C1E2BC43500BA0937 /* FileParser + MethodsSpec.swift */; };
 		0911DAF01E1D90D1001FF3EB /* Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0911DAEF1E1D90D1001FF3EB /* Extensions.swift */; };
 		0911DAF21E212817001FF3EB /* Annotations.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0911DAF11E212817001FF3EB /* Annotations.swift */; };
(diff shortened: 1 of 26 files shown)


Sourcery.xcodeproj/project.pbxproj
```diff
@@ -176,7 +184,7 @@
 		098442721E21BF850051E1A1 /* FileParser + AttributesSpec.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = "FileParser + AttributesSpec.swift"; sourceTree = "<group>"; };
 		098442741E2241F10051E1A1 /* Attribute.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Attribute.swift; sourceTree = "<group>"; };
 		09B1DD381E219D95009228EE /* TypedSpec.stencil */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text; path = TypedSpec.stencil; sourceTree = "<group>"; };
-		09B1DD3A1E219EA3009228EE /* TypedSpec.generated.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = TypedSpec.generated.swift; sourceTree = "<group>"; };
+		09B1DD3A1E219EA3009228EE /* TypedSpec.generated.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = TypedSpec.generated.swift; path = ../../Sourcery/CodeGenerated/TypedSpec.generated.swift; sourceTree = "<group>"; };
```
krzysztofzablocki: Wasn't that file in gitignore?

ilyapuchka: Hm, I didn't notice that it's under ignore... I don't think we should ignore it, so I just removed one that was in tests folder and changed a reference in project to one in generated files folder.
I'm going to start to work on yaml file configuration that you proposed in #143 so we could have at some point a way to set specific destinations for specific templates.

PR #173, opened 2017-02-27
Fix trim attributes

### LINKED ISSUE #171: Bug report: Unwrapped type name truncating type name
A use case for Sourcery requires using `MethodParameter`'s `typeName` without any attributes (specifically, I'd like to remove the `@escaping` attribute). There's no way to do that currently without using `unwrappedTypeName`. Here's an example of what I've tried:

`{% for method in type.NetworkProtocol.methods %}
    let {{ method.shortName }}CallTracker = MockMethodCallTracker<({% for parameter in method.parameters %}{{ parameter.typeName }}{% if parameter.name != method.parameters.last.name %}, {% endif %}{% endfor %}), {{ method.returnTypeName }}>()
{% endfor %}`

This should output:
`let getLatestOpportunityCallTracker = MockMethodCallTracker<(ConversationListApiResponse), Request?>()`

Instead, we get `let getLatestOpportunityCallTracker = MockMethodCallTracker<(@escaping ConversationListApiResponse), Request?>()`

If I try changing `{{ parameter.typeName }}` to `{{ parameter.unwrappedTypeName }}`, I get an unexpected name out: it outputs `let getLatestOpportunityCallTracker = MockMethodCallTracker<(ConversationListApiRespo), Request?>()`. Looks like it's truncating from the beginning and end of the word instead of just the beginning?

ilyapuchka:
Resolves #171

### DIFF
--- a/SourceryTests/Models/MethodSpec.swift
+++ b/SourceryTests/Models/MethodSpec.swift
@@ -84,21 +84,10 @@ class MethodSpec: QuickSpec {
 
         describe("MethodParameter") {
 
-            context("given optional parameter type with short syntax") {
-                it("reports type as optional") {
-                    expect(MethodParameter(name: "foo", typeName: TypeName("Int?")).isOptional).to(beTrue())
-                }
-                it("reports non-optional type for unwrappedTypeName") {
-                    expect(MethodParameter(name: "foo", typeName: TypeName("Int?")).unwrappedTypeName).to(equal("Int"))
-                }
-            }
-
-            context("given optional parameter type with long syntax") {
-                it("reports type as optional") {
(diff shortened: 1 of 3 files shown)


SourceryTests/Models/MethodSpec.swift
```diff
@@ -84,21 +84,10 @@ class MethodSpec: QuickSpec {
 
         describe("MethodParameter") {
 
-            context("given optional parameter type with short syntax") {
```
krzysztofzablocki: why is this removed though?

ilyapuchka: it's a part of TypedSpec already

PR #177, opened 2017-03-06
Podfile fixes; #trivial

Antondomashnev:
Hi there, thanks for an amazing meta-programming tool.

I've only started checking it out and after `bundle exec pod install` the CocoaPods showed some warnings that are fixed in the PR:
- [x] smart quote is removed
- [x] `Sourcery` target's build settings had `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` flag set to true, however, according to CocoaPods suggestion, it has to be inherited from auto generated build settings from CocoaPods.

Let me know if the latter one was on purpose and I will revert the commit 😄


krzysztofzablocki:
this should be fine, thanks :)

PR #178, opened 2017-03-06
Support _ for swift3 functions in automockable template;

Antondomashnev:
I've found that automockable template generates incorrect code for the following protocol:
```swift
protocol MenuActions: AutoMockable {
    func handleIconMouseEvent(_ event: NSEvent)
}
```

Generated code:
```swift
class MenuActionsMock: MenuActions {


    //MARK: - handleIconMouseEvent

    var handleIconMouseEventCalled = false
    var handleIconMouseEventReceivedEvent: NSEvent?

    func handleIconMouseEvent( event: NSEvent) {

        handleIconMouseEventCalled = true
        handleIconMouseEventReceivedEvent = event
    }

}
```

As you can see the `func handleIconMouseEvent( event: NSEvent)` doesn't have the `_` before event, which is incorrect and leads to an compile error.

So far my changes in template fixes the problem and it generates the expected result:
```swift
class MenuActionsMock: MenuActions {


    //MARK: - handleIconMouseEvent

    var handleIconMouseEventCalled = false
    var handleIconMouseEventReceivedEvent: NSEvent?

    func handleIconMouseEvent(_ event: NSEvent) {

        handleIconMouseEventCalled = true
        handleIconMouseEventReceivedEvent = event
    }

}
```

Also, have you ever considered add tests for the templates from `/Templates`. Like check the generated code with the reference file?


krzysztofzablocki:
thanks for contributing :)

Issue #182, opened 2017-03-09
Add isDictionary check for TypeName


Antondomashnev:
Hi, I'm using Sourcery's `AutoMockable` template and for the specific code, it generated code which compiles with an error.

Let me give an example:
```
protocol MenuInteractorInput: class, AutoMockable {
    /// Cached test results per one application run
    var foundTestResults: [TestResult] { get }
}
```

For this protocol Sourcery generates:
```swift
class MenuInteractorInputMock: MenuInteractorInput {

    var foundTestResults: [TestResult] 

}
```

As you can notice the swift compiler fails with the error that `var foundTestResults: [TestResult] ` doesn't have a default value. I was going to update the template to try to fix it, but I realized that the same issue could be if the `foundTestResults` is a type of `Dictionary`, but Sourcery doesn't have such helper method on TypeName`. So I've added it and going to submit the PR. But as contributing guideline stays I want to submit an issue first which I link to PR 😄

PR #183, opened 2017-03-09
Add flag for TypeName to check whether it's dictionary or not;

Antondomashnev:
This PR implements the proposal from #182.
For Danger 😉 I haven't updated the CHANGELOG yet because I'm not sure whether is should be in next minor version or as a patch.

### DIFF
--- a/Sourcery/Parsing/Composer.swift
+++ b/Sourcery/Parsing/Composer.swift
@@ -306,24 +306,18 @@ struct Composer {
     fileprivate func parseDictionaryType(_ typeName: TypeName) -> DictionaryType? {
         let name = typeName.unwrappedTypeName
         guard name.isValidDictionaryName() else { return nil }
-        return DictionaryType(name: typeName.name, valueTypeName: parseDictionaryValueType(typeName), keyTypeName: parseDictionaryKeyType(typeName))
+        let keyValueType: (keyType: TypeName, valueType: TypeName) = parseDictionaryKeyValueType(typeName)
+        return DictionaryType(name: typeName.name, valueTypeName: keyValueType.valueType, keyTypeName: keyValueType.keyType)
     }
 
-    fileprivate func parseDictionaryKeyType(_ typeName: TypeName) -> TypeName {
+    fileprivate func parseDictionaryKeyValueType(_ typeName: TypeName) -> (keyType: TypeName, valueType: TypeName) {
         let name = typeName.unwrappedTypeName
         if name.hasPrefix("Dictionary<") {
-            return TypeName(name.drop(first: 11, last: 1).commaSeparated()[0].stripped())
+            let types = name.drop(first: 11, last: 1).commaSeparated()
+            return (TypeName(types[0].stripped()), TypeName(types[1].stripped()))
         } else {
-            return TypeName(name.dropFirstAndLast().colonSeparated()[0].stripped())
-        }
-    }
-
-    fileprivate func parseDictionaryValueType(_ typeName: TypeName) -> TypeName {
-        let name = typeName.unwrappedTypeName
-        if name.hasPrefix("Dictionary<") {
-            return TypeName(name.drop(first: 11, last: 1).commaSeparated()[1].stripped())
-        } else {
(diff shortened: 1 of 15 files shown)


SourceryTests/Models/TypeNameSpec.swift
```diff
@@ -62,6 +62,23 @@ class TypeNameSpec: QuickSpec {
...
+                    expect(TypeName("[[Int: Int]]").isArray).to(beTrue())
+                }
+            }
+
+            context("given dictionary type") {
+                context("as name") {
+                    it("reports dictionary correctly") {
+                        expect(TypeName("[Int: Int]").isDictionary).to(beTrue())
+                        expect(TypeName("[[Int]: [Int]]").isDictionary).to(beTrue())
+                        expect(TypeName("[Int: [Int: Int]]").isDictionary).to(beTrue())
```
ilyapuchka: will it work for array of dictionaries?

ilyapuchka:
Hi 😉 
Regarding changelog we usually put everything new under Master at the top and then replace it with version on release.

SourceryTests/Models/TypeNameSpec.swift
Antondomashnev: I'm not sure if I completely understand the code base at the moment, so let me know if I miss any case here.
But probably you mean that the `TypeName` of array of dictionaries should return for `isArray` true, if so I've added a test case for this and it passes on my local machine:

```swift
expect(TypeName("[[Int: Int]]").isArray).to(beTrue())
```

ilyapuchka: Yes, isArray should be true, and isDictionary should be false. Just add this case here and maybe some more negative cases so that we don't have regressions here.

Antondomashnev: That makes sense, I've added couple of text cases in e1e321b.

Antondomashnev:
> I could not see where do you actually parse dictionary content. We have array parsing in Composer, I guess you can add it there.

@ilyapuchka you're right. Thanks for pointing it to me 👍. I will implement it.

Antondomashnev:
@krzysztofzablocki @ilyapuchka ready to be checked 🎉

Sourcery/Parsing/Composer.swift
```diff
@@ -299,6 +303,30 @@ struct Composer {
...
+    fileprivate func parseDictionaryKeyType(_ typeName: TypeName) -> TypeName {
+        let name = typeName.unwrappedTypeName
+        if name.hasPrefix("Dictionary<") {
+            return TypeName(name.drop(first: 11, last: 1).commaSeparated()[0].stripped())
+        } else {
+            return TypeName(name.dropFirstAndLast().colonSeparated()[0].stripped())
+        }
+    }
+
+    fileprivate func parseDictionaryValueType(_ typeName: TypeName) -> TypeName {
```
krzysztofzablocki: what if you just did both at the same time? would be faster

SourceryTests/Models/TypedSpec.generated.swift
```diff
@@ -14,26 +14,31 @@ class TypedSpec: QuickSpec {
                 expect(AssociatedValue(typeName: TypeName("Int!")).value(forKeyPath: "isImplicitlyUnwrappedOptional") as? Bool).to(equal(true))
                 expect(AssociatedValue(typeName: TypeName("Int?")).value(forKeyPath: "unwrappedTypeName") as? String).to(equal("Int"))
             }
-
+
```
ilyapuchka: looks like you don't have swiftlint =)

Sourcery/Parsing/Composer.swift
Antondomashnev: Sounds reasonable, I've merged them into one in b08a95b.

SourceryTests/Models/TypedSpec.generated.swift
Antondomashnev: Ah, yes, sorry for that. It's actually fixed also in b08a95b.

ilyapuchka:
👍 thanks for contribution!

Issue #181, opened 2017-03-09
Error parsing reserved keywords as Enum cases


txaiwieser:
Steps to reproduce:
1) Try parsing any file with the following code
Any of the following Enum cases give a parsing error. (theres probably way more words, all reserved keywords maybe?)

```
enum Test {
    case none
    case get(id: String)
    case required
}
```

Thanks! 👍

krzysztofzablocki:
@ilyapuchka possible regression due to the changes we merged recently for the keywords?

ilyapuchka:
`¯\_(ツ)_/¯` maybe, I can take a quick look

ilyapuchka:
Yes, that's it. I will revert and implement a better way to handle this. There are some other minor issues in this case.

ilyapuchka:
I believe this should be fixed, please let us know if you find any other issues related to this @txaiwieser

PR #184, opened 2017-03-09
fixed parsing escaped names

ilyapuchka:
@krzysztofzablocki this fixes the issue #181 but it would be better to also fix another one.
In case of escaped names offsets returned by SourceKitten are not correct, so for example "\`none\`" will give "\`non" for name, but at the same time "name" key in the dictionary contains correct name. I guess other substrings will be extracted not correctly too.
I will write some tests for substrings extraction and will try to fix that later.


krzysztofzablocki:
sounds good, thanks 👍

PR #187, opened 2017-03-12
Scanning multiple paths & YAML

### LINKED ISSUE #143: Support for Xcode Build Phase input files parameters

ilyapuchka:
Resolves #143 

Had to change how CLI arguments are passed to support variadic lists (have to use options instead of arguments as they should be separated by option name).
We can now add more options to YAML configuration, in the end I guess CLI arguments can be deprecated as YAML-based configs are much more flexible.


ilyapuchka:
sure we should, but I guess we can do that step by step as improvements

PR #189, opened 2017-03-12
Respect file extension in file annotation in templates

### LINKED ISSUE #175: Output other than swift files
Its really good to see the ability to redirect to other files. I think it would be good to support any file type. Two cases immediately come to mind; documentation (HTML) and Dot files (graphviz) for visualizations.

Any ideas or objections?

ilyapuchka:
Resolves #175

With this you can wrap all template content in `sourcery:file` annotation to generate file with arbitrary name and extension.

There is yet one thing to be fixed - we always put a comment at the top of generated file, but it can be not valid for non-swift types of files. @krzysztofzablocki any ideas how to handle that?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### New Features
 
 - Added flag to check if `TypeName` is dictionary
+- Added generation non-swift files using `sourcery:file` annotation
 
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


krzysztofzablocki:
if the file isn't a .swift file we could drop the generation header, thats probably best we can do.

PR #190, opened 2017-03-14
Style improvement auto-equatable template; #Trivial

Antondomashnev:
Hi, the PR consists of two small changes in `AutoEquatable` template:
1. It generates code if the type is `based` and not `implementing` AutoEquatable.

This fixes the problem that for the following code Sourcery didn't generate code:
```swift
protocol TestResult: AutoEquatable {
    var referenceImagePath: String { get }
    var diffImagePath: String { get }
    var failedImagePath: String { get }
    var testName: String { get }
}

struct CompletedTestResult: TestResult {
    let referenceImagePath: String
    let diffImagePath: String
    let failedImagePath: String
    let testName: String
}
```

2. Small indents and spaces updates in the template which converts the following autogenerated code:
```swift
extension SnapshotTestImage: Equatable {}
internal func == (lhs: SnapshotTestImage, rhs: SnapshotTestImage) -> Bool {
    switch (lhs, rhs) {
         case (.diff(let lhs), .diff(let rhs)): 
                    return lhs == rhs

         case (.reference(let lhs), .reference(let rhs)): 
                    return lhs == rhs

         case (.failed(let lhs), .failed(let rhs)): 
                    return lhs == rhs

        default: return false
    }
}
```
Into
```swift
extension SnapshotTestImage: Equatable {}
internal func == (lhs: SnapshotTestImage, rhs: SnapshotTestImage) -> Bool {
    switch (lhs, rhs) {
    case (.diff(let lhs), .diff(let rhs)): 
        return lhs == rhs
    case (.reference(let lhs), .reference(let rhs)): 
        return lhs == rhs
    case (.failed(let lhs), .failed(let rhs)): 
        return lhs == rhs
    default: return false
    }
}
```

I think it looks clearer, but I can rollback for sure 😄

### DIFF
--- a/Templates/AutoEquatable.stencil
+++ b/Templates/AutoEquatable.stencil
@@ -38,19 +38,18 @@ fileprivate func compareArrays<T>(lhs: [T], rhs: [T], compare: (_ lhs: T, _ rhs:
 extension {{ type.name }}: Equatable {}
 {{ type.accessLevel }} func == (lhs: {{ type.name }}, rhs: {{ type.name }}) -> Bool {
     switch (lhs, rhs) {
-        {% for case in type.cases %}
-        {% if case.hasAssociatedValue %} case (.{{ case.name }}(let lhs), .{{ case.name }}(let rhs)): {% else %} case (.{{ case.name }}, .{{ case.name }}): {% endif %}
-            {% ifnot case.hasAssociatedValue %} return true {% else %}
-                {% if case.associatedValues.count == 1 %}
-                    return lhs == rhs
-                {% else %}
-                    {% for associated in case.associatedValues %} if lhs.{{ associated.externalName }} != rhs.{{ associated.externalName }} { return false }
-                    {% endfor %} return true
-                    {% endif %}
-            {% endif %}
-
-        {% endfor %}
-        default: return false
+    {% for case in type.cases %}
+    {% if case.hasAssociatedValue %}case (.{{ case.name }}(let lhs), .{{ case.name }}(let rhs)): {% else %} case (.{{ case.name }}, .{{ case.name }}): {% endif %}
+        {% ifnot case.hasAssociatedValue %} return true {% else %}
+        {% if case.associatedValues.count == 1 %}
+        return lhs == rhs
+        {% else %}
+        {% for associated in case.associatedValues %}if lhs.{{ associated.externalName }} != rhs.{{ associated.externalName }} { return false }
+        {% endfor %}return true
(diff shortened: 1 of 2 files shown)


krzysztofzablocki:
`implementing` should also flatten protocols, as long as the types are known to sourcery, is thats not happening for you ? there might be a regression here

Antondomashnev:
@krzysztofzablocki I've done small test now and it seems that it doesn't flatten protocols although I've checked the `GeneratorSpec` and this case seemed to be covered there.

When I use the following template to test: 
```
{% for type in types.structs %}{{ type.name }}:{{ type.implements }} {% endfor %}
```
It outputs for my example the following with omitted `variables` description:
```
CompletedTestResult:{
TestResult = "Protocol: typealiases = [:], isExtension = false, kind = protocol, 
accessLevel = internal, name = TestResult, isGeneric = false, localName = TestResult, 
variables = [...], inheritedTypes = [\"AutoEquatable\"], containedTypes = [], 
parentName = nil, 
parentTypes = AnyIterator<Type>(_box:Swift._IteratorBox<Swift._ClosureBasedIterator<Sourcery.Type>>), 
attributes = [:]kind = protocol";
}
```
As you can notice that the protocol has `AutoEquatable` as inherited type, but it was not flattened for `CompletedTestResult`. Could you confirm if it's a bug or expectec behaviour?

ilyapuchka:
@Antondomashnev most likely you are not scanning sources where you define `AutoEquatable` protocol, so Sourcery does not know if it's protocol or not, that's why it does not appear in `implements` (which is only for protocols), but do appear in `based`. Just run the test with your code + AutoEquatable definition and it's fine.

Antondomashnev:
@ilyapuchka gotcha, you're right. Thanks 😄 
I will revert these changes and leave only style one.

krzysztofzablocki:
@Antondomashnev could you do this style tweak for hashable as well to keep them consistent?

Antondomashnev:
@krzysztofzablocki done in 6a7ebbe 😄 

Looks like this:
```swift
// MARK: - SnapshotTestImage AutoHashable
extension SnapshotTestImage: Hashable {
    internal var hashValue: Int {
        switch self {
        case .diff(let data): 
            return combineHashes([1, data.hashValue])
        case .reference(let data): 
            return combineHashes([2, data.hashValue])
        case .failed(let data): 
            return combineHashes([3, data.hashValue])
        }
    }
}
```

krzysztofzablocki:
thanks 👍

PR #194, opened 2017-03-19
Automockable template update; #trivial

Antondomashnev:
The PR brings two small addition:
1. `AutoMockable` template generates code with a support of non-optional variables in the protocol if the variable of type Dictionary or Array.
2. Add information about recently added `DictionaryType` in README.


krzysztofzablocki:
awesome, thanks for contributing!

Issue #195, opened 2017-03-22
Automatic placement of markers for inline code generation?


Obbut:
I'm working on a [object persistence framework](https://github.com/openkitten/meow) for Swift, for which I use Sourcery for code generation. That works really well.

The models that I let users create look like this:

```swift
class MyModel: Model {
    var sample: String
    init() { self.sample = "example" }
    // sourcery:inline:MyModel.Meow
    // sourcery:end
}
``` 

While this is of course great, it would be even greater if I could remove the need for my users to declare the `sourcery:inline` markers by adding them automatically. In the template, there would be a way to define that the inline code generation markers are added (for example) at the end of the class declaration, if they are not existent already. It could look like this: `// sourcery:inline:MyModel`, where Sourcery would look up the type "MyModel" and place the code (and markers) at the bottom of the declaration.

My questions boil down to this:

- Would you be open to a feature like this? I'm willing to implement it myself, but If you don't like it I'd rather save the time
- Would this technically even be (reasonably) possible?
- If yes and yes, can you point me in the right direction for implementing this so I can make a PR?

Joannis:
This would reduce boilerplate code even more than sourcery already does.

krzysztofzablocki:
First off, love your framework idea and its great to see Sourcery leveraged so much 👍 

It's possible for sure, when you parse entries you can get information on when they end so you could track the last `}` of the original type definition.

In your template you could use a new markup to imply the automatic insertion, something like:
`// sourcery:inline:auto:MyModel`

You can take a look at [annotation parser](https://github.com/krzysztofzablocki/Sourcery/blob/master/Sourcery/Parsing/Utils/AnnotationsParser.swift), the replacement logic happens in [here](https://github.com/krzysztofzablocki/Sourcery/blob/master/Sourcery/Sourcery.swift#L301)

Its not that complicated
ps. if you are touching that code it might be worth looking whether you have idea how to fix #156

I'm traveling right now so my availability is limited, but feel free to shoot either me or @ilyapuchka any questions you have 👍

Obbut:
Thanks, I will look into it tomorrow 🙂

Obbut:
I implemented it in #196. I looked into #156 but I'm not able to reproduce it. I haven't had any crashes with inline code generation myself.

Thanks for the pointers to the code, it really helped!

PR #196, opened 2017-03-24
Implemented automatic inline code generation

Obbut:
This PR implements #195 

To implement this, I made the following changes:

- Added a new class `TypeDefinition` that is used as a variable on `Type` to store the location (file, offset) where a type is defined
- Use this `definition` variable in `processInlineRanges` to add the markers automatically
- Added a test
- Re-ran sourcery on itself

The result is that you can include `// sourcery:inline:auto:MyType`. If no inline range is found by `processInlineRanges`, the comments are automatically added at the end of the declaration.

### DIFF
--- a/Sourcery/Parsing/FileParser.swift
+++ b/Sourcery/Parsing/FileParser.swift
@@ -217,14 +217,13 @@ final class FileParser {
     }
 
     private func makeTypeDefinition(source: [String: SourceKitRepresentable]) -> TypeDefinition? {
-        guard let bodyOffset = source[SwiftDocKey.bodyOffset.rawValue] as? Int64,
-            let bodyLength = source[SwiftDocKey.bodyLength.rawValue] as? Int64,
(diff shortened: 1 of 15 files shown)


Sourcery/Parsing/FileParser.swift
```diff
@@ -215,6 +216,17 @@ final class FileParser {
         }
     }
 
+    private func makeTypeDefinition(source: [String: SourceKitRepresentable]) -> TypeDefinition? {
+        guard let bodyOffset = source[SwiftDocKey.bodyOffset.rawValue] as? Int64,
```
krzysztofzablocki: you can leverage `Substring.body.range(for: source)`

Obbut:
I have updated the readme + changelog and updated the code to use `Substring.body.range(for: source)`

krzysztofzablocki:
thanks! It will be live in next release 👍

PR #199, opened 2017-03-25
Jazzy docs

ilyapuchka:
@krzysztofzablocki I took some time to generate docs with jazzy. Feels like it's better than having everything listed in a single README, so we can clean it up and have a better docs structure.
It still need some tweaks, but all the types that needs to be documented seems to be ready.


ilyapuchka:
@krzysztofzablocki I think I'm done here, I moved some of the stuff from READMY to separate "guides" which will appear in docs navigation menu, left in READMY only the most important stuff. 

I also want to make a danger plugin that will check that documentation is updated when source changes (by running jazzy and checking if there are any changes appear) . Do you think it will make sense?

krzysztofzablocki:
This looks good, can we keep the `Why?` section in the main readme though ? it makes important statement to convience people to take a deeper look

ilyapuchka:
Sure. I will also update it to the latest changes.

ilyapuchka:
Good to merge @krzysztofzablocki ?

krzysztofzablocki:
go ahead

PR #212, opened 2017-04-07
Inline annotations

### LINKED ISSUE #142: Support annotations for associated values
We should add support for annotating associated values, mentioned by @grantjbutler and also hit the need for it today.

ilyapuchka:
Resolves #142 

I feel like annotations parser could become simpler if we use `SyntaxMap`. It provides comments and their positions, we need match them with definitions positions and extract their content. Will look at it some time later.


krzysztofzablocki:
the `AnnotationParser` is pretty simple + its only run once on the file and then its results are re-used, but this inline parsing is a bit more complicated

PR #215, opened 2017-04-10
add explicit type to allTests in LinuxMain

zayass:
Add explicit type for empty test suites


krzysztofzablocki:
thanks for contributing!

PR #228, opened 2017-04-16
Fixed typo in README

jschmid:


ilyapuchka:
Good catch =) Can you also fix that in guides doc and run jazzy, so that this fix make it to the docs too?

jschmid:
@ilyapuchka I updated the docs, but running `jazzy` updates a few other files too, so I don't know if I should commit them.

ilyapuchka:
For now commit everything that is changed after running `jazzy --clean`. That's what I do. We can figure out if we should ignore it or not later.

jschmid:
Alright done 👍

Issue #176, opened 2017-03-02
Error executing via build-phases ${parent_path}


txaiwieser:
I had an old version of Sourcery in a project, and today after I updated to 0.6.8 I'm having this error
Coudn't figure out why. I've already updated my templates using CLI. Bug when using via build-phases I get this error:
```
: line 6: 13376 Illegal instruction: 4  "${parent_path}"/Sourcery.app/Contents/MacOS/Sourcery "$@"
Command /bin/sh failed with exit code 132
```
This is my shell-script
```
"$SRCROOT/Frameworks/Sourcery/bin/sourcery" "${SRCROOT}/Warren" "${SRCROOT}/Frameworks/Sourcery/Templates" "${SRCROOT}/Frameworks/Sourcery/Sourcery.generated.swift"
```
Any info?
Thank you!

txaiwieser:
I found a similar issue, but still couldnt make it work :/

#35

txaiwieser:
Also, #130

krzysztofzablocki:
try clearing the cache, there was a bug in old version of Sourcery so when switching you might need to clean it by hand, remove folder `~/Library/Caches/Sourcery/`

txaiwieser:
Cleaning the cache didn't work either, I've tried everything I could think of :/
same error: 
```
line 6:  1137 Illegal instruction: 4  "${parent_path}"/Sourcery.app/Contents/MacOS/Sourcery "$@"
```

krzysztofzablocki:
the paths setup seems a little weird to me, how did you install the tool?
Can you try pointing the output to a folder rather than a single file?

txaiwieser:
I downloaded the last .zip from release. I have a folder structure inside my proj
../Frameworks/
            ..bin/
            ..Templates/
and on the xcode build-phases I created the script, it was working until I updated last week

I did try changing the output to a folder instead of a .swift file. still results in the same error when calling the app from the script.

krzysztofzablocki:
@txaiwieser can you try building from source and running on your codebase ? its hard to tell what is happening without that

txaiwieser:
I really appreciate your help! Thanks!
I build and then ran from source and got more infos

```
Scanning sources...
GitHub/Warren-iOS/Warren/Core/MainDefines.swift: Unsupported entry "internal varGlobal WEBSITE_LINK"
[...]

PR #232, opened 2017-04-17
Fix Typealias parsing + processing tweaks

### LINKED ISSUE #222: Cannot process large files
Hi!
Today I received error from Sourcery script: "Illegal instruction: 4"
After some research I find, that error occurs when sourcery trying parse Localized.swift (in this file I keep all NSLocalizedString, for now it have 1200 lines)
Error disappears when I remove some func.
e.g. 
I remove one function and error disappears
I remove other function and error remains, but if I remove more other functions (not from the example above), then error disappears

Maybe sourcery have limit on file parse?

As a workaround I select only necessary for scan folders
Would be nice to have ability to exclude some files

swift version:
```
Apple Swift version 3.1 (swiftlang-802.0.48 clang-802.0.38)
Target: x86_64-apple-macosx10.9
```

krzysztofzablocki:
This should fix #222 and #176


ilyapuchka:
Looks great!

PR #237, opened 2017-04-17
Decorator template

ilyapuchka:


Templates/Decorator.swifttemplate
```diff
@@ -0,0 +1,81 @@
+<%# Constrcuts full method declaration string -%>
```
krzysztofzablocki: spelling of Constructs all around this file

Issue #90, opened 2016-12-31
Generate more errors to ease writing templates


krzysztofzablocki:
@kzaher suggested a valid point, 
if someone is doing things like `types.implementing.AutoEquatable` it would be useful to throw error if `AutoEquatable` isn't in our database of known types, so that people understand why it's happening.

If we are doing that then we can as well do following:
- If someone references Class in `implementing` inform them it's used via `inheriting` or `base`, same for mismatches in `inheriting`
- If someone references a type we don't know about via `implementing` `inheriting` let them know about `base`

Joannis:
I didn't see this yet. An ancient issue. I could try contributing to this, if you'd like.

krzysztofzablocki:
@Joannis contributions are always welcomed 👍

PR #239, opened 2017-04-18
Improved logging

ilyapuchka:


ilyapuchka:
@krzysztofzablocki I've added some basic implementation for errors mentioned in #90, let me know what you think about such approach

Sourcery/Models/TemplateContext.swift
```diff
@@ -118,46 +118,87 @@ public final class Types: NSObject, SourceryModel {
...
             }
         }
-        return content
-    }()
+        self.types = content
+    }
+
+}
 
+public class ImplementingTypesCollection: TypesCollection {
```
krzysztofzablocki: rather than duplicating the code, how about just a single class that gets the collection closure injected + add the keyword so that logs are more clear

Sourcery/Parsing/FileParser.swift
```diff
@@ -155,7 +156,7 @@ final class FileParser {
             case .varParameter:
                 return parseParameter(source)
             default:
-                if verbose { print("\(logPrefix)Unsupported entry \"\(access) \(kind) \(name)\"") }
+                Log.warning("\(logPrefix) Unsupported entry \"\(access) \(kind) \(name)\"")
```
krzysztofzablocki: I'd keep the distinction between verbose and warnings here, this is common as there are language concepts we might not support vs we should always be able to parse enum case

ilyapuchka: You mean using `Log.info` here?

Sourcery/Models/TemplateContext.swift
ilyapuchka: Yeah, I think we can extract implementation of `value(forUndefinedKey:)` in a closure parameter 👍

Sourcery/Parsing/FileParser.swift
krzysztofzablocki: `Log.verbose` since this is expected to happen quite often

Sourcery/Models/TemplateContext.swift
```diff
@@ -118,32 +118,60 @@ public final class Types: NSObject, SourceryModel {
...
     // sourcery: skipDescription, skipEquality, skipCoding
     /// Classes inheriting from any known class, grouped by its name.
     /// `types.inheriting.MyClass` returns list of types inheriting from `MyClass`
-    public lazy internal(set) var inheriting: InheritingTypesCollection = {
-        InheritingTypesCollection(types: self.types)
+    public lazy internal(set) var inheriting: TypesCollection = {
+        TypesCollection(
+            types: self.types,
+            collection: { Array($0.inherits.keys) },
+            valueForKey: { (collection: TypesCollection, key: String) -> Any? in
```
krzysztofzablocki: this valueForKey still is the same across all variants, so why not just keep it in the base class rather than parametrize?

ilyapuchka:
@krzysztofzablocki I think I'm done here. Would be nice to implement the same runtime errors for `inherits` and `implements` properties of `Type`, but I can not figure out how to make it effectively for js templates (basically we need to wrap each type into proxy, but I'm not js expert so don't know how good/bad is that), so I'll leave it for later.

Sourcery/main.swift
```diff
@@ -114,8 +117,9 @@ func runCLI() {
         Option<Path>("output", ".", description: "Path to output. File or Directory. Default is current path."),
         Option<Path>("config", ".", description: "Path to config file. Directory. Default is current path."),
         Argument<CustomArguments>("args", description: "Custom values to pass to templates.")
-    ) { watcherEnabled, disableCache, verboseLogging, prune, sources, templates, output, configPath, args in
+    ) { watcherEnabled, disableCache, verboseLogging, quite, prune, sources, templates, output, configPath, args in
         do {
+            Log.level = verboseLogging ? .verbose : quite ? .errors : .warnings
```
krzysztofzablocki: I wonder if this works correctly without the `( )`, I'd recommend using something like switch for readability imho

PR #242, opened 2017-04-19
Improved error reporting and includes for EJS templates

Obbut:
This PR improves the usability of EJS templates with Sourcery.

## Error reporting

Errors are now correctly reported and show up in generated code like this:

```
// Error parsing template.js:3
// Can't find variable: something
```

or this:

```
// Error compiling/rendering EJS template
// Could not find matching close tag for "<%-".
```

This is extremely more helpful than the current situation: if an error occurs, the template just results in "undefined".

## Includes

EJS templates can include other EJS templates using `<%- include('myTemplate.js') %>`. I implemented the EJS IncludeCallback to facilitate this. Templates are loaded relative to the parent template as stated in the EJS documentation.

Supplying local variables / custom context to templates is also supported: `include('myTemplate.js', {type: myType})`


Sourcery/Generating/Template/JavaScript/JavaScriptTemplate.swift
```diff
@@ -20,11 +20,34 @@ final class JavaScriptTemplate: Template {
...
+
+        let include: @convention(block) (String) -> String? = { [unowned self] path in
+            let path = self.sourcePath.parent() + Path(path)
+            let includedTemplate: String? = try? path.read()
+            return includedTemplate
+        }
+
+        let rethrow: @convention(block) (NSDictionary, String, String, String) -> Void = { error, source, filename, lineno in
+            errorContent += "// Error parsing \(filename):\(lineno)\n"
+            errorContent += "// \(error["message"] as? String ?? "")"
```
ilyapuchka: let's merge this message into one line like you do in js code

Sourcery/Generating/Template/JavaScript/ejsbundle.js
```diff
@@ -1,7 +1,25 @@
...
+      return renderTemplate(template, d, path);
+     }, function(error,source,filename,lineno) {
+      rethrow({message: error.message}, source, filename, lineno);
+    });
+  }
+  
+  try {
+    global.content = renderTemplate(template, templateContext, templateName);
+  } catch (e) {
+    global.content = "// Error compiling/rendering EJS template\n// " + e.message;
```
ilyapuchka: maybe better to rethrow instead of setting content? sourcer, fileame and lineno can be optional.

Sourcery/Generating/Template/JavaScript/JavaScriptTemplate.swift
```diff
@@ -20,11 +20,34 @@ final class JavaScriptTemplate: Template {
         let path = Bundle(for: JavaScriptTemplate.self).path(forResource: "ejsbundle", ofType: "js")!
         let ejs = try String(contentsOfFile: path, encoding: .utf8)
 
+        var errorContent = ""
```
ilyapuchka: Maybe it's better to use `Error` instead of string and just throw it if it's not nil, or return content?

ilyapuchka:
Thanks for contribution @Obbut ! Don't forget also to update guide on JavaScript templates (guides/Writing templates.md) and regenerate docs (with `jazzy --clean`) 📖

Obbut:
Thanks, I'll update the PR with your feedback 🙂

Sourcery/Generating/Template/JavaScript/JavaScriptTemplate.swift
```diff
@@ -20,11 +20,34 @@ final class JavaScriptTemplate: Template {
...
+
+        let include: @convention(block) (String) -> String? = { [unowned self] path in
+            let path = self.sourcePath.parent() + Path(path)
+            let includedTemplate: String? = try? path.read()
+            return includedTemplate
+        }
+
+        let rethrow: @convention(block) (NSDictionary, String, String, String) -> Void = { error, source, filename, lineno in
```
ilyapuchka: You can set exception handler on the js context and throw error directly in js instead of calling a `rethrow` block. Also you will not need a try/catch. I think it will simplify things.

Sourcery/Generating/Template/JavaScript/JavaScriptTemplate.swift
Obbut: That would work, but the error is only printed in the console and only when Sourcery is not running in watch mode. I used this method because the error is visible in the generated code.

How would the error be visible to the user?

Obbut:
I made a few changes:

- added a test
- updated the docs
- moved to an exception handler on the JSContext
- included templates now inherit the parent context, which I think makes sense in the context where Sourcery is used

Errors do currently not get reported in watch mode, but I'll open a separate issue and/or PR for that as it's not related to just EJS.

ilyapuchka:
@Obbut I'm already working on logging improvements in #239 btw

SourceryTests/Stub/JavaScriptTemplates/Includes.js
```diff
@@ -0,0 +1 @@
+<%- include('Equality.js') %>-%><%- include('Other.js') %>-%>
```
ilyapuchka: is it a valid syntax to have just a closing `-%>`? I though it needs open tag too, like `<% -%>`

Obbut: Yes, that's a newline slurp tag, which is needed to match the expected result (otherwise it appends a newline)

ilyapuchka: Yep, I know this tag but was not aware that it can be used alone. Actually it's now supported in swift too, but I guess just using closing tag will not work there...

Obbut: When comparing the result of a template with just `<%- include('Equality.js') %>` to the result of the `Equality.js` itself I noticed that EJS adds an additional newline. To match the result files that were already existing without changing them, I added this. So without it the tests would fail 🙂

Obbut:
Btw, now we're on the subject of EJS templates, currently templates with the [`js` extension](https://github.com/krzysztofzablocki/Sourcery/blob/ddc70f3eed66d9bacf71874f5ca893f1b2f30cbb/Sourcery/Sourcery.swift#L146) are parsed as EJS templates. However, most (not Sourcery related) templates I see [online](https://github.com/mde/ejs/tree/master/examples) have the `ejs` extension. Maybe it's an idea to support that as well?

SourceryTests/Stub/JavaScriptTemplates/Includes.js
ilyapuchka: Yes, that's ok, though it still feels like undocumented feature, in all examples on github it was using `<% -%>`. But if it works - fine.
Another thing is that ejs as far as I can tell does not require file extension in include. I think we could do the same.

ilyapuchka:
@Obbut good point, we should use `ejs`extension. We can support both of them, but personally I'm ok with doing a breaking change and requiring `ejs` extension from now on. @krzysztofzablocki ?

SourceryTests/Stub/JavaScriptTemplates/Includes.js
Obbut: Ok, I didn't really know what you meant – I never used the `-%>` with a starting tag. If you want I can change it to `<% -%>`. I'll look into the file extension requirement after we discussed if the file extension should be changed to `ejs` or not

krzysztofzablocki:
breaking change is fine if its more aligned with the industry standards, at least before we finally mark it up 1.0, `.ejs` 👍

ilyapuchka:
@Obbut do you plan to make the changes to file extension?

Obbut:
@ilyapuchka I've been a bit busy with other things over the last few days, sorry. I'm on it now 🙂

Obbut:
I changed the file extension to "ejs", and includes without an extension now also work: `include('Equality')` includes `Equality.ejs`

Is there anything else I should change/update before the PR can be merged?

Sourcery/Generating/Template/JavaScript/JavaScriptTemplate.swift
```diff
@@ -38,7 +38,13 @@ final class JavaScriptTemplate: Template {
 
         let include: @convention(block) (String) -> String? = { [unowned self] path in
             let path = self.sourcePath.parent() + Path(path)
-            let includedTemplate: String? = try? path.read()
+            var includedTemplate: String? = try? path.read()
```
ilyapuchka: Minor thing - I think we can just check right away if the path has extension and append it on demand, then try to read once.

Obbut: Well, I chose to do it like this so you can do things like `include('My.template.with.custom.dot.based.naming.scheme')` (which needs the `ejs` extension) and `include('Mytemplate.js')` (which does not have the `ejs` extension) which could be tricky otherwise.

If you really think it's better to check if the extension is `ejs` before importing (and limit the import function to `ejs` files), I'll change it of course, just let me know 🙂

ilyapuchka:
Looks good, lets solve the merge conflicts (I've already added error reporting in #90) and merge this 🎉

Obbut:
One thing I just noticed: when running Sourcery in watch mode, it doesn't update when a change occurs in an included template. However, I don't know how the watcher works and if this would be easy to solve?

Obbut:
I also just noticed that the behaviour of including a template from an included template in a different folder isn't according to the spec currently - they should be relative from the template that includes the `include`, but currently they are relative from the parent template (with the first `include`).

If we can wait a bit longer before merging the PR, I'll fix it 🙂

Obbut:
Alright, I've updated the PR again. Do I need to resolve the conflicts?

ilyapuchka:
I can resolve them if you ok with that. Thanks for your work 👍

krzysztofzablocki:
@Obbut watcher will look at file changes in the folders that you specify as templates or source and react to their changes, right now the requirement are that the files that changes are either template file or swift file, its here:

https://github.com/krzysztofzablocki/Sourcery/blob/master/Sourcery/Sourcery.swift#L125

and the file types are declared here:
https://github.com/krzysztofzablocki/Sourcery/blob/a093692142ccc39904faef32fcdedee0db1d9e90/Sourcery/Utils/Path%2BExtensions.swift#L33

PR #247, opened 2017-04-20
Better support for generics

Joannis:
This WIP PR aims to improve support for generics. Any commentary on implementation and the exposed API is welcome.


Sourcery/Models/TypeName.swift
```diff
@@ -82,11 +82,21 @@ public final class TypeName: NSObject, AutoCoding, AutoEquatable, AutoDiffable,
...
+            
+        }
     }
 
     /// Type name used in declaration
     public let name: String
-
+    
+    /// Type name without the generics
+    public var baseTypeName: String
```
ilyapuchka: I would go for moving that to `GenericType` the same way as we do for dictionary, array and tuples types. Then we can have properties there for base type name, generic parameter and so on (even if we have just one now it could be more extensible). And in `TypeName` we will have `isGeneric: Bool` and `generic: GenericType?` properties. What do you think?

Joannis: Thanks for the input, good idea :) I'll change this

Joannis: How would this work with respect to classes, enums and structs? They can be generic but aren't necessarily. I'm assuming you're on the same line where all three have a `generic: GenericType?` variable?

ilyapuchka: all the types have `typeName: TypeName` property, so generic info will be accessed like this `type.typeName.generic.baseTypeName` or something like that for any type, even for `Protocol`

Joannis:
```swift
public protocol StorageMechanism {}

public struct S3 : StorageMechanism {}

public struct File<S: StorageMechanism> {
    public var name: String
    
    public init(named name: String) {
        self.name = name
    }
}

public struct User {
    public let profilePicture: File<S3>
    public let name: String
    public let email: String
    
    public init() {
        self.name = "Joannis"
        self.email = "joannis@orlandos.nl"
        self.profilePicture = File<S3>(named: "test.png")
    }
}
```

I'm developing using this example now. When I'm running this template it works as expected:

```
<% types.all.forEach(mytype => { %>
<%-mytype.name%>
  <% mytype.allVariables.forEach(variable => {%>
    <%-variable.name%> <%-variable.type%> <%- variable.typeName.isGeneric %>
      <% if(variable.typeName.isGeneric) {
        variable.typeName.generic.referencedTypes.forEach(referenced => {%>
          <%- variable.typeName.generic.name %> <%- referenced.name %> <%- referenced.kind %>
<%        }); 
          }
       });
   });
%>
```

Whenever I try to print `<%- variable.type.name %>` I'm getting thrown an error (which is what @obbut is working on now, so no idea why). `Undefined`. That's all I need to do I think.

Sourcery/Parsing/Composer.swift
```diff
@@ -98,6 +98,30 @@ struct Composer {
...
+            guard genericCharacters.count > 1, genericCharacters.removeLast() == ">" else {
+                return nil
+            }
+            
+            let genericTypes = genericCharacters.split(separator: ",").map { characters in
+                return String(characters).trimmingCharacters(in: [" "])
+            }
+            
+            lookupName.generic = GenericType(name: String(typeCharacters), referencedTypes: genericTypes.flatMap { genericTypeName in
+                return unique[genericTypeName]
```
ilyapuchka: type parameter can be typealias, or another generic, so we should use `resolveType` here, we should test all these cases.

Sourcery/Parsing/Composer.swift
```diff
@@ -113,8 +137,7 @@ struct Composer {
             }
         }
         
-        return unique[lookupName.unwrappedTypeName] ?? unique[lookupName.baseTypeName]
-            ?? typeFromModule(lookupName.unwrappedTypeName, modules: modules)
+        return unique[lookupName.unwrappedTypeName] ?? unique[lookupName.generic?.name ?? ""] ?? typeFromModule(lookupName.unwrappedTypeName, modules: modules)
```
ilyapuchka:
```swift
unique[lookupName.unwrappedTypeName] 
  ?? lookupName.generic.flatMap({ unique[$0.name] }) 
  ?? typeFromModule(lookupName.unwrappedTypeName, modules: modules)`
```

Sourcery/Models/TypeName.swift
```diff
@@ -78,24 +79,22 @@ public final class TypeName: NSObject, AutoCoding, AutoEquatable, AutoDiffable,
                     unwrappedTypeName = name.drop(first: "ImplicitlyUnwrappedOptional<".characters.count, last: 1)
                 }
                 self.unwrappedTypeName = unwrappedTypeName.bracketsBalancing()
+                self.isGeneric = unwrappedTypeName.contains("<") && unwrappedTypeName.characters.last == ">"
```
ilyapuchka: I guess if you move this after `else` branch you can do this just once using `unwrappedTypeName`

Sourcery.xcodeproj/project.pbxproj
```diff
@@ -787,7 +788,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "diff \"${PODS_PODFILE_DIR_PATH}/Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [ $? != 0 ] ; then\n    # print error to STDERR\n    echo \"error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\" >&2\n    exit 1\nfi\n";
+			shellScript = "diff \"${PODS_ROOT}/../Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [ $? != 0 ] ; then\n    # print error to STDERR\n    echo \"error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\" >&2\n    exit 1\nfi\n";
```
ilyapuchka: Do we have different versions of CocoaPods?

Sourcery.xcodeproj/project.pbxproj
```diff
@@ -33,6 +33,7 @@
 		09C6B0D41E11C1C500246CDF /* ClassSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 09C6B0D31E11C1C500246CDF /* ClassSpec.swift */; };
 		09F64DA41EA40A0E0039FF80 /* TemplateContext.swift in Sources */ = {isa = PBXBuildFile; fileRef = 09F64DA11EA3EEC00039FF80 /* TemplateContext.swift */; };
 		09F64DA51EA40A310039FF80 /* TemplateContext.swift in SwiftTemplates */ = {isa = PBXBuildFile; fileRef = 09F64DA11EA3EEC00039FF80 /* TemplateContext.swift */; };
+		0A6A64361EA8DC9800326383 /* GenerationContext.swift in Sources */ = {isa = PBXBuildFile; fileRef = C894A1AC1E1931500098327C /* GenerationContext.swift */; };
```
ilyapuchka: This type is not used any more, please remove it (it's removed in one of others prs, but it's not merged yet)

Sourcery/Models/TypeName.swift
Joannis: ` 'self' captured by a closure before all members were initialized` is what happens then

Sourcery/Models/TypeName.swift
```diff
@@ -204,6 +205,37 @@ public final class TypeName: NSObject, AutoCoding, AutoEquatable, AutoDiffable,
...
 }
 
+/// Descibes Swift class
+public final class GenericType: NSObject, SourceryModel {
+    /// the name of the base type.
+    ///
+    /// `Array<Int>`'s GenericType.name is `Array`
+    public let name: String
+    
+    public let referencedTypes: [Type]
```
Joannis: I'll add them now :) Still trying to figure out some edge-case scenarios, definitely didn't think about the two annotations.

Sourcery/Models/TypeName.swift
ilyapuchka: This can be solved by using local variable for `unwrappedTypeName`

Sourcery/Parsing/Composer.swift
```diff
@@ -98,6 +98,30 @@ struct Composer {
         }
 
         let lookupName = typeName.actualTypeName ?? typeName
+        
+        if lookupName.generic == nil && lookupName.isGeneric {
```
Joannis: Good thing my code is being checked now :) Never worked with code like this before (NSCoding/NSCoder). Working on my own projects is a lot easier being familiar with the code base.

Sourcery.xcodeproj/project.pbxproj
Joannis: I think so. I think I noticed someone (maybe you) running the beta in something like a podspec. I'm running the latest off of homebrew.

Joannis:
The main problems I'm facing are the unit tests. Since it parses generics now we have additional variables being set causing a failure.

Joannis:
When I try to adjust the tests according to the generics' expectations I'm getting errors roughly like `expected nil, received GenericType(...)` even though the test specifically requires a `GenericType(...)`. Do the unit tests get cached?

Joannis:
The unit tests work locally now. I have to say, I learned to appreciate XCTest now that I've worked with Quick.

krzysztofzablocki:
Quick is nice, our test suite is a tad complicated now, the development has been really fast

Joannis:
I've been having a tonne of strange issues. Some appear/disappear randomly after reopening the project in Xcode. But I'm glad it works for you.

Joannis:
@ilyapuchka @krzysztofzablocki Anything I need to take care of before leaving this PR as is?

Joannis:
Danger is failing with a non-descriptive error, can one of you fix this?

ilyapuchka:
It requires CHANGELOG entry I guess. I will look again at this pr on a weekend

Joannis:
Sure. Thanks for your time!

Sourcery/Parsing/Composer.swift
```diff
@@ -91,13 +91,74 @@ struct Composer {
...
+        guard genericCharacters.count > 1, genericCharacters.removeLast() == ">" else {
+            return
+        }
+
+        var genericTypeStrings = [String]()
+
+        var subGenericsCounter = 0
+        var characterBuffer = [Character]()
+
+        for character in genericCharacters {
```
ilyapuchka: Very hard to follow this method and I'm not sure it's doing the right thing. There are already few extension for `String` that will help and I think the whole method can be simplified to something like this (also applying my previous comment about using `TypeParameter` type):

```swift
    private func parseGenericType(typeName: TypeName) {
        let genericComponents = typeName.unwrappedTypeName
            .characters.split(separator: "<", maxSplits: 1)
            .map({ String($0).stripped() })

        guard genericComponents.count == 2 else {
            return typeName
        }

        let name = genericComponents[0]
        let typeParametersString = genericComponents[1].dropLast()

        typeName.generic = GenericType(name: name, typeParameters: parseGenericTypeParameters(typeParametersString))
    }
    
    private func parseGenericTypeParameters(typeParametersString: String) -> [TypeParameter] {
        let typeParametersList = typeParametersString
            .commaSeparated()
            .map({
                let typeName = TypeName($0.stripped())
                resolveGenerics(typeName: typeName)
                return TypeParameter(typeName: typeName)
        })
    }
```

Resolving type parameter type should happen outside this method, similar to resolving tuple elements types. This way parsing generic type matches closer to other similar methods.

Sourcery/Parsing/Composer.swift
```diff
@@ -112,8 +173,8 @@ struct Composer {
                 tupleElement.type = resolveType(typeName: tupleElement.typeName, containingType: containingType, unique: unique, modules: modules, typealiases: typealiases)
             }
         }
-        return unique[lookupName.unwrappedTypeName]
-            ?? typeFromModule(lookupName.unwrappedTypeName, modules: modules)
+
+        return unique[lookupName.unwrappedTypeName] ?? unique[lookupName.generic?.name ?? ""] ?? typeFromModule(lookupName.unwrappedTypeName, modules: modules)
```
ilyapuchka: This still will not work if generic type comes from a different module, it should be something like this:

```swift
        let typeName = lookupName.generic?.name ?? lookupName.unwrappedTypeName
        return unique[typeName] ?? typeFromModule(typeName, modules: modules)
```

SourceryTests/Parsing/ComposerSpec.swift
```diff
@@ -151,7 +151,7 @@ class ParserComposerSpec: QuickSpec {
...
-                                TupleElement(name: "generic", typeName: TypeName("Dictionary<String, Dictionary<String, Float>>", dictionary: DictionaryType(name: "Dictionary<String, Dictionary<String, Float>>", valueTypeName: TypeName("Dictionary<String, Float>", dictionary: DictionaryType(name: "Dictionary<String, Float>", valueTypeName: TypeName("Float"), keyTypeName: TypeName("String"))), keyTypeName: TypeName("String")))),
+                                TupleElement(name: "generic", typeName: TypeName("Dictionary<String, Dictionary<String, Float>>", generic: GenericType(name: "Dictionary", referencedTypeNames: [TypeName("String"), TypeName("Dictionary<String, Float>", generic: GenericType(name: "Dictionary", referencedTypeNames: [TypeName("String"), TypeName("Float")]))]), dictionary: DictionaryType(name: "Dictionary<String, Dictionary<String, Float>>", valueTypeName: TypeName("Dictionary<String, Float>", generic: GenericType(name: "Dictionary", referencedTypeNames: [TypeName("String"), TypeName("Float")]), dictionary: DictionaryType(name: "Dictionary<String, Float>", valueTypeName: TypeName("Float"), keyTypeName: TypeName("String"))), keyTypeName: TypeName("String")))),
```
ilyapuchka: There should be also dedicated set of tests for parsing generic types that check that it parses generics correctly with all possible edge cases (known/unknown types as type parameters, nested generics), resolves types from different modules and typealiases correctly.

Sourcery/Parsing/Composer.swift
```diff
@@ -91,13 +91,74 @@ struct Composer {
...
+
     private func resolveType(typeName: TypeName, containingType: Type?, unique: [String: Type], modules: [String: [String: Type]], typealiases: [String: Typealias]) -> Type? {
         let actualTypeName = self.actualTypeName(for: typeName, containingType: containingType, unique: unique, typealiases: typealiases)
         if let actualTypeName = actualTypeName, actualTypeName != typeName.unwrappedTypeName {
             typeName.actualTypeName = TypeName(actualTypeName)
         }
 
         let lookupName = typeName.actualTypeName ?? typeName
+
+        if lookupName.isGeneric, lookupName.generic == nil {
```
ilyapuchka: I wonder if we should actually consider dictionary and array types as generics, given that they are already represented with corresponding properties. The thing that parsing results will be different depending on how these types are defined (the same applies to optionals too btw) - if they are defined with a full syntax the will be parsed both as generic and dictionary/array, if shorthand syntax is used they will be not recognised as generics. Also generic types data is duplicated in this case and additional parsing seems to me as unneeded. And most likely arrays and dictionaries should be handled separately in templates - they can not be always handled as generics because of shorthand syntax case and they are usually treated as a special case of types.

So I would put parsing generics in the very end after parsing type as array, dict and tuple.
What do you think @krzysztofzablocki ?

SourceryTests/Parsing/ComposerSpec.swift
Joannis: Sure! I'll add those.

Sourcery/Parsing/Composer.swift
Joannis: Why wouldn't it? I'm using it from a different module now.

ilyapuchka: because type from different module can be referenced with module name, which is not part of `unique` keys. `typeFromModule` solves that.

Sourcery/Models/TypeName.swift
```diff
@@ -192,6 +209,45 @@ public final class TypeName: NSObject, AutoCoding, AutoEquatable, AutoDiffable,
 
 }
 
+/// Descibes Swift class
+public final class GenericType: NSObject, SourceryModel {
+    /// the name of the base type.
+    ///
+    /// `Array<Int>`'s GenericType.name is `Array`
+    public let name: String
```
ilyapuchka: I think we should also provide `Type` property that will reference to this generic type definition.

Sourcery/Parsing/Composer.swift
Joannis: I understand that, I didn't test this here. But if `OtherModule.ForeignType` is referenced in the generics it wouldn't work for both lookups in `unique`. So it would look in `typeFromModule` regardless.

ilyapuchka: But in this case you will search for types from modules by the full type name which includes type parameters, not just by name of generic type, so `typeFromModule` will also fail, right?

ilyapuchka: That makes me think that `unwrappedTypeName` should consider not just `Optional` type as it does now, but also drop type parameters of any generic types. Then your logic will work here.

Joannis: You're right! Didn't think of that.

SourceryTests/Parsing/ComposerSpec.swift
Joannis: I'm not sure how to programmatically (using Quick) test multiple modules. I'd need to scan 2 source folders.

ilyapuchka: You don't need to scan anything for these tests, just create types programmatically and pass them to `Composer`. There are tests like that (to test types from different modules) in `ComposerSpecs`

Sourcery/Parsing/Composer.swift
krzysztofzablocki: agreed

Sourcery/Parsing/Composer.swift
Joannis: Oh, sorry! I'm used to writing performant parsers for formats like JSON, BSON and the like. I naturally keep performance as much as possible in the back of my mind as a habit and already gave in to using Strings rather than bytes 😂

Sourcery/Parsing/Composer.swift
Joannis: I specifically thought Array and Dictionary generics were useful. You might not care about them being a Dictionary or Array but only care about them being generic and what they contain.

ilyapuchka: Then I guess we should do that while parsing array/dictionary, just creating a `GenericType` from information that we parsed already.

ilyapuchka:
@Joannis do you need any help with this pr? I can pick it up if thats ok

Joannis:
@ilyapuchka I've been sick the last 2 days, if you want and have the time you can pick it up.

danielr:
Any news on this? Can I help with something? I'm currently trying to add some code for each property of type `Set<T>`, where `T` is a subclass of some class. As far as I can see, there is currently no way to do that, right?

ilyapuchka:
@danielr currently we don't fully support generic types, we just provide a boolean flag if type is generic or not. Contribution will be very welcome! There are some review comments that were not addressed and branch is out of date with master already.

ilyapuchka:
Will try to resurrect it now, would be cool if we finally have it implemented.

SourceryRuntime/Sources/TypeName.swift
```diff
@@ -199,6 +216,81 @@ public final class TypeName: NSObject, AutoCoding, AutoEquatable, AutoDiffable,
...
+public final class GenericType: NSObject, SourceryModel {
+    /// the name of the base type.
+    ///
+    /// `Array<Int>`'s GenericType.name is `Array`
+    public let name: String
+
+    /// This generic type parameters
+    public let typeParameters: [GenericTypeParameter]
+
+    public init(name: String, typeParameters: [GenericTypeParameter] = []) {
```
SourceryBot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/krzysztofzablocki/Sourcery/blob/b311fa99d89127c124845d746d1ce7450600eb9c/SourceryRuntime/Sources/TypeName.swift#L275"></span>Undocumented symbol.</td>
    </tr>
  </tbody>
</table>

ilyapuchka:
@krzysztofzablocki I addressed all of mine comments and updated tests =) If you want - take a look. There are probably some corner cases with nested types and type aliases that are not tested, but I guess we can extend it later. There are still plenty of things to add like associated types, generic constraints etc.
And thanks to @Joannis for all the work for that!

Joannis:
You're welcome! This is ages ago, though!

kdawgwilk:
@ilyapuchka Some merge conflicts left to address but I am excited to see this merged in!

ilyapuchka:
🎉

Issue #10, opened 2016-12-12
How should we approach sharing generic templates with the community?


ilyapuchka:
We can expect more useful templates to pop up in future and we even might to create few to drive and test some new features in generator or even just for documentation purposes. It would be cool if there will be a way to share these templates in some known place.

So what then will be the strategy to distribute and share such templates? 
Subfolder in root of this repo? Or maybe through separate repository? Via subspec (I don't remember accurately if CocoaPods supports repos without targets)? Any other options?

krzysztofzablocki:
Was about to write about this, you beat me to the punch. 
I think a reasonable approach would be to:

- Add a folder for generic Templates in this repository
- Add generation tests to always keep Templates up-to date with tool version thus avoiding breaking them with updates
- Bundle templates with the binary distribution of the Pod, so that people can use our default implementation if they wish

Makes sense?

bejar37:
I think it would be really interesting to have a project that people can pull in as a dependency (like SwiftGen, via Cocoapods or Carthage) which provides some basic templates for protocols like `AutoEquatable` or `AutoHashable`. 
Do you think this is feasible? Another future direction for that would be providing a way for each user of that repository  to add their own templates that are less widely applicable (maybe JSON parsing, lenses, etc.). 

Basically, I think it would be cool for the example templates in #123 to be usable for an end-user. I'm not sure if that belongs in Sourcery itself, or in a project which has Sourcery as a dependency.

krzysztofzablocki:
@bejar37 that's the idea, I made those files separate so we can bring them in as embedded templates later on, I didn't bundle them yet because we are far from 1.0 and a lot of changes are happening

bejar37:
Oops, I see that I hadn't read your previous comment throughly @krzysztofzablocki . Sorry!

krzysztofzablocki:
Actually, lets embeed them in binary distrubtion, people can choose to use them or not :)

PR #252, opened 2017-04-25
Made AutoCases uses let constants instead of computed vars

AliSoftware:
This feels more swifty and efficient (computed only once)

BTW, not sure if:

* there are any tests for templates provided in the repo (I didn't found any, and tests are passing locally)
* there's something to regenerate in the documentation in case the doc quotes this template somewhere (and if so, how to)


ilyapuchka:
Hey @AliSoftware !
To update the docs just run `jazzy --clean` (will add this to the pr template)
I guess the only thing to update in docs is `guides/Enum cases.md` where it has example output

AliSoftware:
Thx. It has been some time since I've installed jazzy, have a way outdated version that doesn't work anymore. Will probably not have much time to reinstall it properly to regen docs before NSBudapest/CraftConf, so feel free to run it and push there, would probably be faster (otherwise I'll try not to forget to do it after the conferences)

AliSoftware:
Also, do you expect a CHANGELOG entry for changes in the template examples?

ilyapuchka:
Yep, just put it under "Internal changes" maybe

ilyapuchka:
"there are any tests for templates provided in the repo (I didn't found any, and tests are passing locally)" - that's a good point actually. If we want to present these templates as ready for use, not just as an example and starting point for writing something on your own, we definitely need to test it. 
But maybe it will make more sense to separate it in a submodule (maybe using sub spec or something) and write more integration tests there based on these templates. It also popped up several times lately that people try to use Sourcery with these templates though they don't even need them. We discussed that back in #10, maybe we need to reopen it.

AliSoftware:
@ilyapuchka you could get some inspiration from our SwiftGen/templates repo that contains both templates and fixtures. We now tests those templates on CircleCI to ensure the code generated by those templates compile properly. Would be nice to do the same for Sourcery.

PR #254, opened 2017-04-26
Add lowerFirst function in stencil.

yageek:
Generating enum cases from type name would require the first letter to
be lowercase.


krzysztofzablocki:
Thanks for contributing, we already have `lowerFirstWord` since we inherit filters from https://github.com/SwiftGen/StencilSwiftKit, but this works a little different, since its just first letter not whole word, so I'm ok merging it.

Could you add changelog entry to make Danger happy?

yageek:
Sure. Should I regenerate jazzy documentation too?

krzysztofzablocki:
@yageek yeah, good catch :) I should add it to danger when I fly back home

AliSoftware:
Shouldn't we actually migrate that new filter into StencilSwiftKit instead? (I know I'm quite late on releases but there's normally a new one coming soon to catch up)

krzysztofzablocki:
We'll need to look through all filters and move the generic ones to stencil swift kit

PR #255, opened 2017-04-26
AutoEquatable/Hashable for protocols uses computedVariables 

### LINKED ISSUE #253: AutoEquatable/Hashable changes for protocols
We recently switched to `storedVariables` for the templates, but I think for protocols we should still use computed ones, otherwise nothing will actually be generated right now. thoughts @ilyapuchka ?

krzysztofzablocki:
closes #253


ilyapuchka:
Can we use `macro` node from StencilSwiftKit for these additions? They look exactly the same as for not-procols, isn't it?

krzysztofzablocki:
they do, but I never used the macro node so will have to look at it when I'm back from CraftConf

PR #257, opened 2017-04-26
Support init functions for AutoMockable protocol

Antondomashnev:
The template update fixes the incorrectly generated method for `init` methods.

For the following protocol:
```swift
protocol FolderEventsListener: AutoMockable {
    /// Designated initializer to create new instance of FolderEventsListener
    ///
    /// - Parameter folderPath: absolute folder path to watch
    /// - Parameter filter: filter for received event
    /// - Parameter fileWatcherFactory: factory to create underlying watcher
    init(folderPath: String, filter: FolderEventFilter?, fileWatcherFactory: FileWatcherFactory)
}
```

Before was:
```swift
func init(folderPath: String, filter: FolderEventFilter?, fileWatcherFactory: FileWatcherFactory) {

    initReceivedArguments = (folderPath: folderPath, filter: filter, fileWatcherFactory: fileWatcherFactory)
}
```

After:
```swift
required init(folderPath: String, filter: FolderEventFilter?, fileWatcherFactory: FileWatcherFactory) {

    initReceivedArguments = (folderPath: folderPath, filter: filter, fileWatcherFactory: fileWatcherFactory)
}
```

### DIFF
--- a/Templates/AutoMockable.stencil
+++ b/Templates/AutoMockable.stencil
@@ -8,6 +8,14 @@
     {% endif %}
 {% endmacro %}
 
+{% macro methodCalledVariableName method %}
+{{ method.shortName }}Called
+{% endmacro %}
+
+{% macro methodReturnValueVariableName method %}
+{{ method.shortName }}ReturnValue
+{% endmacro %}
+
 {% for type in types.based.AutoMockable %}{% if type.kind == 'protocol' %}
 {% if not type.name == "AutoMockable" %}
 class {{ type.name }}Mock: {{ type.name }} {
@@ -19,9 +27,13 @@ class {{ type.name }}Mock: {{ type.name }} {
 {% for method in type.allMethods %}
     //MARK: - {{ method.shortName }}
 
-    {% if not method.isInitializer %}var {{ method.shortName }}Called = false{% endif %}
-    {% if method.parameters.count == 1 %}var {{ method.shortName }}Received{% for param in method.parameters %}{{ param.name|upperFirst }}: {{ param.typeName.unwrappedTypeName }}?{% endfor %}{% else %}{% if not method.parameters.count == 0 %}var {{ method.shortName }}ReceivedArguments: ({% for param in method.parameters %}{{ param.name }}: {% if param.typeAttributes.escaping %}{{ param.unwrappedTypeName }}{% else %}{{ param.typeName }}{% endif %}{% if not forloop.last %}, {% endif %}{% endfor %})?{% endif %}{% endif %}
-    {% if not method.returnTypeName.isVoid and not method.isInitializer %}var {{ method.shortName }}ReturnValue: {{ method.returnTypeName }}!{% endif %}
+    {% if not method.isInitializer %}var {% call methodCalledVariableName method %} = false{% endif %}
+    {% if method.parameters.count == 1 %}
+    var {{ method.shortName }}Received{% for param in method.parameters %}{{ param.name|upperFirst }}: {{ param.typeName.unwrappedTypeName }}?{% endfor %}
+    {% else %}{% if not method.parameters.count == 0 %}
+    var {{ method.shortName }}ReceivedArguments: ({% for param in method.parameters %}{{ param.name }}: {% if param.typeAttributes.escaping %}{{ param.unwrappedTypeName }}{% else %}{{ param.typeName }}{% endif %}{% if not forloop.last %}, {% endif %}{% endfor %})?
+    {% endif %}{% endif %}
+    {% if not method.returnTypeName.isVoid and not method.isInitializer %}var {% call methodReturnValueVariableName method %}: {{ method.returnTypeName }}!{% endif %}
 
 {% if method.isInitializer %}
     required {{ method.name }} {
@@ -30,9 +42,9 @@ class {{ type.name }}Mock: {{ type.name }} {
 {% else %}
     func {{ method.name }}{% if not method.returnTypeName.isVoid %} -> {{ method.returnTypeName }}{% endif %} {
 
-        {{ method.shortName }}Called = true
+        {% call methodCalledVariableName method %} = true
         {% call methodReceivedParameters method %}
-        {% if not method.returnTypeName.isVoid %}return {{ method.shortName }}ReturnValue{% endif %}
+        {% if not method.returnTypeName.isVoid %}return {% call methodReturnValueVariableName method %}{% endif %}
     }
 {% endif %}
 {% endfor %}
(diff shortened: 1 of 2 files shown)


Templates/AutoMockable.stencil
```diff
@@ -13,13 +15,19 @@ class {{ type.name }}Mock: {{ type.name }} {
...
-        {% if not method.returnTypeName.isVoid and not method.shortName == "init" %}return {{ method.shortName }}ReturnValue{% endif %}{% endif %}
+    {{ method.shortName }}Called = true
+    {%if method.parameters.count == 1 %}{{ method.shortName }}Received{% for param in method.parameters %}{{ param.name|upperFirst }} = {{ param.name }}{% endfor %}{% else %}{% if not method.parameters.count == 0 %}{{ method.shortName }}ReceivedArguments = ({% for param in method.parameters %}{{ param.name }}: {{ param.name }}{% if not forloop.last%}, {% endif %}{% endfor %}){% endif %}
+    {% if not method.returnTypeName.isVoid %}return {{ method.shortName }}ReturnValue{% endif %}{% endif %}
     }
+    {% else %}
+    required init({% for param in method.parameters %}{% if param.argumentLabel == nil %}_{% else %}{{ param.argumentLabel }}{% endif %}{% if not param.argumentLabel == param.name %} {{ param.name }}{% endif %}: {{ param.typeName }}{% if not forloop.last %}, {% endif %}{% endfor %}) {
```
ilyapuchka: Doesn't `method.name` do the same? It should contains the method name with all arguments labels and parameters names already. If it's not - it's a bug in method parsing.

Antondomashnev:
> Looks good to me, but I would take a chance to improve template to use latest features if it's fine for you. I commented on some of them but I guess there might be more.

@ilyapuchka I think this a great idea 👍 , thanks for the hint in the comments, I will check it later today.

krzysztofzablocki:
we definitely will need to add some simple tests for the bundled templates, that might be as simple as having the Sample Code folder and Expected result (for each template) and veryfing it matches

Templates/AutoMockable.stencil
Antondomashnev: Perfect looks much better. I've updated it in fa38c87

Templates/AutoMockable.stencil
```diff
@@ -13,13 +15,19 @@ class {{ type.name }}Mock: {{ type.name }} {
     {% if method.parameters.count == 1 %}var {{ method.shortName }}Received{% for param in method.parameters %}{{ param.name|upperFirst }}: {{ param.typeName.unwrappedTypeName }}?{% endfor %}{% else %}{% if not method.parameters.count == 0 %}var {{ method.shortName }}ReceivedArguments: ({% for param in method.parameters %}{{ param.name }}: {% if param.typeAttributes.escaping %}{{ param.unwrappedTypeName }}{% else %}{{ param.typeName }}{% endif %}{% if not forloop.last %}, {% endif %}{% endfor %})?{% endif %}{% endif %}
     {% if not method.returnTypeName.isVoid and not method.shortName == "init" %}var {{ method.shortName }}ReturnValue: {{ method.returnTypeName }}!{% endif %}
 
+    {% if not method.shortName == "init" %}
```
Antondomashnev: Updated in e3d359e

Antondomashnev:
According to the covering the templates with tests, I've created an issue #261 and will try to propose a PR soon.

Templates/AutoMockable.stencil
```diff
@@ -1,3 +1,5 @@
+import Cocoa
```
ilyapuchka: Why Cocoa, maybe just import Foundation? Does it even needed?

Templates/AutoMockable.stencil
```diff
@@ -9,17 +11,23 @@ class {{ type.name }}Mock: {{ type.name }} {
...
+    {% if method.isInitializer %}
+    required {{ method.name }} {
 
-        {% if not method.shortName == "init" %}{{ method.shortName }}Called = true{% endif %}
-        {%if method.parameters.count == 1 %}{{ method.shortName }}Received{% for param in method.parameters %}{{ param.name|upperFirst }} = {{ param.name }}{% endfor %}{% else %}{% if not method.parameters.count == 0 %}{{ method.shortName }}ReceivedArguments = ({% for param in method.parameters %}{{ param.name }}: {{ param.name }}{% if not forloop.last%}, {% endif %}{% endfor %}){% endif %}{% if not method.returnTypeName.isVoid %}{% endif %}
-        {% if not method.returnTypeName.isVoid and not method.shortName == "init" %}return {{ method.shortName }}ReturnValue{% endif %}{% endif %}
+        {%if method.parameters.count == 1 %}{{ method.shortName }}Received{% for param in method.parameters %}{{ param.name|upperFirst }} = {{ param.name }}{% endfor %}{% else %}{% if not method.parameters.count == 0 %}{{ method.shortName }}ReceivedArguments = ({% for param in method.parameters %}{{ param.name }}: {{ param.name }}{% if not forloop.last%}, {% endif %}{% endfor %}){% endif %}{% endif %}
```
ilyapuchka: can we extract that and the code above for not-initialiser case into macro node (from SwiftStencilKit)? Look very similar to me.

ilyapuchka: I think `var` is missing here after first `if`
p.s. yep, we need tests 😌

Templates/AutoMockable.stencil
```diff
@@ -9,17 +11,23 @@ class {{ type.name }}Mock: {{ type.name }} {
...
     }
+    {% else %}
+    func {{ method.name }}{% if not method.returnTypeName.isVoid %} -> {{ method.returnTypeName }}{% endif %} {
 
+        {{ method.shortName }}Called = true
+        {%if method.parameters.count == 1 %}{{ method.shortName }}Received{% for param in method.parameters %}{{ param.name|upperFirst }} = {{ param.name }}{% endfor %}{% else %}{% if not method.parameters.count == 0 %}{{ method.shortName }}ReceivedArguments = ({% for param in method.parameters %}{{ param.name }}: {{ param.name }}{% if not forloop.last%}, {% endif %}{% endfor %}){% endif %}
```
ilyapuchka: this part also looks very similar to the one above

Templates/AutoMockable.stencil
Antondomashnev: Nope, everything is fine, because it's already declared variable 😉

Templates/AutoMockable.stencil
Antondomashnev: Yep, thanks for good point. I've creates a macro for it in ca84457

Templates/AutoMockable.stencil
Antondomashnev: I think we don't actually need it. Removed in 9e0ffbc.

ilyapuchka:
Looks cool!

PR #265, opened 2017-04-29
Required methods and initializers fix

### LINKED ISSUE #263: Extract initialiser properties in a separate object
It might be useful to have access to such attributes of initialisers as `required` or `convenience`, but they only make sense for initialisers, not for methods in general, so we can add `initialiser: InitialiserType` property, similar to how we approach dictionary and array types.

ilyapuchka:
Resolves #263 
There are just two boolean properties that can be extracted to a separate type, so it feels like it does not worth it. So this PR adds missing property for `required` attribute and fixes typo to make initializer related properties consistent.


Sourcery/Models/Method.swift
```diff
@@ -150,11 +150,23 @@ public final class Method: NSObject, SourceryModel, Annotated {
...
-    /// Whether method is a convenience initializer
+    // sourcery: skipEqaulitey, skipDescription, skipCoding, skipJSExport
+    /// :nodoc:
+    @available(*, deprecated: 0.7, message: "Use isConvenienceInitializer instead")
     public var isConvenienceInitialiser: Bool {
         return attributes[Attribute.Identifier.convenience.name] != nil
     }
 
+    /// Whether method is a convenience initializer
+    public var isConvenienceInitializer: Bool {
```
ilyapuchka: Changed here from `s` to `z` as we use `z` in two other properties, but I'm ok with changing `z` to `s` too, @krzysztofzablocki let me know if you prefer other option =)

PR #269, opened 2017-04-29
added info about `--version` and `--help` flags in docs

ilyapuchka:


krzysztofzablocki:
but we already have `--version` and had since my first release, so this only adds it to docs

ilyapuchka:
You are right! Completely missed it. I'll remove unneeded code then and will leave only docs update.

PR #270, opened 2017-05-01
Fixed documentation : Annotation filtering

iDonJose:
I found out that the documentation is giving a wrong example of annotation filtering.
The correct format is 'varToFilter|annotated:"myAnnotation"', and it is the one used in Sourcery/Templates/Equality.stencil for example.


ilyapuchka:
Thanks! 👍

PR #277, opened 2017-05-02
Release docs

### LINKED ISSUE #271: Include docset in repository or ignore it?
Currently docset is included in at docs path in the repo together with its zipped version. That creates conflicts with each doc update. So I would say we should ignore this zip for good. Also we maybe should not keep docset in the repo either and just attach zip archive, mentioned before, to the release along with binary zip. This operation (updating and uploading docset zip) can be part of release process. Any intermediate docs update will be still available via html version of the docs on github. What do you think @krzysztofzablocki ?

ilyapuchka:
Resolves #271 
@krzysztofzablocki what would be the best way to update html docs when doing release? I'm still thinking it might be better to have docs up to date with master, not waiting for the release.


krzysztofzablocki:
I had few people come to me with questions why X doesn't work and pointing to docs, I feel like outside of us almost no one uses Master version of Sourcery so its better to have published docs reflect last released version.

What we can do is that we just add new task to `Rake` `release` phase, similar how we check whether you updated changelog, you could check whether doc command is not generating a new diff meaning its up to date, if it does we fail the release and print out the command to update them

Antondomashnev:
> you could check whether doc command is not generating a new diff meaning its up to date, if it does we fail the release and print out the command to update them

@krzysztofzablocki would it be a better automation if we just update the docs during the release?

krzysztofzablocki:
yeah good idea, we'll need to make automatic add, commit and push then so need to make sure we don't mess it up :)

ilyapuchka:
@krzysztofzablocki how should we do that? I guess it should create a pr for release with updated docs and version bump. But that will break release process into few steps - create pr, manually merge it (unless there is automated way to do that), then actually create release.

Antondomashnev:
@ilyapuchka do you think it can potentially break something? I mean, I would treat this part as a part of bumping podspec or updating changelog.

ilyapuchka:
Not break, but we can simply end up with lots of undocumented things. I've tried danger plugin for jazzy that should warn if there is something undocumented in the pr, but it does not seem to work.

Antondomashnev:
> we can simply end up with lots of undocumented things

I guess without danger plugin it's anyhow manual process to check if everything is documented, so maybe it can be checked as a part of the PR review process, so the undocumented code will be catched during the feature PR, not during the actual release process. But I'm still new to the codebase, so I'm not sure if I get the process and your idea right 😄

ilyapuchka:
Sure, but there is Danger for that =) Doing it manually is very easy to miss such things. I would expect it to at least check if undocumented rate changed, at most outline undocumented symbols.

I've tried [danger-jazzy](https://github.com/fwal/danger-jazzy), maybe I missed something, cause I'm not familiar with Danger plugins. I guess second pair of hands to try that will not hurt.

If we do that then we simply generate and commit updated docs after version bump and we can be sure it's complete (again if not for some reason - danger will warn us in release pr).
Sounds reasonable?

Antondomashnev:
@ilyapuchka totally reasonable 👍 
I've checked briefly the plugin and I see that it checks the `undocumented.json` file (you can check it [here](https://github.com/fwal/danger-jazzy/blob/master/lib/jazzy/plugin.rb#L82)) generated by jazzy. So I guess to be able to check the documentation with the plugin on each PR we should generate documentation on each PR.

Sourcery/Models/Type.swift
```diff
@@ -169,7 +169,6 @@ public class Type: NSObject, SourceryModel, Annotated {
     }
 
     // sourcery: skipEquality, skipDescription
-    /// Contained types groupd by their names
     public private(set) var containedType: [String: Type] = [:]
```
SourceryBot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/krzysztofzablocki/Sourcery/blob/ddc9c6e1e0adc6e87e9384086e73bfe41f08c53a/Sourcery/Models/Type.swift#L172"></span>Undocumented symbol.</td>
    </tr>
  </tbody>
</table>

SourceryBot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/krzysztofzablocki/Sourcery/blob/ddc9c6e1e0adc6e87e9384086e73bfe41f08c53a/Sourcery/Models/Type.swift#L172"></span>Undocumented symbol.</td>
    </tr>
  </tbody>
</table>

Antondomashnev: Perfect 🎉

ilyapuchka: Except that it supposed to post warning only. But I guess we can just fail always 😈

SourceryBot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/krzysztofzablocki/Sourcery/blob/207907ddfbf23b7ab54c3a515a767e235cee84a9/Sourcery/Models/Type.swift#L172"></span>Undocumented symbol.</td>
    </tr>
  </tbody>
</table>

Antondomashnev: I prefer the strategy to fail earlier just because it will push contributor better to keep the docs updated 😄

Dangerfile
```diff
@@ -19,3 +19,5 @@ no_changelog_entry = !git.modified_files.include?("CHANGELOG.md")
 if has_app_changes && no_changelog_entry && not_declared_trivial
   fail("Any changes to library code need a summary in the Changelog.")
 end
+
+jazzy.check fail: :modified
```
Antondomashnev: @ilyapuchka shouldn't we check also for `:all` - added and modified?

> Except that it supposed to post warning only.
It is failing, because you pass `fail:`, we can pass `warn:` if we wanna show the warning only. But as I mentioned already I vote for fail 😄

circle.yml
```diff
@@ -15,4 +15,5 @@ test:
       - swift build
   post:
       - bash <(curl -s https://codecov.io/bash)
+      - gem install jazzy && jazzy --clean
```
Antondomashnev: Maybe we can add jazzy for the `Gemfile` as a source of all `Sourcery` ruby dependencies 😉

Dangerfile
ilyapuchka: I tried with warn first, it posted both failure and warning messages, so I changed it to fail. I think just `modified` should be enough. If we would want for any reason to keep something undocumented, other PRs should not be affected by that. And if everything is documented there will be no problem at all.

ilyapuchka:
So, what we have now:

- danger checks if there are any undocumented changes in PRs
- docs are updated and checked for modifications before release
- docset is copied to release zip
- docsets folder is ignored in the repo to avoid merge conflicts

What is not automated:
- pushing updated docs to master during release

Do we need anything else? @krzysztofzablocki @Antondomashnev

Dangerfile
Antondomashnev: @ilyapuchka sorry, I'm not sure I got the reason to not check `added` files for documentation.

> If we would want for any reason to keep something undocumented, other PRs should not be affected by that. And if everything is documented there will be no problem at all.

As I understood the point (but I'm not sure) - we could want to add new file to Sourcery and keep it undocumented so the danger will pass the PR. But I think once somebody decides to modify that file - danger will fail the PR. So As I've mentioned in the beginning of the comment, I'm confused 😕

ilyapuchka: You are right. Then I guess it does not matter much if it's modified or all

.gitignore
```diff
@@ -75,3 +75,7 @@ fastlane/Preview.html
 fastlane/screenshots
 fastlane/test_output
 Sourcery/CodeGenerated/TypedSpec.generated.swift
+
+# jazzy
+docs/docsets
+docs/undocumented.json
```
Antondomashnev: If this is ignored, will Danger Plugin be able to check it?

ilyapuchka: we run jazzy before danger, so it will regenerate everything, including this file

PR #279, opened 2017-05-02
Enum cases annotations parsing fix.

eugeneego:


ilyapuchka:
Hey, @eugeneego ! Thanks for reporting an issue and submitting the pr, very appreciated. But I think that the same can be achieved with simpler changes. As I understand the problem is in the way we currently extract the string where to parse annotations in, right?. Then instead of doing what we are doing now we should extract everything up to previous case boundary, or if we are parsing first case, up to enum body start.

I came up with this diff that passes the tests that you updated - https://gist.github.com/ilyapuchka/5068b523b842d902e44b269f6ec5de5c

I noticed that you actually was doing something very similar in cb61be6 , but then changed your approach.

eugeneego:
Hi @ilyapuchka
As you noticed at the start I've tried simple approach. But it didn't cover all my cases. Then I decided to write more complex parser based on SourceKit information.
Your approach is partly based on strings, I've tried to rely on SourceKit information without string magic.
If your code will pass the new tests I'll be happy to update PR with it. :)

ilyapuchka:
Yes, all the tests are passing.

eugeneego:
Should I update the current PR or create a new one with these improved code?

ilyapuchka:
I don't mind if you apply this path in this pr 👍

Sourcery/Parsing/FileParser.swift
```diff
@@ -573,33 +578,46 @@ extension FileParser {
...
-            annotations = AnnotationsParser(contents: body).all
+        var annotationRangeStart: Int64?
+        if let previousCase = `enum`.cases.last, let previousCaseKeyRange = Substring.key.range(for: previousCase.__underlyingSource) {
+            // if current enum case is not the first one in containing enum
+            // get everything after the previous case declaration
+            annotationRangeStart = previousCaseKeyRange.offset + previousCaseKeyRange.length
+        } else if let enumBody = Substring.body.range(for: `enum`.__underlyingSource) {
+            // if current enum case is the first one in containing enum
+            // get everything after enum body start
+            annotationRangeStart = enumBody.offset
```
krzysztofzablocki: what if I had a variable in enum and added annotation to it, would you parse it as part of the first enum case? 

The rules for parsing annotations I had originally were that you go backwards from the current code element and stop immediately if there is a non comment line, they were only tweaked for the enum associated value annotations

eugeneego: Yes, you're right. Test with included variables fails.

ilyapuchka: @krzysztofzablocki with enum cases it's a bit different because enum case can have inline annotations for associated value which can add up to the case annotations, IIRC the reason for different logic here. Also AnnotationsParser is designed to parse line by line which does not fit inline annotations. I would say that we need to revisit it at some point cause it's not good that we now have different kinds of annotations parsing logic in different places.

I do believe though this particular case can be handled with minor changes, not with reworking all the annotations parsing just for enum cases. We can just get the reference not to the previous enum case but to the previous substructure on the same level in the containing enum case, no matter if it's enum case, method, var or inner type.

krzysztofzablocki: @ilyapuchka Yeah, only associated values annotation requires more complex parsing, the line by line works everywhere else, thats what I meant by above comment

ilyapuchka:
Thanks @eugeneego for working on that and pointing out the issue. The issue is solved with #282 with fixes in AnnotationsParser.

PR #282, opened 2017-05-04
Improved inline annotations parsing

### LINKED ISSUE #278: Wrong parsing of enums that contain multiple annotations separated by comma.
```swift
// sourcery: transformer, type = "String"
enum Kind {
    // sourcery: value = "NEWS"
    case news
    // sourcery: value = "ARTICLE"
    case article
    // sourcery: value = "TWEET"
    case tweet
}
```

The sample produces the following enum structure:
- `Kind` contains annotations: `[ "transformer": 1,  "type": "String"]`
- `news` element contains annotations: `[ "value": "NEWS" ]`
- `article` element contains annotations: `[ "value": [ "NEWS", "ARTICLE" ] ]`
- `tweet` element contains annotations: `[ "value": [ "NEWS", "ARTICLE", "TWEET" ] ]`

Each enum element contains all previous annotations up to comma symbol from `Kind` annotation.

ilyapuchka:
Resolves #278 and #283 . I used the same tests added in #279 with some minor additions
Everything related to annotations parsing is now moved back to `AnnotationsParser`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Fixed detecting computed properties
 - Fixed typo in `isConvenienceInitialiser` property
 - Fixed creating cache folder when cache is disabled
+- Fixed parsing multiple enum cases annotations
 
 ### Internal changes
 
--- a/Sourcery/Parsing/FileParser.swift
+++ b/Sourcery/Parsing/FileParser.swift
@@ -161,7 +161,7 @@ final class FileParser {
             }
 
             type.isGeneric = isGeneric(source: source)
-            type.annotations = parseAnnotations(from: source)
+            type.annotations = annotations.from(source)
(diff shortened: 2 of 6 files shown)


bolismauro:
@ilyapuchka This also solves multiple annotations in enums which are not ignored in the current master branch

```
enum T: TestProtocol {

  // sourcery: type= one, two, three
  case one
}
```

Returns 0 annotations, while

```
enum T: TestProtocol {

  // sourcery: type= one
  case one
}
```
returns "type"

krzysztofzablocki:
the first one should have 3 annotations according to our annotation rules so that seems like a bug

bolismauro:
@krzysztofzablocki correct. By using this PR's branch it works as expected while in master it doesn't work
I just wanted to report it :)

PR #286, opened 2017-05-04
Add tests for bundled templates

Antondomashnev:
Eventually, this PR will bring the tests for the bundled templates to be on the safe side, while editing them. Currently, it's a work in progress state as I've implemented proposal for the first template so I would like to get a feedback from @krzysztofzablocki, @ilyapuchka on the overall approach 😄  

I've added a new project to the workspace called `Templates` that has only `TemplatesTests` target. The project consists of 4 main folders:
* Context - files to be used to generate a code for;
* Generated - files generated by Sourcery;
* Expected - files that will be used as master files to compare with generated.

Each template is supposed to be tested against various object's interfaces, for example currently implemented test for `AutoCases` template checks the generated code against three different enums. 

All tests are supposed to be in the file `TemplatesTests.swift` and each `it` of the spec is a unit to check templated.

P.S. the PR is big and has plenty of files because I've added a new project and Pods for it 😄


Antondomashnev:
@krzysztofzablocki @ilyapuchka hey, during the implementation of tests for `AutoEquatable` I've found that it doesn't work as expected for `Protocol` but I'm not sure if the reason of it is expected behavior or not.

In template we have:
```
{% call compareVariables type.computedVariables %}
```
where `type` is a protocol. But for example for protocol:
```swift
protocol AutoEquatableProtocol: AutoEquatable {
    var width: Double { get }
    var height: Double { get}
}
```
it doesnt return `computedVariables`. I've been looking through the Sourcery code and have found that `FileParser` has the following logic to check whether the variable is computed or not:
```swift
let computed = !containedInProtocol && (
            (setter == nil && !constant) ||
            (setter != nil && !body.isEmpty && hasPropertyObservers == false)
        )
```
So there is condition that variable is not contained in the protocol and this is the exact reason to return empty array for `computedVariables`. So the question is - is there a reason for it or it's a bug?

ilyapuchka:
@Antondomashnev I guess for protocols we can not say if variable is computed or not, as it declares only access level for getter/setter, not storage option. We can do that for extensions of protocols, because there properties are actually declared and only computed properties are allowed, but not just for protocols.

Antondomashnev:
@ilyapuchka thanks, let me open the issue because I want to continue discussion but I guess it's out of the scope if this PR 😄

Antondomashnev:
@ilyapuchka ready to be checked 🚀

Templates/Tests/Expected/AutoLenses.expected
```diff
@@ -0,0 +1,78 @@
+//
+//  AutoLenses.expected.swift
+//  Templates
+//
+//  Created by Anton Domashnev on 16.05.17.
+//  Copyright © 2017 Pixle. All rights reserved.
+//
```
ilyapuchka: should we strip these header comments?

Antondomashnev: You're right, cleaned up in 62f36ae

Templates/Tests/Context/AutoHashable.swift
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    // Method with optional return value
+    func books(sharedWith name: String) -> String? {
+        return nil
+    }
+}
+
+/// It should generate correct code for general class
+class AutoHashableClass: AutoHashable {
```
Antondomashnev: I'm not actually sure, I think current configuration maybe contains some duplicates, but from the other side it limits the code that you have to check if the specific test case fails.
@ilyapuchka what so you think?

SourceryRuntime/Sources/Diffable.generated.swift
```diff
@@ -1,6 +1,7 @@
 // Generated using Sourcery 0.6.0 — https://github.com/krzysztofzablocki/Sourcery
 // DO NOT EDIT
 
+
```
Antondomashnev: Thanks, removed in 46dfad7.

Templates/Templates/AutoMockable.stencil
```diff
@@ -14,44 +16,37 @@ import AppKit
...
 {{ method.shortName }}ReturnValue
 {% endmacro %}
 
 {% for type in types.based.AutoMockable %}{% if type.kind == 'protocol' %}
 {% if not type.name == "AutoMockable" %}
 class {{ type.name }}Mock: {{ type.name }} {
-
     {% for variable in type.allVariables %}
-    var {{ variable.name }}: {{ variable.typeName }}{% if not variable.isOptional %}{% if variable.isArray %} = []{% endif %}{% if variable.isDictionary %} = [:]{% endif %}{% endif %}
+    var {{ variable.name }}: {{ variable.typeName }}{% if not variable.isOptional %}{% if not variable.isArray and not variable.isDictionary %}!{% else %}{% if variable.isArray %} = []{% endif %}{% if variable.isDictionary %} = [:]{% endif %}{% endif %}{% endif %}
```
Antondomashnev: Done in abe1f02

ilyapuchka:
Thanks Anton! I'm sure this will improve the quality of the tool. And we need to keep contributing to this suite with edge cases.

,
```diff
@@ -0,0 +1 @@
+M
```
Liquidsoul: Sorry to comment on an already merged PR, but what is the purpose of this newly introduced file? 😊

Antondomashnev: @Liquidsoul hah, thanks for the catch. Definitely was introduced by my mistake. I will remove it from master.

ilyapuchka: Oops =) But I had much more funny issues like that in my experience =)

Issue #285, opened 2017-05-04
Add check for the app version in rake check_versions task


Antondomashnev:
This will help us to prevent the issue like #284. Maybe the [Xcodeproj](https://github.com/CocoaPods/Xcodeproj) from CocoaPods team could help to get the `Current Project Version` from build settings.

PR #290, opened 2017-05-05
Check bundle version in rake release task.

Antondomashnev:
This PR brings the feature proposed in #285. I've decided to use a simple shell script with `xcodebuild` instead of bringing another dependency of [Xcodeproj gem](https://github.com/CocoaPods/Xcodeproj).


Antondomashnev:
@krzysztofzablocki are we waiting for 1 approval, 2 approvals?

PR #292, opened 2017-05-08
Don't add default clause if enum has only one case

iandundas:
Preventing "Default will never be executed" warnings from Xcode.

For example: 

```swift
enum Result {
    case decrypted(key: Key, refreshToken: String)
}
```

This change will output: 

```swift

extension Result: Equatable {}
public func == (lhs: Result, rhs: Result) -> Bool {
    switch (lhs, rhs) {
    case (.decrypted(let lhs), .decrypted(let rhs)): 
        if lhs.key != rhs.key { return false }
        if lhs.refreshToken != rhs.refreshToken { return false }
        return true
    }
}
```

instead of: 

```swift
extension Result: Equatable {}
public func == (lhs: Result, rhs: Result) -> Bool {
    switch (lhs, rhs) {
    case (.decrypted(let lhs), .decrypted(let rhs)): 
        if lhs.key != rhs.key { return false }
        if lhs.refreshToken != rhs.refreshToken { return false }
        return true
    default: break; // warning here
    }
}
```

### DIFF
--- a/Templates/AutoEquatable.stencil
+++ b/Templates/AutoEquatable.stencil
@@ -56,7 +56,7 @@ extension {{ type.name }}: Equatable {}
         {% endif %}
         {% endif %}
     {% endfor %}
-    default: return false
+    {% if type.cases.count > 1 %}default: return false {% endif %}
     }
 }
 {% endfor %}


Antondomashnev:
Hi @iandundas, thanks for your first contribution, looks goot to me 🎉

PR #293, opened 2017-05-09
Fix inline annotations parsing when access level is defined

iDonJose:
It appears that an inline annotation followed by the access level won't parse right.

For example :
```
// sourcery: anAnnotation
class Foo {}
```
works well, annotation 'anAnnotation' is parsed, whereas :
```
// sourcery: anAnnotation
internal class Foo {}
```
won't parse any annotation.

The bug was introduced with inline annotations improvements, #282.

### DIFF
--- a/Sourcery/Parsing/Utils/AnnotationsParser.swift
+++ b/Sourcery/Parsing/Utils/AnnotationsParser.swift
@@ -92,19 +92,14 @@ internal struct AnnotationsParser {
         // `case` is not included in the key of enum case definition, so we strip it manually
         prefix = prefix.trimmingSuffix("case").trimmingCharacters(in: .whitespaces)
 
-        // Strips the access level if any
-        if let accessLevel = AccessLevel.all
-            .map({ $0.rawValue })
-            .filter({ prefix.hasSuffix($0) })
-            .first {
-
-            prefix = prefix.trimmingSuffix(accessLevel).trimmingCharacters(in: .whitespaces)
-        }
+        var inlineCommentFound = false
 
         while !prefix.isEmpty {
             guard prefix.hasSuffix("*/"), let commentStart = prefix.range(of: "/*", options: [.backwards]) else {
                 break
             }
+
+            inlineCommentFound = true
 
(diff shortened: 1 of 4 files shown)


Sourcery/Parsing/Utils/AnnotationsParser.swift
```diff
@@ -92,6 +92,15 @@ internal struct AnnotationsParser {
         // `case` is not included in the key of enum case definition, so we strip it manually
         prefix = prefix.trimmingSuffix("case").trimmingCharacters(in: .whitespaces)
 
+        // Strips the access level if any
+        if let accessLevel = AccessLevel.all
```
ilyapuchka: I guess the issue happens not only when using access level, but i.e. using `final` or any attributes on the same line with declaration. We can of course add it to the list of prefixes to trim, but it does not look like bulletproof solution...

ilyapuchka:
@iDonJose I think what we should do here is to not set `stop` to `true` if there are no inline comments at all. This `stop` means (sorry for not having doc comments for that) that there is some code preceding inline annotation and we can not make sure that it does not belong to other declaration in the same line (I think it's possible technically, but this will make the code much more complicated). So what we need to do is after the loop that parses inline comments to check if there was any inline comment found and if not - do not return at this point. Something like this https://gist.github.com/ilyapuchka/f313782cc11d3be9536f5081677fae8d

iDonJose:
@ilyapuchka , totally agree with you. Looking for a specific attribute and trimming it isn't resilient enough. Your implementation is way better.

By the way, with your changes the line before where 'case' was trimmed :
`prefix = prefix.trimmingSuffix("case").trimmingCharacters(in: .whitespaces)`
isn't needed anymore.

ilyapuchka:
`trimmingSuffix("case")` is still there. It's needed because `case` is not part of the key and will be included in the key prefix, so the prefix will not end with the comment and we will assume that we should not parse annotations any further.

iDonJose:
Ok I got it ! Thanks for the explanation.
I added your changes and it is all good for me.

PR #296, opened 2017-05-10
do not write into file logs when skipping/removing files

### LINKED ISSUE #294: Sourcery watch mode erase the content of some of my generated files
I have the following `.sourcery.yml` file:

```yml
sources:
    - Paparazzi
templates:
    - Templates
output: Paparazzi
```

A couple of templates (AutoEquatable, etc.).

When I run `sourcery`, It works A1. When I run `sourcery -w`, the content of my `AutoEquatable.generated.swift` file is replaced with

```swift
// Generated using Sourcery 0.6.0 — https://github.com/krzysztofzablocki/Sourcery
// DO NOT EDIT

STATUS:
Skipping Paparazzi/AutoList.generated.swift as it is empty.
```

I'm using Sourcery `0.6.0` and I just checked, this doesn't happen with sourcery 0.5.9. My templates are available [here](https://github.com/Palleas/Paparazzi/blob/master/Templates/AutoSluggable.stencil), in case that helps. 

Let me know if I can provide additional informations.

ilyapuchka:
Resolves #294

@krzysztofzablocki I also noticed that `w` flag is used both for `disableCache` and `watch` options. Shouldn't they be different?

### DIFF
--- a/Sourcery.xcodeproj/xcshareddata/xcschemes/Sourcery.xcscheme
+++ b/Sourcery.xcodeproj/xcshareddata/xcschemes/Sourcery.xcscheme
@@ -89,18 +89,6 @@
       <CommandLineArguments>
          <CommandLineArgument
             argument = "--help"
-            isEnabled = "NO">
-         </CommandLineArgument>
-         <CommandLineArgument
-            argument = "--config"
-            isEnabled = "YES">
-         </CommandLineArgument>
-         <CommandLineArgument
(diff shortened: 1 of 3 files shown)


ilyapuchka:
I guess I'll just remove this flag then 🙂

Antondomashnev:
@ilyapuchka I think next time these kind of changes deserve the CHANGELOG update 😉

Issue #297, opened 2017-05-11
AutoEquatable template doesn't generate correct code for protocols


Antondomashnev:
This is the continuation of the question from #286 about `AutoEquatable` protocol issue.
It seems the commit 8bd91e275f09e79122c3303935d3a1853fe1d0d6 brought the bug, because `computedVariables ` is always an empty array.

I think we can use the same logic for protocols as we use in `AutoMockable` template:
```
{% for variable in type.allVariables %}
    ...
{% endfor %}
```
I think we just need to add new condition that the variable actually has getter access.
@ilyapuchka @krzysztofzablocki thoughts?

ilyapuchka:
Do you mean to add check for getter in a template?

Antondomashnev:
@ilyapuchka I mean to change template to iterate through `allVariables` but filter only for those that have getter access 😄

ilyapuchka:
wouldn't they all have it?...

Antondomashnev:
I thought that you can have `{ set }` for protocol variable, seems you're right. One more thing, will `allVariables` also return static variables?

ilyapuchka:
Yes, it will, so they should be filtered by `isStatic` (latest version of Stencil supports `where` or we can use filter `!static`)

Antondomashnev:
cool, I will prepare the fix 👌

PR #298, opened 2017-05-12
Use all variables for autoequatalbe template for protocols; #trivial

Antondomashnev:
This fixes the issue #297.
@ilyapuchka we've discussed to filter static variables in generated code for protocols, but we don't need to filter them 😄 
For example:
```
protocol Foo {
  var one: Int { get }
  static var name: String { get }
}
```

We actually need to compare `name` in the `==` function. @ilyapuchka do you agree?


Antondomashnev:
@ilyapuchka also for `AutoHashable` 😄

PR #300, opened 2017-05-12
Move to CocoaPods 1.2.1

garriguv:
bundle install failed on my machine because I use ruby 2.4.1 and the
json gem version 1.8.3 was not compatible.

This problem has now been fixed:
https://github.com/flori/json/issues/303

While we're at it, why not update all the gems 😃 

Since we're updating the gems, let's move to the latest CocoaPods as well.


garriguv:
The Cocoapods update triggers a lot of changes to the project file after running `bundle exec pod install`.

I'm looking at them and may push the changes to this branch, or create a new pull-request.

Gemfile
```diff
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 gem 'cocoapods'
 gem 'cocoapods-update-if-you-dare', '~> 0.1.1'
+gem 'colored'
```
Antondomashnev: I'm not sure if we need a new dependency here

garriguv: We actually do, since `cocoapods-update-if-you-dare` does not specify this dependency explicitly. And `cocoapods` moved to `colored2`: https://github.com/CocoaPods/Xcodeproj/pull/463

I opened a PR on the project itself to fix it: https://github.com/Ashton-W/cocoapods-update-if-you-dare/pull/2

Antondomashnev: Ah, thanks, @garriguv, I've missed the commit description 😄 And thanks for the PR to the CocoaPods plugin 👍 Maybe the @Ashton-W can merge it soon and release a new version, so we again won't require the `colored` dependency specified in the gem.

Antondomashnev: But if it's not released soon, the PR looks cool and we could merge it as is. Does it sound fine for you @garriguv ?

garriguv: Yeah that sounds good. I'll open another PR to remove `colored` once https://github.com/Ashton-W/cocoapods-update-if-you-dare/pull/2 is merged and a new version of the gem is released.

Ashton-W: I just released `0.2.0` for you :)

Antondomashnev: @garriguv I think there is no need for another PR. Just remove colored here from this one and update bundle once again and we're good to go 🚀

garriguv: Done 🚀

Antondomashnev: Amazing, congratulations on your first contribution 🎉

PR #304, opened 2017-05-18
Implemented sourcery:file annotation for source code

### LINKED ISSUE #303: Support sourcery:file annotation in source files
Currently `sourcery:file` is used in templates to redirect output, but we can support `//sourcery:file` annotation in source code that will be the same as wrapping the whole content with `//sourcery:begin` and `//sourcery:end`
For simplicity annotation will ignore everything before it so it should be defined at the very top of the file.

ilyapuchka:
Resolves #303


Sourcery/Parsing/Utils/AnnotationsParser.swift
```diff
@@ -178,14 +190,19 @@ internal struct AnnotationsParser {
...
             upperBound = commentLine.characters.indices.endIndex
             insideBlock = true
         } else if commentLine.contains("sourcery:end") {
             return .end
+        } else if commentLine.contains("sourcery:file") {
+            lowerBound = commentLine.range(of: "sourcery:file:")?.upperBound
+            upperBound = commentLine.characters.indices.endIndex
+            insideFileBlock = true
         } else {
             insideBlock = false
```
Antondomashnev: I think with the pre-set `insideBlock` variable to `false` this can be removed.

Sourcery/Parsing/Utils/AnnotationsParser.swift
```diff
@@ -198,7 +215,7 @@ internal struct AnnotationsParser {
 
         if let lowerBound = lowerBound, let upperBound = upperBound {
             let annotations = AnnotationsParser.parse(line: commentLine.substring(with: lowerBound ..< upperBound))
-            return insideBlock ? .begin(annotations) : .annotations(annotations)
+            return insideBlock ? .begin(annotations) : insideFileBlock ? .file(annotations) : .annotations(annotations)
```
Antondomashnev: I would split this for the sake of readability, but it's up to you 😄

SourceryRuntime/Sources/Annotations.swift
```diff
@@ -2,6 +2,23 @@ import Foundation
 
 /// Describes annotated declaration, i.e. type, method, variable, enum case
 public protocol Annotated {
-    /// Annotations, that were created with // sourcery: annotation1, other = "annotation value", alterantive = 2
+    /**
```
Antondomashnev: Thanks for the documentation 👍

PR #308, opened 2017-05-19
Make Array.parallelMap's block non-escaping.

kareman:


Antondomashnev:
Hi @kareman, thanks for your first contribution 🚀 Could you please explain the motivation behind this change?

kareman:
Yes, the `transform` closure is not retained after the `parallelMap` function returns, it is not stored anywhere else, so it should not be marked `@escaping`. It had to be marked `@escaping` before Swift 3.1 because `queue.async` only takes escaping closures, but since we know those closures will not be used again after `group.wait()` we can use the newly added `withoutActuallyEscaping` to work around this.

Issue #317, opened 2017-05-23
Support in-out parameters


samskiter:
I had a quick look here too: https://github.com/krzysztofzablocki/Sourcery/blob/8583c039db54b5956a6b02d8447ce1ed00647941/SourceryTests/Parsing/FileParser%20%2B%20AttributesSpec.swift

Looks like the parser doesn't pull out the 'inout' attribute on params

ilyapuchka:
Thanks for reporting. `inout` is not an attribute, it's part of the type and SourceKit returns it as a typename, so I guess we are currently not just do not parse it but also will not properly infer type of such parameter. Not sure if we should model it with attribute, probably better to do it with boolean property in `MethodParameter`.

samskiter:
Thanks a Boolean would suit my needs fine

Antondomashnev:
Hi @samskiter, the flag `inout` is on master and will be added to the next release.

PR #319, opened 2017-05-24
Inout method parameters

Antondomashnev:
This is the implementation of an enhancement from #317. So far flag in `MethodParameter` as suggested by @ilyapuchka looks enough for it 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Check if Current Project Version match version in podspec in release task
 - Improved swift templates performance
 - Added `// sourcery:file` annotation for source code
+- Added `inOut` flag for `MethodParameter`
 
 ### Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
 
 - Fixed inserting generated code inline automatically at wrong position
 
+### New Features
+
+- Added `inout` flag for `MethodParameter`
+
+
 ## 0.6.1
 
 ### New Features
@@ -21,7 +26,6 @@
 - Check if Current Project Version match version in podspec in release task
 - Improved swift templates performance
 - Added `// sourcery:file` annotation for source code
-- Added `inOut` flag for `MethodParameter`
 
 ### Bug fixes
 
--- a/Sourcery/Parsing/FileParser.swift
+++ b/Sourcery/Parsing/FileParser.swift
@@ -533,11 +533,14 @@ extension FileParser {
 
     internal func parseParameter(_ source: [String: SourceKitRepresentable]) -> MethodParameter? {
         guard let (name, _, _) = parseTypeRequirements(source),
-            let type = source[SwiftDocKey.typeName.rawValue] as? String else { return nil }
+              let type = source[SwiftDocKey.typeName.rawValue] as? String else {
+            return nil
+        }
 
+        let inOut = type.hasPrefix("inout ")
         let typeName = TypeName(type, attributes: parseTypeAttributes(type))
         let defaultValue = extractDefaultValue(type: type, from: source)
-        let parameter = MethodParameter(name: name, typeName: typeName, defaultValue: defaultValue, annotations: annotations.from(source))
+        let parameter = MethodParameter(name: name, typeName: typeName, defaultValue: defaultValue, annotations: annotations.from(source), inOut: inOut)
         parameter.setSource(source)
         return parameter
     }
--- a/Sourcery/Parsing/FileParser.swift
+++ b/Sourcery/Parsing/FileParser.swift
@@ -537,10 +537,10 @@ extension FileParser {
             return nil
         }
 
-        let inOut = type.hasPrefix("inout ")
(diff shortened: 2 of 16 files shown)


SourceryRuntime/Sources/Method.swift
```diff
@@ -13,6 +13,8 @@ public final class MethodParameter: NSObject, SourceryModel, Typed, Annotated {
 
     /// Parameter type name
     public let typeName: TypeName
+    
+    public let inOut: Bool
```
SourceryBot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/krzysztofzablocki/Sourcery/blob/53e18afc9022d39d786d9778fac266f80d74e31b/SourceryRuntime/Sources/Method.swift#L17"></span>Undocumented symbol.</td>
    </tr>
  </tbody>
</table>

Sourcery/Parsing/FileParser.swift
```diff
@@ -533,11 +533,14 @@ extension FileParser {
 
     internal func parseParameter(_ source: [String: SourceKitRepresentable]) -> MethodParameter? {
         guard let (name, _, _) = parseTypeRequirements(source),
-            let type = source[SwiftDocKey.typeName.rawValue] as? String else { return nil }
+              let type = source[SwiftDocKey.typeName.rawValue] as? String else {
+            return nil
+        }
 
+        let inOut = type.hasPrefix("inout ")
```
Antondomashnev: > Can inout be mixed with any other attributes that can be used for method parameters?
I thought about it and quick Googling didn't show anything. So I think for the first version we can keep it as just a boolean.

And I've also thought about putting it into `TypeName` and I have doubts whether it makes sense to put it there. As far as I know the `inout` could be the only method parameter in Swift language, if so - from my point of view it's only related to the method and to parameter itself, rather than the type of the parameter. But I would like to hear your opinion about it 😄

SourceryRuntime/Sources/Method.swift
```diff
@@ -35,23 +38,25 @@ public final class MethodParameter: NSObject, SourceryModel, Typed, Annotated {
...
     /// :nodoc:
-    public init(name: String = "", typeName: TypeName, type: Type? = nil, defaultValue: String? = nil, annotations: [String: NSObject] = [:]) {
+    public init(name: String = "", typeName: TypeName, type: Type? = nil, defaultValue: String? = nil, annotations: [String: NSObject] = [:], inOut: Bool = false) {
         self.typeName = typeName
         self.argumentLabel = name
         self.name = name
         self.type = type
         self.defaultValue = defaultValue
         self.annotations = annotations
+        self.inOut = inOut
```
Antondomashnev: Done in 80ab467

Sourcery/Parsing/FileParser.swift
ilyapuchka: I have the same doubts about where it belongs, I think if it would be treated not as a special kind of _type_, the same as `throws`, it would be an attribute, maybe. I'm fine with keeping it `MethodParameter`, but we still need to trim it in `unwrappedTypeName` of `TypeName`.

Antondomashnev: Yep, check the 6362fe9 😄

ilyapuchka:
Let's merge it after release?

Antondomashnev:
> Let's merge it after release?

Sure

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Check if Current Project Version match version in podspec in release task
 - Improved swift templates performance
 - Added `// sourcery:file` annotation for source code
+- Added `inOut` flag for `MethodParameter`
```
ilyapuchka: I guess now this needs to moved up to new release

SourceryRuntime/Sources/TypeName.swift
```diff
@@ -66,23 +66,24 @@ public final class TypeName: NSObject, AutoCoding, AutoEquatable, AutoDiffable,
             self.isOptional = false
         } else {
             name = name.bracketsBalancing()
+            let trimmedInoutName = name.trimmingPrefix("inout ")
```
ilyapuchka: I think it's better to trim whitespaces using character set. SourceKit only trims spaces around type name, which is `inout` is part of, so it can be any number of spaces in between.

ilyapuchka: I think this should change `name` value, it's later checked for prefix, so `inout` should be trimmed.

CHANGELOG.md
Antondomashnev: Sure thing

SourceryRuntime/Sources/TypeName.swift
Antondomashnev: I think we don't need `let trimmedInoutName` and yes we can just trim `inout` and whitespace. You can check 26019c3

PR #323, opened 2017-05-25
Fix auto inline code generation

### LINKED ISSUE #320: Incorrect insert position of `inline:auto` code
**Swift version:** Apple Swift version 3.1 (swiftlang-802.0.53 clang-802.0.42) (Xcode 8.3.2)  
**Sourcery version:** `0.6.0` and `master` (ae69829)

### What did I do?

I used the `inline:auto` feature to add code in existing structs in a unique file.

Here is the template:

```
{% for type in types.structs %}
// sourcery:inline:auto:{{ type.name }}.AutoInit
test
// sourcery:end
{% endfor %}
```

And the swift files with the structs, one with existing code to update and one where new code wille be inserted:
```swift
public struct User {
// sourcery:inline:auto:User.AutoInit
test
// sourcery:end
}

public struct Session {
}
```

### What did I expect to see?

```swift
public struct User {
// sourcery:inline:auto:User.AutoInit
test
// sourcery:end
}

public struct Session {
// sourcery:inline:auto:Session.AutoInit
test
// sourcery:end
}
```
### What did I see instead?

```swift
public struct User {
// sourcery:inline:auto:User
// sourcery:inline:auto:Session.AutoInit
test
// sourcery:end
.AutoInit
test
// sourcery:end
}

public struct Session {
}
```

### Comments

I suspect that the insert position for the second block of code is not updated after inserting the first one 🤔

ilyapuchka:
Resolves #320 
Inline annotations will not be removed from source any more during generation, only code will be removed.

### DIFF
--- a/Sourcery/Sourcery.swift
+++ b/Sourcery/Sourcery.swift
@@ -336,9 +336,8 @@ extension Sourcery {
 
     private func generate(_ template: Template, forParsingResult parsingResult: ParsingResult) throws -> String {
         guard watcherEnabled else {
-            var result = try Generator.generate(parsingResult.types, template: template, arguments: self.arguments)
-            result = try processFileRanges(for: parsingResult, in: result)
-            return try processInlineRanges(for: parsingResult, in: result)
+            let result = try Generator.generate(parsingResult.types, template: template, arguments: self.arguments)
+            return try processRanges(in: parsingResult, result: result)
         }
 
         var result: String = ""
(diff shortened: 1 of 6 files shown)


Sourcery/Sourcery.swift
```diff
@@ -349,7 +350,8 @@ extension Sourcery {
         }, finallyBlock: {})
 
         result = try processFileRanges(for: parsingResult, in: result)
-        return try processInlineRanges(for: parsingResult, in: result)
+        result = try processInlineRanges(for: parsingResult, in: result)
+        return TemplateAnnotationsParser.removingEmptyAnnotations(from: result)
```
Antondomashnev: These three lines is an exact duplication of the line in the guard above, maybe we can put them somewhere to have only in one place? 😄

PR #338, opened 2017-06-01
Mutating attribute

chriseidhof:
I added support for mutating attributes.

Thanks @ilyapuchka for giving me some guidance!

### DIFF
--- a/Sourcery/Parsing/FileParser.swift
+++ b/Sourcery/Parsing/FileParser.swift
@@ -702,7 +702,7 @@ extension FileParser {
                     var attributeRange = prefix.range(of: $0.description, options: .backwards)
                     // we expect all attributes to be prefixed with `@`
                     // but `convenience` and `required` attribute does not need it...
-                    if $0 == Attribute.Identifier.convenience || $0 == Attribute.Identifier.required {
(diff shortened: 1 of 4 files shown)


Sourcery/Parsing/FileParser.swift
```diff
@@ -702,7 +702,7 @@ extension FileParser {
                     var attributeRange = prefix.range(of: $0.description, options: .backwards)
                     // we expect all attributes to be prefixed with `@`
```
Antondomashnev: I think it makes sense to update comment as well 😄

ilyapuchka: Maybe it's time to add property like `requiresAtPrefix` (can't find a better name ^^) in the `Attribute` type and hide this check there? We can also reuse it in `name` property of `Attribute` type.

Antondomashnev: It will clean the code - makes sense 👍

chriseidhof:
Maybe `hasAtPrefix` would be a better name?

Antondomashnev:
> Maybe hasAtPrefix would be a better name?

For me, it sounds good.

ilyapuchka:
I'll merge this, will do this improvement with a next pr as I noticed that we also do not parse `final` attribute

ilyapuchka:
Thanks a lot for your contribution @chriseidhof !

Issue #342, opened 2017-06-01
Templates tests require Sourcery.app in bin directory


ilyapuchka:
@Antondomashnev running templates tests locally I noticed that they run Sourcery from `bin` folder as a build step, which is copied there only when building for release IIRC. Maybe we should adjust this to make it easier to run tests? Also I think it make sense to run templates tests together with unit tests during development, maybe through a separate scheme.

Antondomashnev:
> run templates tests together with unit tests during development, maybe through a separate scheme.

That's a good idea

> Maybe we should adjust this to make it easier to run tests?

I think the way works fine, but probably we just need to add a build step to build sourcery app and copy it into the bin. I think that should fit out needs.

Issue #243, opened 2017-04-19
Support include in swift templates


ilyapuchka:
We can support `include` function for swift templates similar to EJS (#242)

ilyapuchka:
Done in #385

PR #343, opened 2017-06-02
Template tests scheme;

Antondomashnev:
The PR has two main features:
* This improves the dependency on the Sourcery App for the `TemplatesTest` target.
* Introduces a new scheme `TestAll` that runs test for both projects - `Sourcery` and `TemplatesTest`

See #342 for more information.


ilyapuchka:
Did you make the new scheme shared? I also think we can just add template tests to the tests running by Sourcery scheme so that we never forget to run them.

Antondomashnev:
@ilyapuchka I was not sure about running Templates Tests along with the Sourcery tests in the `Sourcery` scheme, but as long it doesn't take any noticeable time to do so, I think that makes sense 👍 

Also, I had to update tests stubs folder and separate `Include` and `Partial` template in a different folder because apparently #243 broke the tests and the [master branch](https://circleci.com/gh/krzysztofzablocki/Sourcery/836) but Circle CI still reports Green.

Issue #334, opened 2017-05-31
Stored Variables Regression


calebd:
It looks like Type.storedVariables now includes both `private` and `fileprivate` members. This breaks the included `AutoEquatable` template. I tried adding accessibility filters which always resulted in an empty variables array.

ilyapuchka:
To filter variables by their access level there are `privateGet` and `privateSet` and others filters ([docs](https://cdn.rawgit.com/krzysztofzablocki/Sourcery/master/docs/writing-templates.html)):

```
{% for variable in variables|!privateGet|!fileprivateGet %}
```

or you can use `where`: 

```
{% for variable in variables where variable.readAccess != "private" and variable.readAccess != "fileprivate" %}
```

@Antondomashnev should we check other templates for such regression? Do you mind doing that? You could also maybe look at #342

Antondomashnev:
@ilyapuchka sure 👌

Antondomashnev:
@calebd could you please clarify the prior behavior? Did it include only `public` and `internal` variables?

ilyapuchka:
@Antondomashnev yes, we changed recently that private declarations are not filtered out any more (https://github.com/krzysztofzablocki/Sourcery/pull/201) with introduction of access level filters

calebd:
In `0.5.9`, `storedVariables` included both `public` and `internal` and included variables that had both `get` and `set` blocks. I relied on this behavior for my own `SourceryEquatable` template.

I have a type like this:

```
struct Thing {
    fileprivate _nested: Box<Thing>?

    var nested: Thing? {
        get { return _nested.value }
        set { _nested = newValue.map(Box.init) }
    }
}
```

Old Sourcery output:

```
func == (lhs: Thing, rhs: Thing) -> Bool {
    if lhs.nested != rhs.nested { return false }
    return true
}
```

New Sourcery output with no template changes:

```
func == (lhs: Thing, rhs: Thing) -> Bool {
    if lhs._nested != rhs._nested { return false }
    return true
}
```

 I have yet to find a set of filters or variables that mirrors that behavior

Antondomashnev:
@ilyapuchka @calebd thanks, I've opened the PR that should fix an issue with updated templates.

ilyapuchka:
Solved with #345

PR #345, opened 2017-06-02
Stored variables regression fix

Antondomashnev:
This updates the `AutoEquatable` and `AutoHashable` templates to skip `private` and `fileprivate` variables in generated code. Referenced issue #334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Bug fixes
 
 - Fixed inserting generated code inline automatically at wrong position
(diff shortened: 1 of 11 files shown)


ilyapuchka:
@Antondomashnev @krzysztofzablocki how do you think will it be valuable to introduce a filter like `minAccessLevel:someAccessLevel` that can be used to easily filter out declarations with "lower" access level, i.e. if passed `internal` it will filter out everything public and private, if passed `public` it will filter out everything but open declarations? For Swift templates we can just implement kindof `Comparable` on `AccessLevel` that will compare it with its string corresponding value.

Issue #348, opened 2017-06-08
Sourcery 0.6.1 default templates conflict with user generated templates


mergesort:
Starting in Sourcery 0.6.1, my AutoLenses template is generating extra lenses for types I don't have in my app, and is causing a compilation error on my CI.

My current build step for Sourcery looks like this:
```ruby
"${PODS_ROOT}/Sourcery/bin/sourcery" --sources "${SRCROOT}" --templates "${SRCROOT}/${PROJECT_NAME}/Sourcery-Templates" --output "${SRCROOT}/${PROJECT_NAME}/Sourcery-Generated"
```

I have my own `AutoLenses` template in my `~${SRCROOT}/${PROJECT_NAME}/Sourcery-Templates` folder, and it seems to be conflicting with the one bundled in `Pods/Sourcery/Templates/Tests/Generated`. 

One example of a lenses on a non-existent type being generated is this.

```swift
extension Room {

  static let peopleLens = Lens(
    get: { $0.people },
    set: { people, room in
       Room(people: people, name: room.name)
    }
  )
  static let nameLens = Lens(
    get: { $0.name },
    set: { name, room in
       Room(people: room.people, name: name)
    }
  )
}
```

I was only able to trace that type down to the `Pods/Sourcery/Templates/Tests/Generated` folder.

Let me know if there's any more information I can provide. Thanks!

ilyapuchka:
Hey @mergesort. `${SRCROOT}` contains `${PODS_ROOT}`, so Sourcery scans all source files from pods, include Sourcery own templates tests sources. Narrow your input folder to just your sources if you want to scan just them or you can use latest master with exclude configuration (it's not supported by CLI arguments, only by yaml configuration file)

ilyapuchka:
Another option can be that we would not include those sources into release archive. @Antondomashnev what do you think? I think we don't need them there.

Antondomashnev:
@ilyapuchka I completely agree to not include them. Will prepare a fix for that 👍

ilyapuchka:
Will it work if we re upload 0.6.1 archive without releasing new version?

Antondomashnev:
Why not? The link will remain the same in the podspec.

Antondomashnev:
@ilyapuchka I will repack the 0.6.1 release with only stencil templates files, however, I think you can also prepare the 0.7.0 (I thing it's semantically correct) release as we already have something to share such as:
```
Added inout flag for MethodParameter
[...]

PR #349, opened 2017-06-08
Exclude templates project files from release; #trivial

Antondomashnev:
This reflects the decision we made with @ilyapuchka in #348. As it may produce confusion for the user and it adds not much value we've decided to not include templates in the release.


ilyapuchka:
Should we exclude everything in Templates folder or just test source code? I was thinking only about second point. Until we have better option for sharing templates maybe we should include them in release for users who wish to use them and get their updates.

Antondomashnev:
Well, yes, you're right, let's copy and paste not templates, but `Templates/Templates`

PR #350, opened 2017-06-09
Fix `sources` parsing and add a ConfigurationSpec

Liquidsoul:
I noticed that my configuration file was not working with the latest `master`.
So I tracked and found that the support for excluded paths (#347) introduced a regression in `sources` parsing in `.sourcery.yml` files.

To prevent a future regression I have added a `ConfigurationSpec` with just tests exposing the regression.
As I am not quite used to Quick/Nimble, I do not know if my tests are properly named so any feedback is welcomed 😊


ilyapuchka:
Good initiative, @Liquidsoul , we need to contribute more in configs tests.

PR #351, opened 2017-06-09
Fix swiftlint warning (version 0.19.0)

Liquidsoul:
When compiling the project I noticed that there was some swiftlint warnings using swiftlint `0.19.0`.
So I decided to try and fix them all here 🔧 
There are only 3 left about the `force_unwrapping` rule (like [here](https://github.com/krzysztofzablocki/Sourcery/blob/1d9b09eddf450c9efbea5e7dda2f7c12346b342a/Templates/Tests/Context/AutoEquatable.swift#L133)). I did nothing about them because this is a problem in swiftlint because we are not force unwrapping 😁


Liquidsoul:
@Antondomashnev I am not sure that these warnings are fixed there 🤔
Anyway, I tried to reproduce this in the `master` version of swiftlint and it still reports the error but creating a test case scenario is not straightforward because just using the line is not enough. It appears that it needs more context to generate the warning. I'll let them know.

Antondomashnev:
@Liquidsoul sorry I've referenced another PR 😄 https://github.com/krzysztofzablocki/Sourcery/pull/343 is correct

Issue #361, opened 2017-06-13
Option to lock generated files


fpg1503:
I think a `--lock` or `--autoLock` option (that automatically locks all the output files to avoid someone accidentally trying to change them) would be really nice.

ilyapuchka:
Closing as there is PR for that

PR #362, opened 2017-06-13
Option to lock generated files

fpg1503:
Implements feature request #361


Sourcery/main.swift
```diff
@@ -67,8 +67,22 @@ fileprivate enum Validators {
...
-    static func isWriteable(path: Path) -> Path {
-        if path.exists && !path.isWritable {
+    static func isWritable(path: Path, autoLock: Bool) -> Path {
+        let pathExistsAndIsNotWritable = path.exists && !path.isWritable
+        let pathExistsAndWillNotBeWritableIfUnlocked: Bool = {
+            guard path.exists, autoLock, path.isImmutable else {
+                return false
+            }
+            
+            try! path.setImmutable(false)
```
ilyapuchka: Is it correct if we try to right to the folder that is already locked and it was not Sourcery who locked it.

Sourcery/Configuration.swift
```diff
@@ -106,19 +106,22 @@ struct Configuration {
     let templates: Paths
     let output: Path
     let args: [String: NSObject]
+    let autoLock: Bool
 
-    init(dict: [String: Any], relativePath: Path) {
+    init(dict: [String: Any], relativePath: Path, autoLock: Bool = false) {
```
Antondomashnev: I like the feature as well, but maybe it makes sense to have it true by default? I think the case to have not locked should be an edge case or as @ilyapuchka mentioned - the watch mode.

Sourcery/main.swift
fpg1503: The idea here is checking if the file would be writable if it were unlocked so we unlock it, check if it's writable and lock it again to try to avoid any side effects

Sourcery/Sourcery.swift
```diff
@@ -318,6 +320,18 @@ extension Sourcery {
 
         track("Generating code...", terminator: "")
         status = ""
+        
+        if autoLock && outputPath.exists {
+            do { try output.setImmutable(false) } catch { track("\(error)") }
```
fpg1503: I've tested and it works, the idea is: we unlock the file, write to it and than lock it again

ilyapuchka: This works with a "happy" path when file is successfully unlocked, but will silently do nothing in case of any error. I think we should  just log with `info` scope

Sourcery/main.swift
ilyapuchka: I got the idea, but if the path can be unlocked by the Sourcery it does not mean it's safe to do that. Is there any way to set owner of the file/dir and check if it's Sourcery? 
Also I don't think it's safe to use `try!` here. It can easily fail and we should not crash in this situation. It's probably better to wrap it with do-catch:

```
do {
  try path.setImmutable(false)
  let isWritableIfUnlocked = path.isWritable
  try path.setImmutable(true)
} catch {
...
}
```

fpg1503:
I'm working on some tests. Regarding Sourcery owning the files: it's *not possible* (not recommended) because a user must have the ownership and I'd say that creating a new user just for that is unnecessarily intrusive.

We can use `xattr` (extended file attributes) to store if a file was created by Sourcery, does that approach look ok for you?

I didn't try it with inline code generation, I'll look into it.

ilyapuchka:
@fpg1503 extended file attributes will do the work I think. Will it work only on macOS or on Linux too?

Sourcery/Configuration.swift
Antondomashnev: @ilyapuchka @fpg1503 what you guys think about making it true by default?

ilyapuchka:
@fpg1503 did you have any success with using extended file attributes for this feature?

krzysztofzablocki:
Any update on this @fpg1503 ?

ilyapuchka:
I wonder how it will work with APFS, don't know much about it.

Issue #354, opened 2017-06-11
0.7.1 and 0.7.0 print "Major.Minor.Patch" not the version when building with SPM


ilovezfs:
Homebrew test block fails with

```
==> /usr/local/Cellar/sourcery/0.7.0/bin/sourcery --version
/usr/local/Homebrew/Library/Homebrew/debrew.rb:11:in `raise'
MiniTest::Assertion: Expected /0\.7\.0/ to match "Major.Minor.Patch".
```

This is a regression from 0.6.1.

See https://github.com/Homebrew/homebrew-core/pull/14480.

ilyapuchka:
Thanks for reporting @ilovezfs 
Something went wrong during releasing 0.7.0, we will need to re-upload its binary. Can you reproduce Homebrew issue locally with this release? When doing `bin/sourcery --version` I got 0.6.1 (when it should be 0.7.0)

ilyapuchka:
@Antondomashnev I guess the issue with version is simply that in Rakefile `update_metadata` task is done after `build` task. It should the the other way around, right?

Antondomashnev:
@ilyapuchka may be, but then it should've been the same for 0.6.1 release as I used the same rake tasks.

Antondomashnev:
@ilyapuchka I think that makes sense because the binary is built with `old_metadata`. I will prepare the fix along with the other in #359

ilovezfs:
Unfortunately it looks like 0.7.1 still has this issue: https://jenkins.brew.sh/job/Homebrew%20Core%20Pull%20Requests/3039/version=sierra/console

```
==> /usr/local/Cellar/sourcery/0.7.1/bin/sourcery --version
Error: sourcery: failed
Expected /0\.7\.1/ to match "Major.Minor.Patch".
```

ilyapuchka:
@ilovezfs how to reproduce this locally?

ilovezfs:
The Homebrew formula downloads and unpacks https://github.com/krzysztofzablocki/Sourcery/archive/0.7.1.tar.gz and then does this:
```
    system "swift", "build", "-c", "release", "-Xswiftc", "-static-stdlib"
    bin.install ".build/release/sourcery"
    lib.install Dir[".build/release/*.dylib"]
```

ilyapuchka:
[...]

PR #363, opened 2017-06-15
Use Version.swift to build Sourcery with SPM;

Antondomashnev:
The PR introduces the new `Version.swift` structure that represents the version for `Sourcery` command line tool. Prior approach with only `build_settings` doesn't work due to #354.
The PR also contains the update in the `Rakefile` release process to also automate the version change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,22 @@
 
 ---
 
+## Master
+
+### Internal changes
+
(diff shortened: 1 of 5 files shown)


Antondomashnev:
> @Antondomashnev want to make 0.7.2 release with that?

@ilyapuchka yep, let's make homebrew great again 😄

Issue #355, opened 2017-06-11
Add support for array variables for AutoHashable


Antondomashnev:
Currently AutoHashable doesn't work for the objects with `Array` variables. For example:

```swift
struct TestResultsSectionDisplayInfo: AutoEquatable {
    let titleInfo: TestResultsSectionTitleDisplayInfo
    let itemInfos: [TestResultDisplayInfo]
    
    init(title: TestResultsSectionTitleDisplayInfo, items: [TestResultDisplayInfo] = []) {
        self.titleInfo = title
        self.itemInfos = items
    }
}
```

As a reference to implement support we can use this [answer](https://stackoverflow.com/a/31444592/2487302) on StackOverflow

PR #364, opened 2017-06-21
Hashable template improvements

Antondomashnev:
The PR reflects the issue #355 but it extends the initial issue with the following features:
1. Support for Array<Hashable> variables
2. Support for Dictionary<Hashable, Hashable> variables
3. Support for static variables inside the protocol as for other types we use only `storedVariables`


ilyapuchka:
@Antondomashnev awesome!

PR #366, opened 2017-06-27
Fix link for template in docs

sora0077:
`s|Sourcery/blob/master/Templates/|Sourcery/blob/master/Templates/Templates/|g`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 
 - Fixed linker issue when using Swift templates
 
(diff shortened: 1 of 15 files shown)


Antondomashnev:
Hi @sora0077, thanks for the catch 👍 
Could you please update the changelog under the new section like in the example below 
👇 
```
### Internal changes

- Fix link for template in docs
```

ilyapuchka:
👍

Issue #326, opened 2017-05-25
Default implementation cause method duplication


ManWithBear:
1. What version of Swift are you using (swift --version)?
    ```
    Apple Swift version 3.1 (swiftlang-802.0.53 clang-802.0.42)
    Target: x86_64-apple-macosx10.9
    Sourcery 0.6.0
    ```
2. What did you do?
    Protocol:
    ```swift
    protocol Magic {
        func doMagic(animated: Bool)
    }
    extension Magic {
        func doMagic(animated: Bool = true) { }
    }
    ```
    Template:
    ```
    {% for protocol in types.protocols %}
    {% for method in protocol.methods %}
    {{ method }}
    {% endfor %}
    {% endfor %}
    ```
    Generate:
    ```
    Method: name = doMagic(animated: Bool), shortName = doMagic, callName = doMagic, parameters = [MethodParameter: argumentLabel = Optional("animated"), name = animated, typeName = Bool, typeAttributes = [:], defaultValue = nil, annotations = [:]], returnTypeName = Void, actualReturnTypeName = Void, `throws` = false, `rethrows` = false, accessLevel = internal, isStatic = false, isClass = false, isInitializer = false, isFailableInitializer = false, isConvenienceInitialiser = false, annotations = [:], attributes = [:]
    Method: name = doMagic(animated: Bool = true), shortName = doMagic, callName = doMagic, parameters = [MethodParameter: argumentLabel = Optional("animated"), name = animated, typeName = Bool, typeAttributes = [:], defaultValue = Optional("true"), annotations = [:]], returnTypeName = Void, actualReturnTypeName = Void, `throws` = false, `rethrows` = false, accessLevel = internal, isStatic = false, isClass = false, isInitializer = false, isFailableInitializer = false, isConvenienceInitialiser = false, annotations = [:], attributes = [:]
    ```
3. What did you expect to see?
There should be just second method.

4. What did you see instead?
Two methods

Problem looks similar to #158, but here problem in default value for parameter.

Antondomashnev:
Hi @ManWithBear thanks for creating an issue, I think you're right. The second method should override the first one. I will take a closer look on soon.

ilyapuchka:
[...]

PR #374, opened 2017-07-10
Add reference to type where method or variable is defined

Antondomashnev:
This is a big PR and I expect some questions about implementation 😄 
The origin of the PR comes from #326 and as was proposed by @ilyapuchka I've tried to implement `definedInType` property to know the type where the `Method` or `Variable` is defined. For me, personally, the main goal was to understand if the `Method` was defined in extension or not.

So the main change is in the [Composer.swift](https://github.com/krzysztofzablocki/Sourcery/pull/374/files#diff-c64a46fccebead3994a9133b860dc773R28). I've added the setting `definedInType` property for methods and variables and it's done before the composer filters for unique types. That allows us to have a correct (from my perspective) result when parsing for example:
```
protocol Foo { func foo(bar: Int) }; extension Foo { func foo(bar: Int = 1) {} }
```

So the composer will output type with two methods with the following `definedInType` values:
1. `Protocol(name: "Foo", methods: [method, defaultedMethod])`
2. `Type(name: "Foo", accessLevel: .none, isExtension: true, methods: [defaultedMethod])`


krzysztofzablocki:
few questions on how this will work:

previous type accessors will list joined types? 
e.g. `types.all` will not contain extensions explicitly?  
e.g. `type.Foo` will contain both protocol and its extension methods combined?

Antondomashnev:
@krzysztofzablocki as far as I understood these three questions are about already implemented functionality. If so there is no changes to the existed files parsing apart of assigning value to the new property of `Method` and `Variable`

> types.all will not contain extensions explicitly?

There will be only type -> `Foo` and `types.Foo` will contain it's variables and methods + extension's.

Templates/Templates/AutoMockable.stencil
```diff
@@ -22,29 +22,31 @@ import AppKit
 {% if not type.name == "AutoMockable" %}
 class {{ type.name }}Mock: {{ type.name }} {
     {% for variable in type.allVariables %}
-    var {{ variable.name }}: {{ variable.typeName }}{% if not variable.isOptional %}{% if not variable.isArray and not variable.isDictionary %}!{% elif variable.isArray %} = []{% elif variable.isDictionary %} = [:]{% endif %}{% endif %}
+    var {%if variable.definedInType.isExtension %}extension_{% endif %}{{ variable.name }}: {{ variable.typeName }}{% if not variable.isOptional %}{% if not variable.isArray and not variable.isDictionary %}!{% elif variable.isArray %} = []{% elif variable.isDictionary %} = [:]{% endif %}{% endif %}
```
krzysztofzablocki: would it make sense to add a filter so that you can get all variables from either original type or extension?

Antondomashnev: @krzysztofzablocki I like the idea. Will do that 👍

Antondomashnev: btw I would to make use of macros more, but they always produce a result of calling macro as a new line - which is not super convenient in some cases in terms of code style.

Antondomashnev: @krzysztofzablocki it's done in 1936362

Sourcery/Parsing/FileParser.swift
```diff
@@ -144,16 +144,16 @@ final class FileParser {
             case .enumelement:
                 return parseEnumCase(source)
             case .varInstance:
-                return parseVariable(source, containedInProtocol: containingIn is SourceryProtocol)
+                return parseVariable(source, containedIn: containingIn as? Type)
```
ilyapuchka: lets just rename containedIn to definedIn maybe?

SourceryRuntime/Sources/Definable.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/// Describes that the object is defined in a context of some `Type`
+public protocol Defined: class {
```
ilyapuchka: We already have `Definition` protocol, we need to use just one to avoid confusion.

Sourcery/Parsing/FileParser.swift
```diff
@@ -445,7 +446,13 @@ extension FileParser {
         let writeAccessibility = setter.flatMap({ AccessLevel(rawValue: $0.replacingOccurrences(of: "source.lang.swift.accessibility.", with: "")) }) ?? .none
 
         let defaultValue = extractDefaultValue(type: maybeType, from: source)
-        let variable = Variable(name: name, typeName: typeName, accessLevel: (read: accesibility, write: writeAccessibility), isComputed: computed, isStatic: isStatic, defaultValue: defaultValue, attributes: parseDeclarationAttributes(source), annotations: annotations.from(source))
+
+        var definedInTypeName: TypeName? = nil
+        if let containingInTypeName = containedIn?.name {
+            definedInTypeName = TypeName(containingInTypeName)
+        }
```
ilyapuchka: `let definedInTypeName = containedIn.map({ TypeName($0.name) })` ?

SourceryRuntime/Sources/Definable.swift
```diff
@@ -0,0 +1,15 @@
...
+    /// Reference to type name where the object is defined, 
+    /// nil if defined outside of any `enum`, `struct`, `class` etc
+    var definedInTypeName: TypeName? { get }
+
+    /// Reference to actual type where the object is defined, 
+    /// nil if defined outside of any `enum`, `struct`, `class` etc or type is unknown
+    var definedInType: Type? { get }
+
+    /// Reference to actual type name where the method is defined if declaration uses typealias, otherwise just a `definedInTypeName`
+    var actualDefinedInTypeName: TypeName? { get }
```
ilyapuchka: Implementation of this property can be defined in extension, it is all the same everywhere

SourceryRuntime/Sources/Variable.swift
```diff
@@ -51,6 +51,20 @@ public final class Variable: NSObject, SourceryModel, Typed, Annotated {
         return attributes[Attribute.Identifier.final.name] != nil
     }
 
+    /// Reference to type name where the variable is defined,
+    /// nil if defined outside of any `enum`, `struct`, `class` etc
+    public let definedInTypeName: TypeName?
+
+    /// Actual type name where the variable is defined if declaration uses typealias, otherwise just a `definedInTypeName`
+    public var actualDefinedInTypeName: TypeName? {
+        return definedInTypeName?.actualTypeName ?? definedInTypeName
```
ilyapuchka: `actualTypeName` will be not set if type is not resolved in Composer using `resolveType` method (`TypeName` is class but references are not shared), so instead of setting `definedInType` in unique types it should be done in `resolveVariable` and other methods that resolve types to make sure all types are resolved. The same as how method's returnType is resolved explicitly. There also should be tests ensuring that definitions from type aliased extensions are properly resolved.

Sourcery/Parsing/Composer.swift
```diff
@@ -25,6 +25,17 @@ struct Composer {
         let types = parserResult.types
         let typealiases = self.typealiasesByNames(parserResult)
 
+        //set definedInType for all methods and variables
+        types
+            .forEach { type in
+            type.variables.forEach { variable in
+                variable.definedInType = type
```
ilyapuchka: Should we do it for types too? How does it relate to containedTypes/parent property? Do we resolve parent type properly when typealias is used?

Sourcery/Parsing/FileParser.swift
Antondomashnev: To be honest I'm not sure. Maybe specifically in that context in `FileParser` it works, but for example:
```
private func processContainedDeclaration(_ declaration: Any, within containing: (declaration: Any, source: [String: SourceKitRepresentable])) {
        switch containing.declaration {
        case let containingType as Type:
            process(declaration: declaration, containedIn: containingType)
        case let containingMethod as SourceryMethod:
            process(declaration: declaration, containedIn: (containingMethod, containing.source))
        default: break
        }
    }
```
I think doesn't really fit. Thoughts?

SourceryRuntime/Sources/Definable.swift
Antondomashnev: Good catch, thanks. Moved in e737e58

Sourcery/Parsing/FileParser.swift
Antondomashnev: Yep, much better. I've updated in 411fd65

SourceryRuntime/Sources/Definable.swift
```diff
@@ -0,0 +1,21 @@
...
+}
+
+public extension Defined {
+    public var actualDefinedInTypeName: TypeName? {
```
SourceryBot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/krzysztofzablocki/Sourcery/blob/e737e581e85bcf158b05f725b1ac4d2b76932339/SourceryRuntime/Sources/Definable.swift#L18"></span>Undocumented symbol.</td>
    </tr>
  </tbody>
</table>

SourceryRuntime/Sources/Variable.swift
Antondomashnev: I may need some clarification of how the whole logic works. Initially I resolved the `definedInType` in the `resolve...` methods. But it lead to a problem with the following scenario:
```
class Foo {
  func bar() {}
}
extension Foo {
  func baz() {}
}
```
So that code above will end up being one type in `uniqueTypes` array. As Sourcery iterates through the `uniqueTypes` to resolve variable and method types - the definedInType for `baz` was resolved into `Foo` type which is not extension. But it is needed to be extension. That was the main reason to iterate through not unique types and assign correct (from my POV) types to `definedInType` property.

@ilyapuchka does it make sense?

SourceryRuntime/Sources/Definable.swift
Antondomashnev: What does `Definition` protocol describe? I guess it describes that the object is a `definition` of something? I think `Defined` describes different behaviour.

SourceryRuntime/Sources/Definable.swift
Antondomashnev: That is actually fails because of `VariableAutoJSExport` protocol requirements. Could it be a bug? Because if I even skips JSExport for that variable it still fails.

Sourcery/Parsing/Composer.swift
Antondomashnev: For typealias `actualDefinedInTypeName` is not resolved properly. See the comment above for proposal.

SourceryRuntime/Sources/Variable.swift
Antondomashnev: @ilyapuchka I've checked and yeah for typealiases the `actualDefinedInTypeName` is not resolved into correct value. I have two ideas how to solve it:
1. As we require mainly the information whether the type defined in extension or not, we pass additional parameter to `resolveMethodTypes` or `resolveVariableTypes` whether defined in extension. So when the method is resolving the type it will set it as `isExtension`;
2. Second option is to keep the `definedInType` as we have now, but still call `resolveType` for `definedInTypeName` so it will have `actualDefinedInTypeName` resolved properly.

SourceryRuntime/Sources/Definable.swift
ilyapuchka: How does it fail? I think if it works for `Typed` protocol it should work the same way here.

SourceryRuntime/Sources/Definable.swift
ilyapuchka: `Definition` extends `Parsable` and adds path to the file where it is defined. `Parsable` contains definition sourcekit data. Now only `Type` implements `Definition` and the rest is just `Parsable` but I think it make sense to have just one notion of "definition". For types `definedInType` is the same as parent, so no need to resolve it even. My point is that we should keep one notion of "definition", I'm fine with using different name for current `Definition`, we can maybe even get rid of it as we use it only for `Type` anyway. But then I would rename `Defined` to `Definition`.

Sourcery/Parsing/FileParser.swift
ilyapuchka: In your example its fine, just `containedIn: containingIn` is hard to understand. if not `definedIn` then maybe `containingDeclaration` will be more clear?

SourceryRuntime/Sources/Variable.swift
ilyapuchka: The second option is a way to go I think 👍

Antondomashnev: Done in 4303553

Sourcery/Parsing/FileParser.swift
Antondomashnev: I like `definedIn` and I've replaces the confused part in `FileParser` in d7da21b.

Antondomashnev: I like `definedIn` and I've replaced the confused part in `FileParser` in d7da21b.

SourceryRuntime/Sources/Definable.swift
Antondomashnev: Maybe we can really get rid of current `Definition`? I don't see the benefit of having a protocol with completely default implementation that is used for one type.
Check 5240cf6

SourceryRuntime/Sources/Definable.swift
Antondomashnev: > non-'@objc' property 'actualDefinedInTypeName' does not satisfy requirement of '@objc' protocol 'MethodAutoJSExport'
    var actualDefinedInTypeName: TypeName? {
        ^
    @objc 
/Users/antondomashnev/Work/Sourcery/SourceryRuntime/Sources/JSExport.generated.swift:170:9: note: requirement 'actualDefinedInTypeName' declared here
    var actualDefinedInTypeName: TypeName? { get }

There is no extension for `Typed` it maybe the same eventually.

Antondomashnev:
Hey @ilyapuchka any chance you will be able to check my responses and commits soon?

ilyapuchka:
Will check it on a weekend. All sounds good and reasonable so far 👌

Ilya Puchka

Отправлено с iPhone

Sourcery/Parsing/Composer.swift
```diff
@@ -129,6 +143,9 @@ struct Composer {
 
     private func resolveVariableTypes(_ variable: Variable, of type: Type, resolve: TypeResolver) {
         variable.type = resolve(variable.typeName, type)
+        if let definedInTypeName = variable.definedInTypeName {
+            _ = resolve(definedInTypeName, type)
```
ilyapuchka: I think it will be helpful for future to leave the comment explaining why the result is not used here.

Antondomashnev: Agree, added in 0b8947a

PR #383, opened 2017-07-28
Fix line number display of errors in Swift templates

Liquidsoul:
The line number for an unmatched tag was not computed properly, it was always showing 0.
To fix this, we now keep a `processedComponents` array to keep track of the template source that was already computed. This allows to know which line we are at when we encounter an unmatched tag issue.

### DIFF
--- a/Sourcery/Generating/Template/Swift/SwiftTemplate.swift
+++ b/Sourcery/Generating/Template/Swift/SwiftTemplate.swift
@@ -47,11 +47,11 @@ class SwiftTemplate: Template {
 
         let components = templateContent.components(separatedBy: Delimiters.open)
 
-        var sourceFile = [String]()
+        var processedComponents = [String]()
         var commands = [Command]()
 
         let currentLineNumber = {
-            return sourceFile.joined(separator: "").numberOfLineSeparators
+            return processedComponents.joined(separator: "").numberOfLineSeparators + 1
         }
 
         for component in components.suffix(from: 1) {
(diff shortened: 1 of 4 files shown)


Sourcery/Generating/Template/Swift/SwiftTemplate.swift
```diff
@@ -47,11 +47,11 @@ class SwiftTemplate: Template {
 
         let components = templateContent.components(separatedBy: Delimiters.open)
 
-        var sourceFile = [String]()
+        var processedComponents = [String]()
         var commands = [Command]()
 
         let currentLineNumber = {
-            return sourceFile.joined(separator: "").numberOfLineSeparators
+            return processedComponents.joined(separator: "").numberOfLineSeparators + 1
```
Antondomashnev: @Liquidsoul does it make sense to mention why we do `+1` here?

Liquidsoul: You mean add a comment to tell that line numbers start at 1 and not at 0?

Antondomashnev: Yeah, kinda. Don't you think it makes sense?

Liquidsoul: I'll add a small comment here if you think this is needed 😉 
I was more on the side that the code was self explanatory here 😊 
Here was my rationale: we are computing the `currentLineNumber`, which starts from 1, and to do that we count the `numberOfLineSeparators`, which starts from 0. So the block is just the transformation of a count of lines to a line number, and this transformation is just doing the `+1`.

Antondomashnev: Looks perfect 👍 
Thanks a lot @Liquidsoul

PR #385, opened 2017-07-30
Support includes in Swift templates

Liquidsoul:
This PR implements #243

It introduces the custom tag `<%- include("relative_path_to_template") %>` which will add the code from the referenced template (inspired from the EJS implementation).

The features supported by this tag:
- inclusion of templates relative to the including template
- ability to omit the `swifttemplate` extension in the tag
- include cycles detection (templates including each other) to show user friendly error instead of some stack overflow error

I hope you'll like this! 😃 

_Unrelated side note:_
When generating the docs I noticed a lot of changes across the html files. I've only added the changes related to my updates but maybe all these changes should be integrated to the repository.


ilyapuchka:
👍 awesome!

PR #393, opened 2017-08-05
Allowed string filters to be applied to TypeName

### LINKED ISSUE #391: LowerFirst filter fails for associated value type name
Hey, this is an amazing project! I'm just beginning to  explore its possibilites.
Given this enum definition:
```
enum LowerCaseTest {
    case lowerCase(value: UIViewController)
    case helloWorld(value: UICollectionViewController)
}
```

And this template:
```
let {{ "HelloWorld" | lowerFirst }} = "sanity check"
{% for case in type.LowerCaseTest.cases %}
let {{ case.associatedValues.first.typeName | capitalize }}Identifier = "this works"
let {{ case.associatedValues.first.typeName | lowercase }}Identifier = "this works"
let {{ case.associatedValues.first.typeName | lowerFirst }}Identifier = "this fails"
{% endfor %}
```

The generated code is this:
```
let helloWorld = "sanity check"
let UIViewControllerIdentifier = "this works"
let uiviewcontrollerIdentifier = "this works"
let UIViewControllerIdentifier = "this fails"
let UICollectionViewControllerIdentifier = "this works"
let uicollectionviewcontrollerIdentifier = "this works"
let UICollectionViewControllerIdentifier = "this fails"
```

Am I missing something? Reading the associatedValue type's name didn't properly lowercased the first letter.

ilyapuchka:
Resolves #391 and other common complaints that string filters do not work when applied to type name.

Unfortunately StencilSwiftKit filters only support `Strings`, so either we need to reimplement them in Sourcery or suggest to expand those filters on `CustomStringConvertible` types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Added `extensions` filter for stencil template
 - Added include support in Swift templates
 - Swift templates now can throw errors. You can also throw just string literals.
+- Added support for TypeName in string filters (except filters from StencilSwiftKit).
 
 ### Bug fixes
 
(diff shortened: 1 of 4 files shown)


ilyapuchka:
Yes, I agree, the code is a bit more complex then it probably can be. I'm definitely positive about improving it for future releases.

Issue #395, opened 2017-08-13
Better support for path to configuration file


ilyapuchka:
Currently we only support path to location of configuration file meaning it's directory, not file paths. We should accept file paths too so that different configurations files from the same directory can be used.

ilyapuchka:
Resolved with #396

Issue #378, opened 2017-07-25
Improve error reporting related to sourcery invocation


ilyapuchka:
When input parameters or yaml config file are misconfigured we should provide more extensive feedback to avoid issues like #375

PR #396, opened 2017-08-13
added support for file paths in config parameter, improved validation and error reporting

### LINKED ISSUE #394: --config flag doesn't work
As far as I can tell, the `--config` param doesn't work as expected:

With two files with exactly the same content, one named `.sourcery.yml` and the other `application.yml`

Calling `./bin/sourcery`, works as expected and makes use of `.sourcery.yml`, whilst calling `./bin/sourcery --config application.yml` does not, reporting `error: No sources provided.`

Sourcery 0.8.0

ilyapuchka:
Resolves #394, #395, #378


SourceryTests/ConfigurationSpec.swift
```diff
@@ -9,26 +9,136 @@ class ConfigurationSpec: QuickSpec {
...
+                    do {
+                        _ = try Configuration(dict: config, relativePath: relativePath)
+                        return "No error"
+                    } catch {
+                        return "\(error)"
+                    }
+                }
+
+                it("throws error on invalid file format") {
+                    do {
```
ilyapuchka: I've tried, but for some reason it does not even call closure in `execute` and just fails. Why - is beyond me =)

Sourcery/Configuration.swift
```diff
@@ -80,13 +96,20 @@ enum Source {
...
 
-    init(dict: [String: Any], relativePath: Path) {
+    init(dict: [String: Any], relativePath: Path) throws {
         if let projects = dict["project"] as? [[String: Any]] {
-            self = .projects(projects.flatMap({ Project.init(dict: $0, relativePath: relativePath) }))
+            guard !projects.isEmpty else { throw Configuration.Error.invalidSources(message: "No projects provided.") }
+            self = try .projects(projects.map({ try Project(dict: $0, relativePath: relativePath) }))
         } else if let project = dict["project"] as? [String: Any] {
-            self = .projects([Project(dict: project, relativePath: relativePath)].flatMap({ $0 }))
+            self = try .projects([Project(dict: project, relativePath: relativePath)])
```
ilyapuchka: I think just popping error here is fine, though yes, we can catch it and wrap in invalid sources error.

SourceryTests/ConfigurationSpec.swift
Antondomashnev: Interesting, worked for me before, but I think both ways are fine 👍

Issue #377, opened 2017-07-25
Support methods and variables with ` symbol in AutoMockable template


Antondomashnev:
The following code:
```
protocol CNISumUpMigrationIntroModuleInterface: AutoMockable {
    func `continue`()
}
```

Produces:
```
...
class CNISumUpMigrationIntroModuleInterfaceMock: CNISumUpMigrationIntroModuleInterface {

    //MARK: - `continue`

    var `continue`Called = false // ! That doesn't compile

    func `continue`() {
        `continue`Called = true // ! That doesn't compile
    }
}
```

`AutoMockable` should be able to replace ` with empty symbol, so eventually It'll:
```
...
class CNISumUpMigrationIntroModuleInterfaceMock: CNISumUpMigrationIntroModuleInterface {

    //MARK: - `continue`

    var continueCalled = false // ! That doesn't compile

    func `continue`() {
        continueCalled = true // ! That doesn't compile
    }
}
```

PR #397, opened 2017-08-18
Update auto mockable template

Antondomashnev:
It resolves the #377 and additionally methods that throws. So now whether the mock should throw or not can be configured.


Antondomashnev:
This PR is closed in favor of #409

PR #406, opened 2017-08-24
Fix codecov report now showing the correct result

Liquidsoul:
This is a ~_work-in-progress_~ PR to find a solution on the fact that the codecov report only shows a coverage of less that 40% while it should be more around 80%.

This adds `slather` gem and use it in `circle.yml` with the appropriate flags to report coverage of `Sourcery` and `SourceryRuntime` code.

Here is what I tried:
- ~remove `xcpretty` to see if it's related to [this codecov caveat](https://github.com/codecov/example-swift#caveats)~ this does not change the [report](https://codecov.io/gh/krzysztofzablocki/Sourcery/compare/93e5466eeefe6c4c7fb76bd3645fc8204261cdd7...2ee221efab89669d7ce53d2b2ab73421c25ebf0e)
- ~restore `circle.yml` to its state before the codecov reporting stopped~
- ~tweak codecov call~
- use slather ✅

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -3,6 +3,7 @@ machine:
     version: 8.3
   environment:
     XCODE_WORKSPACE: Sourcery.xcworkspace
+    XCODE_PROJECT: Sourcery.xcodeproj
     XCODE_SCHEME: Sourcery
   post:
       - ln -s /Applications/Xcode-8.3.app /Applications/Xcode.app
@@ -18,4 +19,5 @@ test:
       - xcodebuild -workspace "$XCODE_WORKSPACE" -scheme "$XCODE_SCHEME" ONLY_ACTIVE_ARCH=NO CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO test | tee $CIRCLE_ARTIFACTS/xcode_raw.log | xcpretty --color --report junit --output $CIRCLE_TEST_REPORTS/xcode/results.xml
       - swift build
   post:
-      - bash <(curl -s https://codecov.io/bash)
+    - bundle exec slather coverage --scheme "$XCODE_SCHEME" --workspace "$XCODE_WORKSPACE" --output-directory $CIRCLE_ARTIFACTS --binary-basename SourceryRuntime --binary-basename Sourcery --cobertura-xml Sourcery.xcodeproj
+    - bash <(curl -s https://codecov.io/bash) -f $CIRCLE_ARTIFACTS/cobertura.xml -X coveragepy -X gcov -X xcode
--- a/Gemfile
+++ b/Gemfile
@@ -3,8 +3,9 @@ source 'https://rubygems.org'
 gem 'cocoapods'
 gem 'cocoapods-update-if-you-dare'
 gem 'danger'
-gem 'rake'
 gem 'danger-jazzy'
+gem 'rake'
+gem 'slather'
 gem 'xcpretty'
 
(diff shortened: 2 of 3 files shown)


circle.yml
```diff
@@ -18,4 +19,5 @@ test:
       - xcodebuild -workspace "$XCODE_WORKSPACE" -scheme "$XCODE_SCHEME" ONLY_ACTIVE_ARCH=NO CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO test | tee $CIRCLE_ARTIFACTS/xcode_raw.log | xcpretty --color --report junit --output $CIRCLE_TEST_REPORTS/xcode/results.xml
       - swift build
   post:
-      - bash <(curl -s https://codecov.io/bash)
+    - bundle exec slather coverage --scheme "$XCODE_SCHEME" --workspace "$XCODE_WORKSPACE" --output-directory $CIRCLE_ARTIFACTS --binary-basename SourceryRuntime --binary-basename Sourcery --cobertura-xml Sourcery.xcodeproj
+    - bash <(curl -s https://codecov.io/bash) -f $CIRCLE_ARTIFACTS/cobertura.xml -X voeragepy -X gcov -X xcode
```
Antondomashnev: @Liquidsoul could you please explain flag `voeragepy`?

Liquidsoul: I admit that I just copied it from the [codecov section](https://github.com/SlatherOrg/slather/blob/master/README.md#usage-with-codecov) in the slather readme without looking at what it does.

Liquidsoul: And it seems it should be `coveragepy`😅

Liquidsoul: I've squashed the fix in the commit 2704571

Liquidsoul: Nice catch @Antondomashnev BTW 👍

Antondomashnev: Hah, cool 😄

Antondomashnev:
@Liquidsoul what do you think about [this](https://github.com/codecov/example-swift#speed-up-the-build) part of the documentation? Make we can make use of it?

circle.yml
Antondomashnev: Honestly I'm trying to get the idea why we need additional flags at all here, I've tried to read [documentation](https://circleci.com/docs/1.0/code-coverage/#codecov) and it seems we don't need any parameters by default. is it because of Slather?

Liquidsoul: Yup, this is to make this work as we expect.
With `slather`, we just want codecov to read the cobertura file and use only its content. That's why the `-X` options are there. It's to prevent `codecov` to try to gather coverage information on its own which he fails to do correctly as we can see in the[ current `master`](https://codecov.io/gh/krzysztofzablocki/Sourcery).

Liquidsoul:
@Antondomashnev I've use this part of the codecov documentation (the `-J` option) in the `tweak codecov call` step without success 😞

circle.yml
Liquidsoul: If you want to see the available options in the script, look to its source [here](https://codecov.io/bash) (if you're not afraid of `bash` 😉)

Antondomashnev: Aha, thanks, now it's all clear 😄

Antondomashnev:
I think I've got the reason why the codecov stopped working. One day we added a new project `TemplateTests` and in the `Sourcery` scheme we included the step to run also tests for that `TemplateTests` project, so it seems that the last coverage file for the Sourcery workspace was not the `Sourcery` target, but `TemplateTests` target. Anyway current solution seems fine to me, so I'm happy to merge it 😄

Issue #401, opened 2017-08-23
Add isDeinit in Method


BenchR267:
There are a lot of getters to determine if a `Method` is an initializer, but there is no one for detecting if it's a `deinit`. Would be great to have it in addition.

ilyapuchka:
Resolved with #408

PR #408, opened 2017-08-24
Add isDeinitializer property for Method;

Antondomashnev:
This PR resolves request in #401.


SourceryRuntime/Sources/Method.swift
```diff
@@ -157,6 +157,11 @@ public final class MethodParameter: NSObject, SourceryModel, Typed, Annotated {
         return selectorName.hasPrefix("init(")
     }
 
+    /// Whether method is an deinitializer
+    public var isDeinitializer: Bool {
+        return selectorName.hasPrefix("deinit")
```
Liquidsoul: Why not testing for equality here?
Maybe someone could create a `deinitMyObjects()` and I suppose this would report it as a _deinitializer_.

Antondomashnev: That's a good catch, I've fixed it and added a test in 6e0df7e.

SourceryTests/Models/MethodSpec.swift
```diff
@@ -58,6 +58,12 @@ class MethodSpec: QuickSpec {
                 expect(Method(name: "foo()", definedInTypeName: TypeName("BarAlias", actualTypeName: TypeName("Bar"))).actualDefinedInTypeName).to(equal(TypeName("Bar")))
             }
 
+            it("reports isDeinitializer properly") {
+                expect(sut?.isDeinitializer).to(beFalse())
+                expect(Method(name: "deinitObects() {}").isDeinitializer).to(beFalse())
```
Liquidsoul: Add the missing `j` here for `deinitObjects` 😉

Antondomashnev: Ah, silly me, this is fixed in 24f91aa 😄

Issue #407, opened 2017-08-24
AutoMockable produces strange method names


garnett:
0.8.0, latest stencil template:

```
class ConvertBalanceCalculatorViewMock: ConvertBalanceCalculatorView {

    //MARK: - showSourceCurrency

    var showSourceCurrency___Called = false
    var showSourceCurrency___ReceivedCurrency: CurrencyCode?

    func showSourceCurrency(_ currency: CurrencyCode) {
        showSourceCurrency___Called = true
        showSourceCurrency___ReceivedCurrency = currency
    }
```
I don't quite undertand `stencil` syntax, just wanted to confifrm if having `_ _ _` in method names is expected behaviour

ilyapuchka:
I think with #368 this template now replaces `,:()` characters with `_`, though we should probably use empty string instead, @Antondomashnev ?

Liquidsoul:
Looking at the code, I can confirm that with #368, characters `:` and `(` are replaced by `_` and `)` is replaced by empty string.
Maybe we can just use empty string for all the characters. At the same time, maybe we could factorise the stencil's code using a `macro` instead of repeating the replacement code.

Antondomashnev:
@Liquidsoul @ilyapuchka you're right guys, but I've tried the empty string and for me it looked worse. Maybe if we can combine the empty string with capitalized first letter of first method parameter - i'll look better. 
I can take a look on it, how we can make it better.

@Liquidsoul regarding the macro - there is one clear downside for me that prevented me to use them - macro always requires a new line. If you know how to change that behaviour I'm more than happy to replace repeatable code.

ilyapuchka:
@Antondomashnev combining capitalization and empty string seems like a way to go for me.
P.S. swift templates do not have "new line problem" 😉

Liquidsoul:
@Antondomashnev As long as you do not include the newline in the macro you can do that.
I've just tried it on the AutoMockable template and it works pretty well.
Declare the macro like this:
```swift
{% macro fixName name %}{{ name | replace:"(","_" | replace:")","" | replace:":","_" }}{% endmacro %}
```

And replace all:
```swift
{{ method.selectorname | replace:"(","_" | replace:")","" | replace:":","_" }}
```
with:
```swift
{% call fixName method.selectorName %}
```

And you'll have your _replace_ routine in only one place 😃

Antondomashnev:
Amazing, thanks @Liquidsoul

PR #409, opened 2017-08-25
Automockable name improvement

Antondomashnev:
This resolves the discussion around #407 
For
```
protocol CurrencyPresenter: AutoMockable {
    func showSourceCurrency(_ currency: String)
}
```

The generated code now:
```
class CurrencyPresenterMock: CurrencyPresenter {

    //MARK: - showSourceCurrency

    var showSourceCurrencyCalled = false
    var showSourceCurrencyReceivedCurrency: String?

    func showSourceCurrency(_ currency: String) {
        showSourceCurrencyCalled = true
        showSourceCurrencyReceivedCurrency = currency
    }

}
```

Thanks @Liquidsoul for pointing proper macro usage to me 👍


SourceryTests/Generating/StencilTemplateSpec.swift
```diff
@@ -90,7 +90,7 @@ class StencilTemplateSpec: QuickSpec {
                     expect(generate("{{ type.MyClass.variables.1.typeName | capitalise }}")).to(equal("Myclass"))
                 }
 
-                it("checks for string in name") {
+                fit("checks for string in name") {
```
Liquidsoul: I think this is a typo 😄

Antondomashnev: Yeah, @SourceryBot caught it 😄 
Removed in e56fc88

ilyapuchka:
@Antondomashnev maybe will make sense to merge this pr with #397?

Antondomashnev:
@ilyapuchka it does make sense 😄 Will close #397 and update that one.

Templates/Tests/Generated/AutoMockable.generated.swift
```diff
@@ -15,26 +15,44 @@ import AppKit
 
 
 
+
```
Liquidsoul: Maybe we should try to avoid adding some empty lines here.
This comes from the empty lines between the `macro` declarations, maybe we could just remove them from the template?

Issue #400, opened 2017-08-23
0.8.0 Zip release shows 0.7.2 as its version


Liquidsoul:
Hi there,

I downloaded the 0.8.0 zip version of Sourcery and when running `bin/sourcery --version` from the unzipped folder, it prints `0.7.2`.

Is this still the new code showing the incorrect version? Or just the previous version?

BenchR267:
The Info.plist also shows 0.7.2. I think it wasn't built again before releasing.

Liquidsoul:
I've updated the binary in the 0.8.0 release with a build from the 0.8.0 tag.
The release still contains the zip initially uploaded when the release was created (`Sourcery-0.8.0-initial.zip`).
I'll let the issue open a while for other collaborators to stop by before closing the issue (and maybe remove the initial zip).

Antondomashnev:
Thanks @Liquidsoul for fixing my mistake 😄 I've checked `info.plist` of new file and it contains correct version. I'm still thinking how the issue appeared as we have all process automated and it worked fine...

PR #410, opened 2017-08-25
Remove clutter from releases

Liquidsoul:
In zip versions, I noticed that some unrelated `Release` and `Sourcery.build` folders were present.
This is because of the `:validate_docs` step which was writing them in the `build/` during the process while the `:clean` step trashing `build/` had already been done.
To prevent this, we now specify the derivedDataPath like in the `:build` step and remove it after the process is done.

I noticed that after fixing #400.
This will reduce the zip file by several MB (almost 4MB for the `0.8.0` release)!
You can see this because I've updated the 0.8.0 with an uncluttered zip and kept the initial invalid release containing a 0.7.2 build.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -110,7 +110,8 @@ end
 desc "Validate docs"
 task :validate_docs do
   print_info "Checking docs are up to date"
-  sh "sourcekitten doc --module-name SourceryRuntime > docs.json && bundle exec jazzy --skip-undocumented --no-download-badge && rm docs.json"
+  sh "sourcekitten doc --module-name SourceryRuntime -- -workspace Sourcery.xcworkspace -scheme Sourcery-Release -derivedDataPath #{BUILD_DIR}tmp/ > docs.json && bundle exec jazzy --skip-undocumented --no-download-badge && rm docs.json"
+  sh "rm -fr #{BUILD_DIR}tmp/"
 end
 
 ## [ Release ] ##########################################################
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -110,7 +110,8 @@ end
 desc "Validate docs"
 task :validate_docs do
   print_info "Checking docs are up to date"
-  sh "sourcekitten doc --module-name SourceryRuntime > docs.json && bundle exec jazzy --skip-undocumented --no-download-badge && rm docs.json"
+  sh "sourcekitten doc --module-name SourceryRuntime -- -workspace Sourcery.xcworkspace -scheme Sourcery-Release -derivedDataPath #{BUILD_DIR}tmp/ > docs.json && bundle exec jazzy --skip-undocumented --no-download-badge && rm docs.json"
+  sh "rm -fr #{BUILD_DIR}tmp/"
```
Antondomashnev: Does it make sense to create a rake task `clean_temp` and run it after `build` and `validate_docs` tasks?

Something like:
```
task : validate_docs => [:clean_temp] do
...
end
```

Liquidsoul: I am not a `Rakefile` specialist but I do not think that would work.
What you describe here is a prerequisite (meaning it will run _before_ the task), where we want a "post" step.

I find it easier to understand to keep related commands close to each other.
In our case here, we put files in the `tmp/` folder and remove them right after. It makes the task _self contained_.

Antondomashnev: > In our case here, we put files in the tmp/ folder and remove them right after

I think it's a good point, maybe then I would only declare a variable `#{BUILD_DIR}tmp/`. Other than that - good job 👍

Liquidsoul: I've added the variable in 5cfd48b 😉

PR #413, opened 2017-09-07
Implement forloop.counter0

garrett-davidson:
Currently there is no way to directly get a zero-indexed iteration count in a for loop. Stencil solves this using [`forloop.counter0`](http://stencil.fuller.li/en/latest/builtins.html?highlight=counter0). This patch adds that functionality.


krzysztofzablocki:
Hey, thanks for contributing but you need to update stencil / StencilSwiftKit versions and not just change code in the pod that we use as this will break with any update

garrett-davidson:
Oops! Sorry I did this quickly and wasn't paying enough attention to notice it was in the pod. Closing.

Issue #430, opened 2017-10-31
NSRegularExpression bug in parseAnnotations


artyom-razinov:
Hi.

There is an issue in `parseAnnotations(...)`. If I write too much code between `// sourcery:file:` and `// sourcery:end:` it skippes those annotations and doesn't put code in right files.

I localized the bug. The attachement contains scripts to reproduce this bug in Sourcery, also there is playground that shows that NSRegularExpression fails to find ranges.

[bug.zip](https://github.com/krzysztofzablocki/Sourcery/files/1431616/bug.zip)

If I pass too long string to NSRegularExpression it fails to work. E.g.: length 1, 2, 3 work, length 4, 5, 6 doesn't (the actual boundary is around 150K characters).

I will try to fix this issue myself, but I don't guarantee the result.
UPD: PR will be soon.

By the way, Sourcery is adorable, thank you!

PR #431, opened 2017-10-31
Simplify regex for annotation (#430)

artyom-razinov:
This is a workaround for bug in NSRegularExpression that occurs for long strings and regex in regex(annotation:) method, see issue #430.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,13 @@
 
 ### New Features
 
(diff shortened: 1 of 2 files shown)


ilyapuchka:
Hi, @artyom-razinov. Thanks for the fix! Just add an entry to changelog and we merge it.

artyom-razinov:
Done

ilyapuchka:
👍

PR #437, opened 2017-11-10
Add the ability to force the parsing of previousily generated files

lucabartoletti:
Not sure if this is a common use cases.

I was trying to create a template with the goal to create a struct that implementes the same properties of a class that conforms to the protocol `AutoStructType`

```
{% for type in types.implementing.AutoStructType %}
// sourcery:file:{{type.name}}.type.swift
{% include "AutoImport.stencil" %}
// MARK: Struct for {{ type.name }}
public struct {{type.name}}Type {
{% for variable in type.variables %}public let {{variable.name}}: {{variable.typeName}}
{% endfor %}
public init(from {{type.name | lowerFirst}}: {{type.name}}) {
{% for variable in type.variables %}{{variable.name}} = {{type.name | lowerFirst}}.{{variable.name}}
{% endfor %}
}
}
extension {{type.name}}Type : AutoEquatable {}
extension {{type.name}}Type : AutoHashable {}
// sourcery:end
{% endfor %}
```

The problem is that I would also like than to make the new type adopt other templates. In the example above  AutoEquatable, AutoHashable.

To do this I added two separate build steps that

1. Run all the types related templates (The template above)
```
$PODS_ROOT/Sourcery/bin/sourcery --sources ./Improve --templates ./Improve/TemplateTypes --output ./Improve/AutoGeneratedTypes/ 
```
2. Run all the extensions related templates (`AutoEquatable ` and `AutoHashable `)
```
$PODS_ROOT/Sourcery/bin/sourcery --sources ./Improve --templates ./Improve/TemplateExtensions --output ./Improve/AutoGeneratedExtensions/
```

This solution doesn't work with the current Sourcery implementation cause it automatically skips all the sorcery generated files.

With this change I added a parameter that allows to specify extensions that should be parsed even if they are generated so that the two steps build can work.

```
$PODS_ROOT/Sourcery/bin/sourcery --sources ./Improve  --parsableExtensions type --templates ./Improve/TemplateExtensions --output ./Improve/AutoGeneratedExtensions/
```

In this case all the generated files that contains ```.type.``` is parsed in the second Sourcery run

Please let me if there are other way to achieve what I'm trying to do or if you think that Sourcery should not support this use case. If instead you think this can be useful I'll add tests


Sourcery/Parsing/Utils/Verifier.swift
```diff
@@ -16,10 +16,16 @@ enum Verifier {
...
+                                path: Path,
+                                parsableExtensions: [String]) -> Result {
         guard !content.isEmpty else { return .approved }
 
-        if content.hasPrefix(Sourcery.generationMarker) {
+        let hasParsableExtension = (parsableExtensions.filter({ (ext) -> Bool in
+            return path.hasExtension(as: ext)
+        }).isEmpty == false)
+
+        if content.hasPrefix(Sourcery.generationMarker) && hasParsableExtension == false {
```
ilyapuchka: Why not just use a boolean flag to signal that generated files should be parsed?

lucabartoletti: Cause not all generated files needs to be parsed, I want to selected only some files to be parsed in the second pass

lucabartoletti:
Thanks for the feedback, I'll try it but I'm trying to keep my code separated from the generated one, so I'm avoiding to put inline generation inside the project files

Sourcery/Parsing/Utils/Verifier.swift
krzysztofzablocki: I think if you just said 'parse generated files' it might lead to chicken and egg scenario, I prefer this approach as its more explicit for the end-user

krzysztofzablocki:
@ilyapuchka the code thats injected via inline code generation is never scanned again, so not sure if this would change anything for @lucabartoletti, the reason we don't scan generated code is to avoid chicken-egg scenarios

I think the way @lucabartoletti deals with this by using 2nd pass and explicit filter for generated files is something we can accept, it won't affect normal users and it makes sense

ilyapuchka:
@krzysztofzablocki you are right! I would maybe change naming of the command line argument to something like `force-parse` or something, but that's nitpicking =)

krzysztofzablocki:
@ilyapuchka you happy with this pr? if so its good to merge before new release

ilyapuchka:
Pretty much, thanks @lucabartoletti! Can you also add this option description to the docs [here](https://github.com/krzysztofzablocki/Sourcery#command-line-options) and [here](https://github.com/krzysztofzablocki/Sourcery/blob/master/guides/Usage.md)?

lucabartoletti:
Thanks,

@ilyapuchka Is it ok if I do that on a separate pull request? So that I can merge the docs change only once the new release is out

ilyapuchka:
It's better to have docs in sync with release. No rush. If you busy I can update them myself later.

lucabartoletti:
No, I can do it. Just that I was not sure what is the best practice in this case. I can do the change later today.

Cheers

lucabartoletti:
@ilyapuchka I updated the README.md

lucabartoletti:
@ilyapuchka If you have feedbacks about the docs let me know. I'm happy to open another pull request

migonin:
@lucabartoletti @ilyapuchka hi! Is there any way to get inline code parsed? I need to parse code generated by tamplate A by another template B.

migonin:
CC @krzysztofzablocki

Issue #412, opened 2017-09-07
Sourcery ignores command line arguments when .sourcery.yml exists in current directory


garrett-davidson:
I just started trying out Sourcery (prebuilt binary of 0.8.0) and ran into a very confusing issue. I created a test .sourcery.yml file in the root directory one of my projects, but it didn't do quite what I wanted (my fault). So I tried playing around with the command line arguments while cd'ed into that root directory, but couldn't figure out why nothing I tried was having any effect. Eventually I realized that the configuration file in the current directory will silently take priority over command line arguments, even without passing --config.

ilyapuchka:
This is expected behaviour. What we can surely improve is to add logs telling that yaml file is used and switch to using comand line arguments if they are provided
@Antondomashnev wdyt?

Antondomashnev:
@ilyapuchka I'd personally see two options:
1. Override configuration value if there is a value directly provided in command
2. Throw exception if there is a conflict between configuration and command line argument

I'm not sure about the first option, although this is the behavior I'd want to have for me personally as a developer.
The second option may sound like an overkill for the design though.

ilyapuchka:
I think it will make sense to override options of config file is specified explicitly, not when it is discovered automatically, what do you think? but to be honest I wouldn't override at all to make it simpler.

Antondomashnev:
> specified explicitly, not when it is discovered automatically

What do you mean by discovered automatically?

Have you seen by any chance this overriding behavior somewhere?

ilyapuchka:
By automatically I mean that we have a default value for config path parameter which is current path and we have a default name for the config file to search for, so thus the behaviour described by @garrett-davidson.
Regarding overriding behaviour I don't think I saw it anywhere. I think it will be more straight forward if we use one or another way and just print out a hint to the user about it.

Antondomashnev:
> I think it will be more straight forward if we use one or another way and just print out a hint to the user about it.

Yes, if Sourcery finds command line parameter and configuration file warn the user about possible collisions.

> By automatically I mean that we have a default value for config path parameter which is current path and we have a default name for the config file to search for

Absolutely agree 👍

PR #441, opened 2017-11-15
Add a warning when the user is passing arguments but a yml config is used

lucabartoletti:
If the user is specifing any parameter aside than config and a yml is used it shows a WARNING to avoid confusion

Related to discussion in #412

### DIFF
--- a/Sourcery/main.swift
+++ b/Sourcery/main.swift
@@ -149,16 +149,13 @@ func runCLI() {
                     let relativePath: Path = configPath.isDirectory ? configPath : configPath.parent()
                     configuration = try Configuration(path: yamlPath, relativePath: relativePath)
 
-                    // Doesn't consider 'configPath' cause it's related to the yaml file
-                    let hasAnyParameter = (watcherEnabled == true ||
-                        disableCache == true ||
-                        verboseLogging != true ||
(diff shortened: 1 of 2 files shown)


Sourcery/main.swift
```diff
@@ -146,9 +146,25 @@ func runCLI() {
                 do {
                     let relativePath: Path = configPath.isDirectory ? configPath : configPath.parent()
                     configuration = try Configuration(path: yamlPath, relativePath: relativePath)
-                    Log.info("Using configuration file at '\(yamlPath)'")
+
+                    // Doesn't consider 'configPath' cause it's related to the yaml file
+                    let hasAnyParameter = (watcherEnabled == true ||
```
ilyapuchka: Not sure I understand. In yaml file you can only specify sources/templates/outputs and extra runtime arguments, for the rest, like quite, prune and others you need to use command line arguments. I'd like to keep it like that and not clutter yaml files with booleans. So I guess here we only need to check if any sources/templates/outputs were provided with command line arguments.

lucabartoletti: That make sense, I didn't though about those booleans, I agree that it doesn't make sense to have them in the yaml.

I'll fix it. Thanks!

PR #445, opened 2017-11-20
Disable code coverage generation for release schema;

Antondomashnev:
`Sourcery-Release` schema from now one doesn't generate code coverage. I've tested the changes on the temp project and it fixed the problem. As far as I can see it won't affect anything since for test script we run `Sourcery` scheme, not `Sourcery-Release`.

P.S as we have now 5 core team members, do you think it makes sense to define a review policy? Should we ask all team members for review and wait for the first one accepted?


krzysztofzablocki:
Totally, one core member is fine :)

Antondomashnev:
@Liquidsoul @krzysztofzablocki sounds like a plan 👍

PR #450, opened 2017-12-05
Improvements in AutoMockable example template

lukkas:
Three improvements were introduced:
1. If generated "received" variable is a closure, question mark that is applied at the end is treated as a closure's return parameter optionality marker, not as entire closure's optionality marker. It has been fixed by adding parentheses.
Example:
```
func setCompletion(_ completion @escaping () -> ()) {
...
}
```
Fragment of generated code, that is relevant for us, will look like this:
```
var setCompletionReceivedCompletion: () -> ()?
```
After fix:
```
var setCompletionReceivedCompletion: (() -> ())?
```
2. Instead of boolean marking if given mocked method was called, integer counting each call was introduced.
{methodName}Called boolean still is present, but now is a computed variable utilizing new calls count integer.
3. `upperFirstLetter` was replaced with `capitalize`. I belive this must have been a recent change to Stencil. Correct me if I'm wrong since I had not been using Sourcery for a while. `upperFirstLetter` just didn't work for me.

### DIFF
--- a/Templates/Templates/AutoMockable.stencil
+++ b/Templates/Templates/AutoMockable.stencil
@@ -33,7 +33,7 @@ import AppKit
 
 {% macro extensionMethodReceivedParameters method %}
     {%if method.parameters.count == 1 %}
-        {% call swiftifyExtensionMethodName method.selectorName %}Received{% for param in method.parameters %}{{ param.name|upperFirstLetter }} = {{ param.name }}{% endfor %}
+        {% call swiftifyExtensionMethodName method.selectorName %}Received{% for param in method.parameters %}{{ param.name|capitalize }} = {{ param.name }}{% endfor %}
     {% else %}
     {% if not method.parameters.count == 0 %}
         {% call swiftifyExtensionMethodName method.selectorName %}ReceivedArguments = ({% for param in method.parameters %}{{ param.name }}: {{ param.name }}{% if not forloop.last%}, {% endif %}{% endfor %})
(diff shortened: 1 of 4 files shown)


ilyapuchka:
@lukkas thanks for your contribution! As far as I can tell there were no changes to this filter, can you you explain what exactly does not work of even submit a failing test?
Other changes look awesome to me.

Antondomashnev:
@lukkas hi =)
What exactly was the error regarding the `upperFirstLetter ` filter? It seemed to be working and we're using it from [StecilSwiftKit](https://github.com/SwiftGen/StencilSwiftKit/blob/master/Sources/Filters%2BStrings%2BLettercase.swift#L51). 
The `capitalized` filter is not the same as `upperFirstLetter`, because the first will transform `i love sourcery` to `I Love Sourcery` and the latter to `I love sourcery`.

lukkas:
Hmm. When I try to use upperFirstLetter instead I get message `Unknown filter 'upperFirstLetter'`. I use Sourcery via Cocoapods. I assumed that `upperFirstLetter` was some kind of deprecated form of `capitalized`, therefore I wasn't researching it further, but I can see now that they are two different things. I hope I'll find some time today's evening and try find to find out why it happens for me.

lukkas:
Issue disappeared for me after updating Sourcery to 0.10.0. I brought back `upperFirstLetter`.

krzysztofzablocki:
@Antondomashnev back to you

Antondomashnev:
@lukkas amazing! To be sure we'll never break new template syntax you've introduced, can you please update the template test for `AutoMockable`?

lukkas:
@Antondomashnev Sure. Done :)

PR #451, opened 2017-12-07
Add sort filter

Lutzifer:
for my templates, I ran into issues, where the order of the variables was not always the same for different calls.

This filter allows to order them:

example usage:
```
    {% for variable in protocol.allVariables|sorted:"name" %}
      {{ variable.name }}: {{ variable.typeName }}{% if not forloop.last %},{% endif %}
    {% endfor %}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### New Features
 
+- Added new sorted Stencil filter. This can sort arrays by calling e.g. `protocol.allVariables|sorted:"name"`
 - Added new toArray Stencil filter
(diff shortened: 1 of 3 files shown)


krzysztofzablocki:
Thanks for contributing!

ilyapuchka:
@krzysztofzablocki @Lutzifer maybe having two different filters for ascending and descending sorting would be more flexible?

krzysztofzablocki:
@ilyapuchka good idea

ilyapuchka:
@Lutzifer would you like to submit a pr for that too?

Lutzifer:
let's just add sortedDescending, as I would assume that ascending is the default

PR #459, opened 2017-12-14
Fix generating non-optional variable mocks in AutoMockable

lukkas:
Right now if we have a protocol like this one
```
protocol OurProtocol: AutoMockable {
      var nonOptionalVariable: VariableType { get }
}
```
and we generate mock for it we get something like this:
```
class OurProtocolMock: OurProtocol {
      var nonOptionalVariable: VariableType!
}
```
It won't compile since protocol requires non optional variable, which it isn't.
This is the effect after introduced changes:
```
class OurProtocolMock: OurProtocol {
      var nonOptionalVariable: VariableType {
            get { return underlyingnonOptionalVariable }
            set(value) { underlyingnonOptionalVariable = value }
      }
      var underlyingnonOptionalVariable: VariableType!
}
```


Antondomashnev:
Hi @lukkas thanks for your contribution. Could you please update related [Template Test](https://github.com/krzysztofzablocki/Sourcery/tree/master/Templates/Tests)

lukkas:
Hi @Antondomashnev. Sure. I've updated tests and also improved underlying variables naming, so it's `underlyingAge` instead of `underlyingage`.

ilyapuchka:
@lukkas can you please add CHANGELOG entry? We can merge it then.

lukkas:
@ilyapuchka It is done :)

PR #468, opened 2017-12-28
Return empty array instead of optional array

ZevEisenberg:
Related to a584ce88421a47ea096354fa3143b4c186c52519. Makes it easier to loop over types in Swift-based templates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@
 ### New Features
 
 - Added support for subscripts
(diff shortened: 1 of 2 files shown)


ZevEisenberg:
Note that this would be a source-breaking change for anyone who has a Swift template that does something like this:

```swift
<% for type in (types.implementing["SomeProtocol"] ?? []) { %>
```

The new correct syntax would be:

```swift
<% for type in types.implementing["SomeProtocol"] { %>
```

PR #469, opened 2017-12-30
Add swiftlint autocorrect call before linting

Liquidsoul:
This will prevent some unwanted warnings which are easily fixed automatically by swiftlint (e.g. `trailing_whitespaces`).


ilyapuchka:
Regarding trailing spaces on unneeded lines there is a setting in Xcode, just found it few days ago =)

PR #472, opened 2018-01-03
Watch mode improvements

### LINKED ISSUE #465: Watcher mode terminating with error
Hey there!

I was using Sourcery `0.10.1` and when trying to use the `--watch` option It terminated with an error.

* I'm using a simple _.sourcery.yml_
* 2 simple `.stencil` files, one is empty
* The base command runs fine and generates the files well.

Watcher mode output
``` 
Using configuration file at '.sourcery.yml'. WARNING: Ignoring the parameters passed in the command line.
Scanning sources...
Found "extension <-- deleted this bit of the output -->
Sourcery[84297] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-12-21 22:36 Sourcery[84297] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-12-21 22:36 Sourcery[84297] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-12-21 22:36 Sourcery[84297] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-12-21 22:36 Sourcery[84297] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
...
error: Starting watching templates.
2017-12-21 22:42 Sourcery[84404] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
2017-12-21 22:42 Sourcery[84404] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-21)
```

If any of these is worth: 
normal output
```
Using configuration file at '.sourcery.yml'. WARNING: Ignoring the parameters passed in the command line.
Scanning sources...
Found 528 types.
Loading templates...
Loaded 2 templates.
Generating code...
Finished.
Processing time 0.712023019790649 seconds
```

_.sourcery.yml_ file
```
sources:
  - App/
templates:
  - App/Sourcery/Templates/
output:
  App/Sourcery/AutoGenerated/
prune: true
```

Btw, thanks for Sourcery 👍

krzysztofzablocki:
- Fixes FSEvents errors reported in #465 that happen on Sierra by avoiding observing thousands of files, instead focusing on the top folders.
- Supports adding new templates files while in watcher mode
- Supports adding new source files while in watcher mode
 
closes #465


Sourcery/Sourcery.swift
```diff
@@ -96,7 +96,7 @@ class Sourcery {
 
         Log.info("Starting watching sources.")
```
ilyapuchka: Maybe we should add here list of directories that are being observed?

krzysztofzablocki: why ? this doesn't change the behaviour of the watcher, all the files you were observing previously would still be observed, this only changes the implementation detail, since any file changes will cause regeneration like they did before

ilyapuchka: Just for better visibility what is happening. Also when new folder is added and it is started being watched we may log this event too, for the same reason. Maybe in a more verbose mode of watcher.
To make sure I don't misunderstood the change. Watcher now will watch only folders and not folders and individual files in it? But events will still come in when files change, right?

krzysztofzablocki: Some time ago we broke the following behaviour, that I wanted to work (and it did work in the past):

- When a new code or template is added, it should be immediately supported in watcher mode so you don't need to restart Sourcery session

Watcher will observe all top level directories, observing directory still generates file events. It shouldn't change anything for the end-user except again supporting the above scenario

ilyapuchka: Ok, got it =) Would still add folder names to the log, but that's up to you =)

PR #480, opened 2018-01-13
Fix func signature for void return throws

kdawgwilk:
AutoMockable wasn't generating valid code for requirements that were void return throwing.  This PR adds an else for void returning methods and checks if throwing and adds throws if needed.  I have testing this in my project is now generating code that compiles

### DIFF
--- a/Templates/Templates/AutoMockable.stencil
+++ b/Templates/Templates/AutoMockable.stencil
@@ -77,7 +77,7 @@ import AppKit
         {% call methodReceivedParameters method %}
     }
 {% else %}
-    func {{ method.name }}{% if not method.returnTypeName.isVoid %} {% if method.throws %}throws {% endif %}-> {{ method.returnTypeName }}{% endif %} {
+    func {{ method.name }}{% if not method.returnTypeName.isVoid %} {% if method.throws %}throws {% endif %}-> {{ method.returnTypeName }}{% else %}{% if method.throws %} throws{% endif %}{% endif %} {
         {% if method.throws %}
         {% call methodThrowableErrorUsage method %}
         {% endif %}


Templates/Templates/AutoMockable.stencil
```diff
@@ -77,7 +77,7 @@ import AppKit
         {% call methodReceivedParameters method %}
     }
 {% else %}
-    func {{ method.name }}{% if not method.returnTypeName.isVoid %} {% if method.throws %}throws {% endif %}-> {{ method.returnTypeName }}{% endif %} {
```
ilyapuchka: I think we can actually rewrite it like this:

```
func {{ method.name }}{% if method.throws %} throws {% endif %}{% if not method.returnTypeName.isVoid %} -> {{ method.returnTypeName }}{% endif %} {
```

ilyapuchka:
@kdawgwilk thanks for the fix, I came up with a simpler suggestion though, what do you think?
Also do not forget to update CHANGELOG

@krzysztofzablocki @Antondomashnev @Liquidsoul is it possible with dagger to check for changes in non-swift files? I think we should extend this to templates.

krzysztofzablocki:
@ilyapuchka sure, its just doing grep on Sourcery folder, you can add the templates folder for changes check

Liquidsoul:
@ilyapuchka @krzysztofzablocki I've added the `Templates` folder for changelog warnings in Danger here #484

krzysztofzablocki:
@kdawgwilk could you apply @ilyapuchka comment to template code before merging?

kdawgwilk:
Sorry it took me so long to get back to this but thanks @ilyapuchka for the other PR fixing this issue 👍

PR #488, opened 2018-01-14
Fixed parsing method return type

### LINKED ISSUE #489: Protocol methods return types are not extracted correctly when broken in several lines
When protocol defines a method, usually generic, and generic constraint takes few lines, we only extract first line into return type:

```swift
protocol Foo {
  func foo<T>(t: T) -> T where
    T: Equatable
}
```

will parse return type as "T where".

This is due to the fact that protocol methods do not have bodies, and their key in this case only contains `func foo<T>(t: T) -> T`. Then we only use lines containing this string, which will only include first line in this case.

To fix this we need to manually parse method suffix until next declaration starts or protocol declaration ends. For that we can pass in next substructure to get its offset and use it as upper bound.

ilyapuchka:
Void return type was not correctly parsed in methods without spaces between `)` and `{`, like `func foo(){ }`
Also resolves #489

### DIFF
--- a/SourceryTests/Parsing/FileParser + MethodsSpec.swift
+++ b/SourceryTests/Parsing/FileParser + MethodsSpec.swift
@@ -56,7 +56,7 @@ class FileParserMethodsSpec: QuickSpec {
                 }
 
                 it("extracts class method properly") {
-                    expect(parse("class Foo { class func foo() }")).to(equal([
+                    expect(parse("class Foo { class func foo() {} }")).to(equal([
                         Class(name: "Foo", methods: [
                             Method(name: "foo()", parameters: [], isClass: true, definedInTypeName: TypeName("Foo"))
                             ])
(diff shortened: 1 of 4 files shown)


SourceryTests/Parsing/FileParser + MethodsSpec.swift
```diff
@@ -56,7 +56,7 @@ class FileParserMethodsSpec: QuickSpec {
                 }
 
                 it("extracts class method properly") {
-                    expect(parse("class Foo { class func foo() }")).to(equal([
+                    expect(parse("class Foo { class func foo() {} }")).to(equal([
```
krzysztofzablocki: what if its a protocol definition and it doesn't have the `{}` ?

ilyapuchka: it is covered with other tests

PR #490, opened 2018-01-14
Fix generic type tests

hartbit:
Re-generated boilerplate, which added missing `GenericType` and `GenericTypeParameter` equality functions, which uncovered failing tests, which I fixed.


ilyapuchka:
Thanks @hartbit !

hartbit:
@ilyapuchka thanks for the quick merge!

PR #494, opened 2018-01-17
update readme installation with homebrew information

JakubMazur:


krzysztofzablocki:
thanks for contributing :)

PR #497, opened 2018-01-18
Various fixes/changes for methods parsing

ilyapuchka:
- fixed parsing `throws`/`rethrows` for initialisers
- now `selectorName` for methods without arguments will not contain `()` at the end. This will let use such methods by reference, currently it results in a method call. This is a breaking change
- initialisers now have `returnType` and `returnTypeName` set to the type they produce (type where they are defined). This is more aligned with other methods than using empty return type. This is a breaking change.


Sourcery/Parsing/FileParser.swift
```diff
@@ -289,6 +289,11 @@ final class FileParser {
                 for (index, parameter) in method.parameters.enumerated() where index < argumentLabels.count {
                     parameter.argumentLabel = argumentLabels[index] != "_" ? argumentLabels[index] : nil
                 }
+
+                // adjust method selector name as methods without parameters do not have ()
+                if method.parameters.isEmpty {
+                    method.selectorName.trimSuffix("()")
```
Antondomashnev: Does it make sense to have this suffix as a constant somewhere? As far as I remember we use it in some other places as well.

SourceryRuntime/Sources/Description.generated.swift
```diff
@@ -122,26 +122,17 @@ extension Method {
     override public var description: String {
         var string = "\(type(of: self)): "
         string += "name = \(String(describing: self.name)), "
-        string += "shortName = \(String(describing: self.shortName)), "
-        string += "callName = \(String(describing: self.callName)), "
+        string += "selectorName = \(String(describing: self.selectorName)), "
```
Antondomashnev: Why we removed some strings from the descirption?

Templates/Templates.xcodeproj/project.pbxproj
```diff
@@ -285,7 +285,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "\"${BUILD_DIR}/Debug/Sourcery.app/Contents/MacOS/Sourcery\" --sources \"${SRCROOT}/Tests/Context\" --templates \"${SRCROOT}/Templates\" --output \"${SRCROOT}/Tests/Generated\"";
+			shellScript = "\"${BUILD_DIR}/Debug/Sourcery.app/Contents/MacOS/Sourcery\" --sources \"${SRCROOT}/Tests/Context\" --templates \"${SRCROOT}/Templates\" --output \"${SRCROOT}/Tests/Generated\" --disableCache";
```
Antondomashnev: Thanks for this 👍

Sourcery/Parsing/FileParser.swift
Antondomashnev: Up to you 👍

SourceryRuntime/Sources/Description.generated.swift
ilyapuchka: @krzysztofzablocki should we maybe improve it with your https://github.com/krzysztofzablocki/Difference ?

Antondomashnev: Is that a performance downgrade or what specifically? I thought it was useful to show them when you print an object.

ilyapuchka: It's more of an issue in tests. Not sure how much printing descriptions in templates is used, probably only during development. I would say that docs should be good enough to not need to inspect properties like that, but I can see your point. @krzysztofzablocki what do you think? Maybe if we use Difference we can leave computed properties in description and still have more readable failure descriptions in tests?

krzysztofzablocki: I'd keep it as it is, with removed computed properties. Equality wouldn't even compare them so no point in diffing them, good call.

Our real implementation is better than Difference as it uses real AST and specialized implementation. Difference is for those people that don't use Sourcery for some bizarre and completely wrong reason 😹

PR #499, opened 2018-01-22
fix broken url's

4brunu:
Hey, I was navigation thought some docs and saw some broken url's, so here is the fix.

### DIFF
--- a/guides/Usage.md
+++ b/guides/Usage.md
@@ -6,14 +6,14 @@ Sourcery is a command line tool, you can either run it manually or in a custom b
 $ ./sourcery --sources <sources path> --templates <templates path> --output <output path>
 ```
 
-> Note: this command may be different depending on the way in which you installed Sourcery (see [Installing](installing.html))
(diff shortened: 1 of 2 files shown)


guides/Usage.md
```diff
@@ -6,14 +6,14 @@ Sourcery is a command line tool, you can either run it manually or in a custom b
 $ ./sourcery --sources <sources path> --templates <templates path> --output <output path>
 ```
 
-> Note: this command may be different depending on the way in which you installed Sourcery (see [Installing](installing.html))
+> Note: this command may be different depending on the way in which you installed Sourcery (see [Installing](Installing.md))
```
ilyapuchka: I believe this and other links in md files are not broken. They point to online documentation which is in html files. md files in guides folder are used to generate html documentation with jazzy. You can use it to generate documentation locally so that all links work.

4brunu: Oh, I see.
Should I close this PR them?

ilyapuchka:
I'll just cherry pick your first change

Issue #501, opened 2018-01-23
Modular architecture


ilyapuchka:
Currently Sourcery can be used only as a tool, though theoretically it can be used as a dependency to power other tools. We got such requests several times in the past but for now were rejecting them. I'd like to suggest to think about that again because I think that not only Sourcery parsing functionality can be useful but also its js and swift templates. Here are some steps that we will need to make for that:

- Extract parser and composer to a separate framework so that it will provide all parsing functionality.
- Extract js and swift templates to a separate frameworks
- Make Sourcery just a command line tool that utilise other frameworks (maybe we'll find a way to conditionally import templates frameworks as plugins)
- Make sure everything works with all dependency managers
- [Optionally] move Sourcery to dedicated organisation so that we can group all those projects. Alternatively we can keep everything in one repo and add sub specs, but it will not play very well with Carthage (need to check that and maybe suggest improvement to support building only selected schemes)

krzysztofzablocki:
Other tools use Sourcery as dependency, there are quite a few that do it, what benefits do you see of using this as a framework rather than a command line tool like they do now with swifttemplates?

Sourcery uses macOS target because embedding swift framework has not been supported properly in command line tools, not sure if that changed yet.

I'm not against us modularizing the project into frameworks so people can use just parsing part of Sourcery, but I'd very much would like to keep all of the frameworks in this repo, I've experienced a mess that managing dependencies across multiple repos can be and I'd rather keep using mono repo here.

ilyapuchka:
I think it will be more convenient to get parsed data from some sort of API provided by a framework than invoking command line tool and then working with its results in swift templates, if I understand what you mean. The same way we could probably use SourceKitten as command line tool and then parse its output, but using it as a framework is much better.

If they use it as a framework users will also be able to extend it to their needs (if we allow that of course), maybe using different templates engine, or even utilising parsing results for something other than code generation, i.e. displaying types dependencies graph.

Also it will probably push us to cleaning up our test suite and maybe improve some APIs as they will need to be exposed publicly.

Agree with you about managing dependencies from different repos, was there.

krzysztofzablocki:
templating engines will be pluggable when we do this as our current arch is already pluggable interface so hopefully they will send their extensions upstream to us :)

I remember you wanted to work on extracting Sourcery to a framework @hartbit, would you like to start on this plan ?

hartbit:
[...]

PR #478, opened 2018-01-09
Migrate to Swift 4, SwiftPM 4 and Xcode 9.2

hartbit:
In preparation for a future Pull Request, I migrated Sourcery to Swift 4, I migrated the SwiftPM Package.swift manifest to the version 4, and I got everything working and building in Xcode 9.2

PR #503, opened 2018-01-27
Moved js templates to a separate framework

ilyapuchka:
Moved ejs templates to a separate framework as discussed in #501 and slightly changed its API to make it more generic. Have no idea how to set it up for CocoaPods so that SourceryJS can be used as a separate dependency. Probably requires creating another spec for it, but it will depend on SourceryRuntime which should be then extracted too, but then we will need to publish all of them separately. Unfortunately subspec seems to be not suitable for such setup. Advices will be appreciated.


krzysztofzablocki:
you can have a single podspec with multiple specifications and single update mechanism, take a look at this: https://github.com/mxcl/PromiseKit/blob/master/PromiseKit.podspec

ilyapuchka:
I was trying to do that with subspecs, but if I understand correctly they still produce a single framework target with a name defined in the parent spec. When I wanted to describe several targets which depend on SourceryRuntime framework for each framework and also a tool itself, which has only binary source, not just a single target with subset of functionality. 
When I tried to do that and was linting SourceryJS subspec I saw that in the dummy project it was creating `Sourcery` framework built of SourceryRuntime sources, but I'm importing SourceryRuntime, so it fails to build.

krzysztofzablocki:
ah well, I guess we'll need to update create separate spec for the each framework then ¯\_(ツ)_/¯

Btw the code looks fine, I've not used ejs, but as long as tests will pass this is fine to merge

ilyapuchka:
I think I'll need #478 to fix SPM (can't find a way to define a dependency for additional target), so I'll merge it first and them will rebase this

ilyapuchka:
Should work now. 
@krzysztofzablocki I would merge it then and continue extraction of other parts. I think we can figure out what to do with CocoaPods later. I don't like much the idea of having two separate specs. We probably can try to use subspecs to allow users to choose what functionality they need, but provide everything under one framework target (similar to SourceKitten). We will need some alternations in imports in that case though (i.e. do not import SourceryRuntime in SourceryJS sources when we build them into one target)

PR #505, opened 2018-01-30
Swift templates as a separate framework

ilyapuchka:
#501


SourcerySwift/Sources/SwiftTemplate.swift
```diff
@@ -0,0 +1,322 @@
...
+}
+
+open class SwiftTemplate {
+
+    public let sourcePath: Path
+    let cachePath: Path?
+    let code: String
+
+    private lazy var buildDir: Path = {
+        guard let tempDirURL = NSURL(fileURLWithPath: NSTemporaryDirectory()).appendingPathComponent("SwiftTemplate.build") else { fatalError("Unable to get temporary path") }
```
ilyapuchka: this can be probably rewritten using PathKit

ilyapuchka:
I'd love to see it (and JS templates) independent of SourceryRuntime, but couldn't figure out a way to do that yet.

Issue #479, opened 2018-01-10
Improve automockable template to support clousure parameter


ignazioc:
I'm using this version of the AutoMockable.stencil
https://raw.githubusercontent.com/krzysztofzablocki/Sourcery/master/Templates/Templates/AutoMockable.stencil

When a protocol contains a method with a clousure parameter like this:

```swift
protocol SampleObjectProtocol: AutoMockable {
    func makeCall(completion: @escaping (String) -> Void)
}
```

The generated code looks like this

```swift
class SampleObjectProtocolMock: SampleObjectProtocol {

    //MARK: - makeCall

    var makeCallCompletionCallsCount = 0
    var makeCallCompletionCalled: Bool {
        return makeCallCompletionCallsCount > 0
    }
    var makeCallCompletionReceivedCompletion: ((String) -> Void)?

    func makeCall(completion: @escaping (String) -> Void) {
        makeCallCompletionCallsCount += 1
        makeCallCompletionReceivedCompletion = completion
    }

}
```

The issue is that the clousure is not executed and the parameters of the clousure cannot be overridden. The solution I have in mind would looks like this

```swift

class SampleObjectProtocolMock: SampleObjectProtocol {

    //MARK: - makeCall

    var makeCallCompletionCallsCount = 0
    var makeCallCompletionCalled: Bool {
        return makeCallCompletionCallsCount > 0
    }
    var makeCallCompletionReceivedCompletion: ((String) -> Void)?

    var makeCallCompletionFirstParameter: String = ""

    func makeCall(completion: @escaping (String) -> Void) {
        makeCallCompletionCallsCount += 1
        makeCallCompletionReceivedCompletion = completion
        completion(makeCallCompletionFirstParameter)
    }

}
```

krzysztofzablocki:
hey @ignazioc, usually you are unit testing that the makeCall was called and with what arguments, calling closure automatically doesn't have a lot of merit if the mock is fully generated. 
Normally you'd just call `makeCallCompletionReceivedCompletion` closure with your arguments ? after mocking if you wanted to have more complex unit tests.

Can you share some example tests you are using?

ignazioc:
The idea behind is similar to what the template already does with functions with a return value.

```swift 
protocol Customer: AutoMockable {
    func fullName() -> String
}


class CustomerMock: Customer {

    //MARK: - fullName

    var fullNameCallsCount = 0
    var fullNameCalled: Bool {
        return fullNameCallsCount > 0
[...]

PR #512, opened 2018-02-05
Added closures to AutoMockable

albsala:
Many times you need to mock a method returning something that depends on the parameters. 
This a backward compatible change on AutoMockable to allow doing that without subclassing.

This would be an example with one of the protocols in `Templates` tests:
```swift
let mock: = ReservedWordsProtocolMock()
mock.continueWithClosure = { $0 }
let result = (mock as ReservedWordsProtocol).`continue`(with: "Foo")
print(result) // prints "Foo"
```


krzysztofzablocki:
I don't know how I feel about this change, when you write tests you usually know the parameters that will be passed so why don't you just set the return value using that knowledge?

This reminds me of discussion in https://github.com/krzysztofzablocki/Sourcery/issues/479

Thoughts @Antondomashnev @ilyapuchka ?

albsala:
Thank you for your comments @krzysztofzablocki, they show that `ReservedWordsProtocolMock` was definitely not the best example :)

Lets say you need to mock a Datasource such as:
```swift
protocol FooDataSource: AutoMockable {
    func item(ForRowAt indexPath: IndexPath) -> Int
}
```

The generated mock would be:
```swift
class FooDataSourceMock: FooDataSource {
    
    //MARK: - item
    
    var itemForRowAtCallsCount = 0
    var itemForRowAtCalled: Bool {
        return itemForRowAtCallsCount > 0
    }
    var itemForRowAtReceivedIndexPath: IndexPath?
    var itemForRowAtReturnValue: Int!
    var itemForRowAtClosure: ((IndexPath) -> Int)?
    
    func item(ForRowAt indexPath: IndexPath) -> Int {
        itemForRowAtCallsCount += 1
        itemForRowAtReceivedIndexPath = indexPath
        return itemForRowAtClosure != nil ? itemForRowAtClosure!(indexPath) : itemForRowAtReturnValue
    }
    
}
```

And you could use it like this:
```swift
let model: [[Int]] = [[1,2,3],[4,5,6]]
let mock = FooDataSourceMock()
mock.itemForRowAtClosure = { model[$0.section][$0.row] }
```

krzysztofzablocki:
ok, your implementation is non-intrusive and backward compatible so that's fine with me, I can see value in using something like beforeEach to configure the default responses for test suite like that rather than do it in each test.

Templates/Tests/Expected/AutoMockable.expected
```diff
@@ -79,41 +88,52 @@ class ExtendableProtocolMock: ExtendableProtocol {
...
 
     var startCallsCount = 0
     var startCalled: Bool {
         return startCallsCount > 0
     }
+	var startClosure: (() -> Void)?
 
     func start() {
         startCallsCount += 1
+		startClosure?()
```
krzysztofzablocki: whitespaces feel off, can you adjust the template so they match other formatting?

Templates/Tests/Expected/AutoMockable.expected
```diff
@@ -125,11 +145,12 @@ class ReservedWordsProtocolMock: ReservedWordsProtocol {
     }
     var continueWithReceivedMessage: String?
     var continueWithReturnValue: String!
+	var continueWithClosure: ((String) -> String)?
 
     func `continue`(with message: String) -> String {
         continueWithCallsCount += 1
         continueWithReceivedMessage = message
-        return continueWithReturnValue
+        return continueWithClosure != nil ? continueWithClosure!(message) : continueWithReturnValue
```
krzysztofzablocki: how about adjusting the template so that you don't need to force unwrap it ?
```swift
continueWithClosure.map({ continueWithClosure(message) }) ?? continueWithReturnValue
```

Antondomashnev:
@krzysztofzablocki will take a look later today. Thanks for the initial review 👍

albsala:
Thank you @Antondomashnev, I've fixed the spacing problem. I'm sorry but I don't know exactly what you mean with putting closures only in one protocol. It's generated for every protocol/class in the same way as Called or CallsCount.

Antondomashnev:
@albsala my bad, you are right =)
Merging it 👍

PR #514, opened 2018-02-07
Automatically add generated files to the project

### LINKED ISSUE #457: How do I include the generated swift files into the Xcode automatically?
Hi,

I have got my sources generating to a folder inside my project. This is my .sourcery.yml file 
```
sources:
  - ./Ayuda/
templates:
  - ./sourcery-templates/
output:
  ./Ayuda/GeneratedCode/
```

How to add any file getting created inside the GeneratedCode folder into the Xcode project automatiically? I have a Run Script section that generates my sourcery code for me.

Thanks
Sethu

ilyapuchka:
Resolves #457 
I've switched to xcproj to read and write project files, where I've also added convenience methods to add source files to projects and targets so that this logic does not pollute our sources.
This change is also backward compatible as output path can be still provided as a string, but optionally can be provided as an object with additional keys.


Sourcery/Sourcery.swift
```diff
@@ -211,9 +213,9 @@ class Sourcery {
         }
     }
 
-    fileprivate func outputPaths(from: Paths, output: Path) throws -> [Path] {
-        return templatePaths(from: from).map { output + generatedPath(for: $0) }
-    }
+//    fileprivate func outputPaths(from: Paths, output: Path) throws -> [Path] {
```
krzysztofzablocki: we should remove this

Sourcery/Sourcery.swift
```diff
@@ -347,36 +348,59 @@ extension Sourcery {
...
+    }
+
     private func output(result: String, to outputPath: Path) throws {
+        var result = result
         if !result.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
-            try writeIfChanged(Sourcery.generationHeader + result, to: outputPath)
+            if outputPath.extension == "swift" {
+                result = Sourcery.generationHeader + result
+            }
+            if !outputPath.parent().exists {
```
krzysztofzablocki: whats going to happen if you auto link a file in --watch mode and then the template is nil and the file gets pruned? or do you have a guard for that

ilyapuchka: Hm, good catch. It will keep the file linked to the project so it will not build any more. So I guess we should "unlink" it when it is being deleted.

ilyapuchka: @krzysztofzablocki it looks cumbersome to implement though as we don't store file reference anywhere and then will have to go through all the project files to find one to remove. I will just add the check for file existence so that it does not add files that are removed.

ilyapuchka: Actually it is [already there](https://github.com/xcodeswift/xcproj/blob/master/Sources/xcproj/PBXProjObjects%2BHelpers.swift#L102) in xcodeproj methods so, I guess no need to do anything on our side

dcoletto:
I am using sourcery 0.16.0 but generated code is not added to my xcode project. The code is actually generated and I can see it with finder, but it's not visible in xcode (I need to add it manually). Am I missing something? I'm calling sourcery like this:
`$PODS_ROOT/Sourcery/bin/sourcery --sources $SRCROOT --templates $PODS_ROOT/RetrofireSwift/RetrofireSwift/Templates/RetrofireSourcery.stencil --output $SRCROOT/Generated/`

Liquidsoul:
Hi @dcoletto,
You need to use a configuration file with the `link` option (as described [here](https://github.com/krzysztofzablocki/Sourcery/blob/master/guides/Usage.md#output)) to achieve this. You cannot do that just using the command line like you did because Sourcery has no knowledge of your Xcode project.

Issue #420, opened 2017-09-22
Sourcery doesn't work with Apollo GraphQL generated API.swift


va1da:
Hi,
I'm using Sourcery to generate methods for GraphQL API.
Currently as generated API.swift is about 710 KB Sourcery gets stuck (eats all my RAM and then available disk space).

I understand that there are a lot of types in API.swift, but even though, it could work on 8GB RAM notebook?

Anyone having similar issue?

Xcode 9
[API.swift.zip](https://github.com/krzysztofzablocki/Sourcery/files/1324683/API.swift.zip)

ilyapuchka:
Thanks for providing this example, we can use it in future for performance testing. For now I think Sourcery code is just not optimised to deal with very big amount of data.

krzysztofzablocki:
@va1da I'll take a look, can you share the template you have been trying to use for it?

va1da:
[GraphQLQuery.stencil.zip](https://github.com/krzysztofzablocki/Sourcery/files/1430868/GraphQLQuery.stencil.zip)

Antondomashnev:
@krzysztofzablocki FYI I've checked it a while ago and because there are plenty of nested of levels of recursion here:
```
private func walkDeclarations(source: [String: SourceKitRepresentable], containingIn: (Any, [String: SourceKitRepresentable])? = nil, processed: inout [[String: SourceKitRepresentable]], foundEntry: (SwiftDeclarationKind, String, AccessLevel, [String], [String: SourceKitRepresentable], Any?) -> Any?) {
        if let substructures = source[SwiftDocKey.substructure.rawValue] as? [SourceKitRepresentable] {
            for substructure in substructures {
                if let source = substructure as? [String: SourceKitRepresentable] {
                    processed.append(source)
                    walkDeclaration(source: source, containingIn: containingIn, foundEntry: foundEntry)
                }
            }
        }
    }
```
It ends up consuming a huge amount of memory.

lvsti:
@Antondomashnev FWIW I've put a recursion counter into `walkDeclarations` and with a similar 500+KB swift file it never exceeded 4 levels, so the culprit might not necessarily be the recursion itself.

Antondomashnev:
@lvsti hey, thanks for the follow-up. The generated Graphql file contains lots of types, so I guess that lead to the problem with recursion. But maybe I'm missing the root cause =)

tamasprezi:
[...]

PR #519, opened 2018-02-11
[WIP] Workaround for string cache bloating in the parsing phase

tamasprezi:
See discussion in #420


ilyapuchka:
I'll merge this, let's see if this will cause any crashes for users, I think it still might on code bases which contain unicode characters.

ilyapuchka:
Resolves #420

ilyapuchka:
Thanks @tamasprezi !

PR #522, opened 2018-02-14
[WIP] Encoding/Decoding Template

thevest:
An implementation of Codable via Sourcery through intermediate objects. I stripped down the template slightly from what I posted on the issue, though theres a lot of other things that can be done with it depending on what you use for persistence. 

I have run into a slight problem, where the tests were working, and now they throw an error when looking for the generated file in the bundle. I'm not quite sure what I'm doing wrong, since according to my build phases settings it is being copied to the bundle. If anyone could take a look and let me know what I'm doing wrong I'd be happy to fix the test.


Templates/Templates/ModelMappable.stencil
```diff
@@ -0,0 +1,162 @@
+// https://github.com/SwiftGen/StencilSwiftKit/blob/master/Documentation/tag-set.md
```
ilyapuchka: I understand that you want this for documentation purposes, but that pollutes generated files.

thevest: I’ll remove the doc links.

Templates/Templates/ModelMappable.stencil
```diff
@@ -0,0 +1,162 @@
...
+{% macro codingKeys variables %}
+  private enum CodingKeys: String, CodingKey {
+    {% for var in variables %}
+    case {{ var.name }} = "{% call remoteName var %}"
+    {% endfor %}
+  }
+{% endmacro %}
+
+{% macro decodeVar var %}
+  {% set itype %}{{ var.annotations.incomingType }}{% endset %}
```
ilyapuchka: It might be that some api methods return the same models with different primitive types, strings in place of numbers and so on. Using annotation to fix type that comes in limits it to one particular scheme. I believe this should be handled by extensions on decoder instead that will not just decode bool from bool, but will support any compatible value (0/1, "true"/"false"), same for other types.

Templates/Templates/ModelMappable.stencil
```diff
@@ -0,0 +1,162 @@
...
+  {% if var.typeName.name == "Bool" and itype == "Int" %} {# its a 0 or 1 #}
+    {{ var.name }} = try container.decodeIfPresent(Int.self, forKey: .{{ var.name }}) == 1
+  {% elif var.typeName.name == "Date" and itype == "String" %} {# iso date string #}
+    if
+      let {{ var.name }} = try container.decodeIfPresent(String.self, forKey: .{{ var.name }}),
+      let date = {{ var.name }}.dateFromISO8601 {
```
ilyapuchka: I didn't use it, but isn't it what date decoding strategy is for?

Templates/Templates/ModelMappable.stencil
```diff
@@ -0,0 +1,162 @@
...
+{% macro decodeWithDefault var %}
+  {% set typename %}{{ var.typeName.name }}{% endset %}
+  {% if typename == "String" %}
+    {{ var.name }} = try container.decodeIfPresent({{ typename }}.self, forKey: .{{ var.name }}) ?? ""
+  {% elif typename == "Bool" %}
+    {{ var.name }} = try container.decodeIfPresent({{ typename }}.self, forKey: .{{ var.name }}) ?? false
+  {% elif typename == "Int" or typename == "Double" or typename == "Float" or typename == "NSNumber" %}
+    {{ var.name }} = try container.decodeIfPresent({{ typename }}.self, forKey: .{{ var.name }}) ?? 0
+  {% elif typename == "Date" %}
+    {{ var.name }} = try container.decodeIfPresent({{ typename }}.self, forKey: .{{ var.name }}) ?? Date()
```
ilyapuchka: not sure that current date is a good default. we could use annotation to specify default value and extend this macros to support any variable annotated with it.

Templates/Templates/ModelMappable.stencil
```diff
@@ -0,0 +1,162 @@
...
+  {% endif %}
+{% endmacro %}
+
+{% macro decodableInitializer variables %}
+  init(from decoder: Decoder) throws {
+    let container = try decoder.container(keyedBy: CodingKeys.self)
+    {% for var in variables where var|annotated:"incomingType" %}
+    {% call decodeVar var %}
+    {% endfor %}
+    {% for var in variables where var|!annotated:"incomingType" %}
```
ilyapuchka: I think would be better if variables will go in the order in which they are defined, so that you check for annotation in loop body, also will make it faster.

Templates/Templates/ModelMappable.stencil
```diff
@@ -0,0 +1,162 @@
...
+    dictionary["{{ var.name }}"] = {{ var.name }}.asTimestamp
+    {% else %}
+    dictionary["{{ var.name }}"] = {{ var.name }}
+    {% endif %}
+    {% endfor %}
+    return dictionary
+  }
+{% endmacro %}
+
+{% for type in types.based.ModelMappable where type|protocol and not type.name == "ModelMappable" %}
```
ilyapuchka: why `ModelMappable`? Other templates use another pattern that will make this `AutoCodable`. I would also split it into `AutoDecodable` and `AutoDecodable`, not always you need encoding, most often you need just decoding.

Templates/Tests/Context/ModelMappable.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+
+import Foundation
+
+protocol ModelMappable { }
+
+protocol ExampleMapping: ModelMappable {
+  var string: String { get }
+  var bool: Bool { get }
+  // sourcery: coding="some_variable_with_a_long_key"
```
ilyapuchka: I would use `coding_` as common prefix for annotations related to this template, so this annotation would be `coding_key`

Templates/Tests/Context/ModelMappable.swift
```diff
@@ -0,0 +1,23 @@
...
+  var someVariable: Double { get }
+  // sourcery: incomingType="Int"
+  var someBoolThatsAnInt: Bool { get }
+  // sourcery: incomingType="String"
+  var someDateString: Date { get }
```
ilyapuchka: Dates should support different strategies, they are enums so this can be just a raw name of enum that will be passed to decoder.

Templates/Tests/Expected/ModelMappable.expected
```diff
@@ -0,0 +1,91 @@
...
+
+// MARK: ExampleModel Mappable
+struct ExampleModel: Codable, ExampleMapping {
+  let string: String
+  let bool: Bool
+  let someVariable: Double
+  let someBoolThatsAnInt: Bool
+  let someDateString: Date
+  let whenItEndsInID: String
+  // MARK: ExampleModel CodingKeys
```
ilyapuchka: I would add spaces between blocks of code to make it more readable

Templates/Tests/Expected/ModelMappable.expected
```diff
@@ -0,0 +1,91 @@
...
+    var container = encoder.container(keyedBy: CodingKeys.self)
+    try container.encode(string, forKey: .string)
+    try container.encode(bool, forKey: .bool)
+    try container.encode(someVariable, forKey: .someVariable)
+    try container.encode(someBoolThatsAnInt ? 1 : 0, forKey: .someBoolThatsAnInt)
+    try container.encode(someDateString.iso8601, forKey: .someDateString)
+    try container.encode(whenItEndsInID, forKey: .whenItEndsInID)
+  }
+  // MARK: ExampleModel Dictionary
+  func asDictionary() -> [String: Any] {
```
ilyapuchka: I think this should be optional, probably with separate protocol, like `AutoDictionaryConvertible` or something

ilyapuchka:
@thevest thanks for you contribution! I left some of my thoughts.

Templates/Templates/ModelMappable.stencil
thevest: Won't the expected type still need to be known for encoding the value correctly?

thevest:
@ilyapuchka I appreciate the feedback, and I agree it make more sense to separate the encoding, decoding, and dictionary conversion into separate protocols / templates. I'll see what I can do.

Templates/Templates/AutoDecodable.stencil
```diff
@@ -0,0 +1,37 @@
...
+    {% for variable in type.allVariables where variable|!annotated:"skipCoding" %}
+    case {{ variable.name }} = "{% call keyedNameFor type variable %}"
+    {% endfor %}
+  }
+{% endmacro %}
+
+{# going to exclude protocols for now, handle it seperatly #}
+{% for type in types.all|!protocol where type.based.AutoDecodable or type|annotated:"AutoDecodable" and not type.name == "AutoCodable" and not type.name == "AutoDecodable" %}
+  // MARK: {{ type.name }}+Decodable | AutoDecodable
+  // sourcery:inline:auto:{{ type.name }}.AutoDecodable
```
ilyapuchka: I'm currently working on such template myself and I noticed that when generating constructor in the type body Swift stops to generate member wise constructors, which may just break code that was relying on such constructor to be generated. So I would suggest to not use inline here and instead generate this constructor in extension of type.

thevest: When I was trying to do that the compiler was complaining about required constructors outside the class body (and I believe the decoding constructor has to be required?), but I'll see if theres a way around that.

ilyapuchka: I think this is only needed for classes, for value types `required` is not needed. So I guess classes should be handled separately, with inheritance in mind for which there is superDecoder/Encoder, as far as I know

Templates/Templates/AutoDecodable.stencil
```diff
@@ -9,28 +11,85 @@ extension KeyedDecodingContainer {
...
+{# todo inheritance #}
+{% for type in types.all|class where type.based.AutoDecodable or type|annotated:"AutoDecodable" and type.name != "AutoCodable" and type.name != "AutoDecodable" %}
   // sourcery:inline:auto:{{ type.name }}.AutoDecodable
   {% call codingKeys type %}
-  {{ type.accessLevel }}{% if type|class %} convenience required{% endif %} init(from decoder: Decoder) throws {
-    {% if type|class %}self.init(){% endif %}
+  {% if type.supertype.based.AutoEncodable or type.based.AutoEncodable %}
+  {% call encode type %}
+  {% endif %}
+  {{ type.accessLevel }} convenience required init(from decoder: Decoder) throws {
```
ilyapuchka: why is this marked as convenience?

Templates/Templates/AutoDecodable.stencil
```diff
@@ -9,28 +11,85 @@ extension KeyedDecodingContainer {
...
+    {% call inheritanceInitializer type %}
```
ilyapuchka: why is this constructor marked as convenience and why this macro is called at the beginning? Shouldn't it just call `super.init` in the end? Also isn't it that only own properties of subclass should be decoded (and have keys defined for them, though extra keys will not raise issues) and the rest should be delegated to super class constructor?

Templates/Templates/ModelMappable.stencil
ilyapuchka: After implementing it myself I actually found it useful to sometimes declare expected type of value, but that always comes with subsequent mapping of this type to the type of corresponding variable, i.e. String -> Double. So I'm using `decoding_type` and `decoding_map` annotations for that.

Templates/Templates/AutoDecodable.stencil
```diff
@@ -9,28 +11,85 @@ extension KeyedDecodingContainer {
...
+  private enum CodableKeys: String, CodingKey {
+    {% for variable in type.instanceVariables where variable|!annotated:"skipCoding" %}
+    case {{ variable.name }} {% call keyedNameFor type variable %}
+    {% endfor %}
+  }
+{% endmacro %}
+
+{% macro inheritanceInitializer type %}
+  {% if type.supertype.based.AutoDecodable or type.supertype.based.Decodable %}
+    try self.init(from: decoder)
```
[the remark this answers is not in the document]
ilyapuchka: Also this should use `decoder.superDecoder`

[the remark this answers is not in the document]
thevest: When we're inheriting a type that is codable the initializer has to be convenience otherwise we can't override the original required initializer. At least thats what I found while working through this. I'll see if I can do it another way, however this seems to work without the superDecoder/encoder. For pure swift serializations using the superDecoder and superEncoder works, however, you end up with all of you super keys being at a different level of the final hash, i.e. inside a top level key `super`, which wouldn't be what you would want for api responses (imo).

Templates/Templates/AutoDecodable.stencil
```diff
@@ -9,28 +11,85 @@ extension KeyedDecodingContainer {
...
+  {% else %}
+    self.init()
+  {% endif %}
+{% endmacro %}
+
+
+{% macro convertTypeForEncoder var %}
```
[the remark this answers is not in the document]
ilyapuchka: I believe you can automatically inline code from different template, they will just place generated code in different annotation sections.

Templates/Templates/AutoDecodable.stencil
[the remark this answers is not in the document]
ilyapuchka: Yes, keys separation is not helpful in this case, but I still don't see how this will work with `let` properties.

Templates/Templates/AutoDecodable.stencil
```diff
@@ -0,0 +1,96 @@
+import Foundation
+
+extension KeyedDecodingContainer {
+  public func decodeOrDefault<T: Decodable>(_ type: T.Type, forKey key: Key, defaultValue: T) throws -> T {
+    guard let value: T = try decodeIfPresent(type, forKey: key) else {
```
ilyapuchka: If there is default value then, I think, there is no need to throw when key is not there and default value should be applied, so it should be `try?` and method should not throw any error. I ended up using pair of `decodeSafely` and `decodeSafelyIfPresent` similar to described here https://kean.github.io/post/codable-tips-and-tricks

Templates/Templates/AutoDecodable.stencil
```diff
@@ -0,0 +1,96 @@
...
+      return defaultValue
+    }
+    return value
+  }
+}
+
+{# these are a giant line because of spacing, to debug just seperate the if into lines#}
+{% macro defaultValueFor variable %}{% set typename %}{{ variable.typeName.name }}{% endset %}{% if variable|annotated:"codingDefaultValue" %}{{ variable.annotations.codingDefaultValue }}){% elif variable.isOptional %}nil){% elif variable.isArray %}[]){% elif typename == "String" %}""){% elif typename == "Int" or typename == "Double" or typename == "Float" or typename == "NSNumber" %}0){% elif typename == "Bool" %}false){% endif %}{% endmacro %}
+{% macro decodeNameWithTransform name transform %}{% if transform == "camelToSnake" or not transform %}{{ name|camelToSnakeCase }}{% elif transform == "camelToSnake:idToUUID" %}{{ name|camelToSnakeCase|replace:"id","uuid" }}{% elif transform == "idToUUID" %}{{ name|replace:"id","uuid" }}{% endif %}{% endmacro %}
```
ilyapuchka: I still don't think "uuid" conversion is something very common, so should not be part of template. And `replace:"id","uuid"` will replace `id` anywhere in the key so it should be checked first if the key is equal to "id"

Templates/Templates/AutoDecodable.stencil
```diff
@@ -0,0 +1,96 @@
...
+{% macro keyedNameFor type  variable %}{% if variable|annotated:"codingKey" %} = "{{ variable.annotations.codingKey }}"{% elif type|annotated:"codingKeyTransform" %}= "{% call decodeNameWithTransform variable.name type.annotations.codingKeyTransform %}"{% endif %}{% endmacro %}
+
+{# the keys to use to decoding #}
+{% macro codingKeys type %}
+  private enum CodableKeys: String, CodingKey {
```
ilyapuchka: I found it useful to have `DecodingKeys` and `EncodingKeys` and make them public. This way they can be used elsewhere in the code.

Templates/Templates/AutoDecodable.stencil
```diff
@@ -0,0 +1,96 @@
...
+    {{ var.name }} = try container.decodeOrDefault({{ var.typeName.name }}.self, forKey: .{{ var.name }}, defaultValue: {% call defaultValueFor var %}
+    {% else %}
+    {{ var.name }} = try container.decode({{ var.typeName.name }}.self, forKey: .{{ var.name }})
+    {% endif %}
+    {% endfor %}
+  }
+}
+{% endfor %}
+
+{# todo inheritance #}
```
ilyapuchka: I think it should be completely removed from template until it is implemented fully or at least up to some extent.

Templates/Tests/Generated/AutoEncodable.generated.swift
```diff
@@ -0,0 +1,35 @@
...
+    case prim3  = "auto_codable_prim_3"
+    case arr 
+    case arrOpt 
+    case date 
+    case optDate 
+  }
+
+  internal func encode(to encoder: Encoder) throws {
+    var container = encoder.container(keyedBy: EncodableKeys.self)
+    try container.encode(bool ? 1 : 0, forKey: .bool)
```
ilyapuchka: this will encode Int, wo it will not be decoded as Bool, I supposed, so this should be just `bool`. The case when API expects 1/0 instead of true/false should be handled manually.

Templates/Tests/Generated/AutoEncodable.generated.swift
```diff
@@ -0,0 +1,35 @@
...
+// DO NOT EDIT
+
+
+
+
+
+
+// MARK: AutoDecodableTest+Encodable | AutoEncodable
+extension AutoDecodableTest {
+  private enum EncodableKeys: String, CodingKey {
```
ilyapuchka: `EncodingKeys` would be a better name.

Templates/Tests/Generated/AutoDecodable.generated.swift
```diff
@@ -0,0 +1,49 @@
...
+
+
+
+
+
+
+
+// MARK: AutoDecodableTest+Decodable | AutoDecodable
+extension AutoDecodableTest {
+  private enum CodableKeys: String, CodingKey {
```
ilyapuchka: As mentioned above this would be better to be `DecodingKeys`

guides/Codable.md
```diff
@@ -0,0 +1,102 @@
+## I want to make my types conform to `Encodable` or `Decodable`
+
+### [Encodable Template](https://github.com/krzysztofzablocki/Sourcery/blob/master/Templates/Templates/Encodable.stencil)
+### [Decodable Template](https://github.com/krzysztofzablocki/Sourcery/blob/master/Templates/Templates/Encodable.stencil)
```
ilyapuchka: link points to wrong template

guides/Codable.md
```diff
@@ -0,0 +1,102 @@
...
+
+#### Available annotations
+
+
+* `skipCoding` - skips decoding and encoding
```
ilyapuchka: I believe this annotation should be applied only for variables (should be mentioned in description) and only for encoding, so should be `skipEncoding`. For types we wouldn't need it until inheritance is implemented.

guides/Codable.md
```diff
@@ -0,0 +1,102 @@
...
+* `codingKey` - The coding key to use, this is optional and if no `codingKey` annotation is found the variable name is used.
```
ilyapuchka: Can be useful to have also `decodingKey` and `encodingKey` in case they happen to be different

guides/Codable.md
```diff
@@ -0,0 +1,102 @@
...
+    case prim1  = "auto_codable_prim_1"
+    case prim2  = "auto_codable_prim_2"
+    case prim3  = "auto_codable_prim_3"
+    case arr 
+    case arrOpt 
+    case date 
+    case optDate 
+  }
+
+  internal init(from decoder: Decoder) throws {
```
ilyapuchka: I discovered that Swift 4.1 will warn that this constructor does not call member-wise constructor, because of cross-module constructors changes. So we need to take care of that in the template, which will be harder as there might be no such constructor (maybe we should generate one if we can't find it in type)

ilyapuchka:
There are a lot of things to consider so I'll close this for now. @thevest if you will have time to come back and address my comments new PR will be very welcome. I'll will work on a different version of the template where will try to rely less on annotations and follow closer to standard library's approach, using types information instead.

thevest:
@ilyapuchka Lookin forward to your version. Thanks for the feedback, just ran out of free time unfortunately.

PR #525, opened 2018-02-18
Create lenses only for stored properties

cnicolai:
Hi there, two small suggestions regarding AutoLenses. I guess instead of opening an issue, a PR is more helpful, so here it comes.

Two changes, the first will create lenses only for stored properties. I guess they are not useful for computed properties?
The second change results in the type's initializer being only called with the stored properties, not the computed.


ilyapuchka:
@cnicolai can you please update/add tests to templates tests?

cnicolai:
@ilyapuchka If that means updating "Context" and "Expected" in the template test folder, I could do that.

PR #554, opened 2018-03-20
option to ignore methods in extensions for auto mockable template

miguelangel-dev:
Added possibility to ignore auto mockable in some methods defined|default implementation in extensions.

In some cases when we give some default implementation to protocols, for example using "when" clause this template generate duplicated methods.

Cause:
```
protocol A: AutoMockable {
   func a() -> String
   func b()
}

extension A where Self: B {
   func a() -> String { return "duplicated.method.a" }
}
```

Solution:
```
protocol A: AutoMockable {
   func a() -> String
   func b()
}

extension A where Self: B {
   // sourcery:begin: ignore
   func a() -> String { return "duplicated.method.a" }
   // sourcery:end
}
```


ilyapuchka:
@miguelangel-dev I think it would be better to give this annotation more domain specific name, like `skipMocking`

Issue #531, opened 2018-02-22
Templates documentation api


ilyapuchka:
Annotations are not compiled and very easy to type wrong. Then it can produce confusing result for which requires to read template carefully. To improve understanding of templates we can introduce documentation api that can be used in header comment of template and can look like this:

```
{#
desc: description of template
annotations:
  some_annotation: description of annotation
  another_annotation: description of annotation
#}
```

Then we can parse annotations from in these comments and when we parsing them from code and find unknown annotation we can suggest user with correct syntax using the same approach as in #471.
We can do the same for arguments, though they are less used and only passed once when running tool.

PR #558, opened 2018-03-21
Feature/557 template to generate protocols from types

doozMen:
I would still need some help with the tests. But this adds:

Template for generating protocols
Example of use


Templates/Templates/AutoGenerateProtocol.stencil
```diff
@@ -0,0 +1,32 @@
...
+//: - initializers marked with annotation // sourcery:includeInitInProtocol
+//: - of the above it does not add it if  // sourcery:skipProtocol
+//: ---
+
+{% for type in types.implementing.AutoGenerateProtocol %}
+// MARK: - {{ type.name }} AutoGenerated protocol
+protocol {{ type.name }}Protocol: AutoMockable {
+    {#loop over all variables from this type and not from the super class#}
+    {% for variable in type.variables where variable.readAccess != "private" %}
+    {%if not method.annotations.skipProtocol %}var {{variable.name}}: {{variable.typeName}} { {%if variable.isMutable %} get set {% else %} get {% endif %} }{% endif %}
```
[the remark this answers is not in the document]
ilyapuchka: And you can simplify by checking for mutability only to add `set` so you will not need `else` branch

Templates/Templates/AutoGenerateProtocol.stencil
```diff
@@ -0,0 +1,32 @@
...
+    {% endfor %}
+
+    {#optionally add initializers when included in protocol#}
+    {% for method in type.methods where method.annotations.includeInitInProtocol %}
+    {{method.name}}
+    {% endfor %}
+
+    {#loop over all methods static class or normal#}
+    {% for method in type.methods where method.accessLevel != "private" and method.isInitializer == 0 %}
```
ilyapuchka: why `method.isInitializer == 0` and not just `not method.isInitializer`?

Templates/Templates/AutoGenerateProtocol.stencil
```diff
@@ -0,0 +1,32 @@
...
+//: From all Types implementing this protocol Sourcery adds:
+//: - public/internal variables // private variables are ignored
+//: - public/internal methods (skips initializers)
+//: - initializers marked with annotation // sourcery:includeInitInProtocol
+//: - of the above it does not add it if  // sourcery:skipProtocol
+//: ---
+
+{% for type in types.implementing.AutoGenerateProtocol %}
+// MARK: - {{ type.name }} AutoGenerated protocol
+protocol {{ type.name }}Protocol: AutoMockable {
```
ilyapuchka: I guess your main intention for this template is to generate protocols for mocks, but I think this conformance should be something that should be added manually.

Templates/Templates/AutoGenerateProtocol.stencil
```diff
@@ -0,0 +1,32 @@
+
+//: Do not change this code as it is autogenerated every time you build.
+//: You can change the code in `../StencilTemplatesForSourcery/Application/AutoGenerateProtocol
```
ilyapuchka: What is this path?

Templates/Templates/AutoGenerateProtocol.stencil
```diff
@@ -0,0 +1,32 @@
+
+//: Do not change this code as it is autogenerated every time you build.
```
ilyapuchka: We already have warning, don't think there is need to state this in template too.

Templates/Templates/AutoGenerateProtocol.stencil
```diff
@@ -0,0 +1,32 @@
...
+//: You can change the code in `../StencilTemplatesForSourcery/Application/AutoGenerateProtocol
+
+// MARK: - AutoGenerateProtocol
+//: From all Types implementing this protocol Sourcery adds:
```
ilyapuchka: Not sure if documentation comments are needed in the generated code. I've proposed a format for templates documentation here #531 but didn't touch on that. @krzysztofzablocki what do you think?

Templates/Tests/Expected/AutoGenerateProtocol.expected
```diff
@@ -0,0 +1,25 @@
...
+// MARK: - AutoGenerateProtocol
+//: From all Types implementing this protocol Sourcery adds:
+//: - public/internal variables // private variables are ignored
+//: - public/internal methods (skips initializers)
+//: - initializers marked with annotation // sourcery:includeInitInProtocol
+//: - of the above it does not add it if  // sourcery:skipProtocol
+//: ---
+
+// MARK: - Car AutoGenerated protocol
+protocol CarProtocol: AutoMockable {
```
ilyapuchka: you should probably change either expected type or type in source. And you should actually test this template in `TemplatesTests.swift`

Templates/Templates/AutoGenerateProtocol.stencil
krzysztofzablocki: Might be useful for people unfamiliar with Sourcery, but if we do add docs to generated code then I'd be 👍 for using the format you described, that we could eventually actually parse and run some extra checks on the tool side

doozMen:
@ilyapuchka thanks for the feedback. I will look into it this week.

PR #560, opened 2018-03-22
Feature/559 template generated quick nimble spec

doozMen:
The spec that it generates is meant to copy paste into a spec you actually implement. It is meant to get you started. Not to be a complete test set.


ilyapuchka:
Thanks @doozMen for contribution!
I'm not sure if we really need things like SwiftFormat in the project. In tests we don't care about spaces, they are effectively removed from generated and expected code.
It also adds too many changes unrelated to the template that you are submitting. I think it will make sense to remove it from the PR and discuss its need separately.

krzysztofzablocki:
yeah I think we should skip `SwiftFormat`, we don't need more dependencies as we try to make this work on Linux etc

doozMen:
Ok will remove that.

doozMen:
I will close this pull request and make a new one when ready

PR #563, opened 2018-03-25
Added ability to skip mocking of methods and variables

ChrisAU:
Added ability to skip mocking of methods and variables which can be helpful when extensions call existing methods.

Example:
```
protocol Dispatcher: AutoMockable {
    func dispatch(action: ActionType)
}

extension Dispatcher {
    //sourcery:skipMock
    func dispatch(actions: [ActionType]) {
        actions.forEach(dispatch)
    }
}
```

Since `dispatch(actions)` calls `dispatch(action)` we don't want to mock the former method, we want to test that it instead calls the latter method multiple times instead.

i.e.:
```
final class DispatcherSpec: QuickSpec {
    override func spec() {
        describe("Dispatcher") {
            var sut: DispatcherMock!
            
            beforeEach {
                sut = DispatcherMock()
            }
            
            context("dispatches") {
                it("single action") {
                    sut.dispatch(ActionTypeMock())
                    expect(sut.dispatchCalled).to(beTrue())
                    expect(sut.dispatchCallsCount).to(equal(1))
                }
                
                it("multiple actions") {
                    sut.dispatch([ActionTypeMock(), ActionTypeMock()])
                    expect(sut.dispatchCalled).to(beTrue())
                    expect(sut.dispatchCallsCount).to(equal(2))
                }
            }
        }
    }
}
```

### DIFF
--- a/Templates/Templates/AutoMockable.stencil
+++ b/Templates/Templates/AutoMockable.stencil
@@ -103,11 +103,11 @@ import AppKit
 
 {% for type in types.protocols where type.based.AutoMockable or type|annotated:"AutoMockable" %}{% if type.name != "AutoMockable" %}
 class {{ type.name }}Mock: {{ type.name }} {
-{% for variable in type.allVariables|!definedInExtension %}
+{% for variable in type.allVariables|!definedInExtension|!annotated:"skipMock" %}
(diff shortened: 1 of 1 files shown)


ilyapuchka:
Thanks for contribution @ChrisAU . `allMethods|!definedInExtension` should filter out methods which are defined in extension, so there is no need for `skipMock` annotation on such method. You may want to update your example to use annotation on the method defined in type declaration.

PR #593, opened 2018-04-09
Revert: fix: use process environment when running commands (#495)

### LINKED ISSUE #584: DispatchOverlayShims.h issue with Sourcery 0.11.0
> ❌  error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/shims/DispatchOverlayShims.h:185:20: error: implicit conversion of C pointer type 'dispatch_data_t _Nonnull' (aka 'struct dispatch_data_s *') to Objective-C pointer type '__swift_shims_dispatch_data_t _Nonnull' (aka 'id') requires a bridged cast

I've tried with both Xcode 9.2 and Xcode 9.3. Reverting to Sourcery 0.10.1 fixes the issue.

ilyapuchka:
Resolves #584 
This change seems to break swift templates. @krzysztofzablocki do you remember what was the reason why you changed that (#495)

### DIFF
--- a/SourcerySwift/Sources/SwiftTemplate.swift
+++ b/SourcerySwift/Sources/SwiftTemplate.swift
@@ -273,7 +273,7 @@ private extension String {
 }
 
 private extension Process {
-    static func runCommand(path: String, arguments: [String], environment: [String: String] = ProcessInfo.processInfo.environment) throws -> ProcessResult {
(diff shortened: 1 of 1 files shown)


krzysztofzablocki:
so this is a fix made by someone at Apple that I can't name publicly 😹, basically its for people that don't have Xcode installed in standard location and it seems that the issue is that the environment is not inherited in the subprocess calls

Issue #594, opened 2018-04-10
Mocking method with optional return types


SergeyKuryanov:
Sourcery: 0.10.0 - 0.11.1
Swift: 4.0, 4.1

When using `AutoMockable` protocol to generate mock for methods with optional return value:
```swift
protocol BasicProtocol: AutoMockable {
    func loadConfiguration() -> String?
}
```

Sourcery generates next mock:
```swift
class BasicProtocolMock: BasicProtocol {

    //MARK: - loadConfiguration
    var loadConfigurationCallsCount = 0
    var loadConfigurationCalled: Bool {
        return loadConfigurationCallsCount > 0
    }
    var loadConfigurationReturnValue: String?!
    var loadConfigurationClosure: (() -> String?)?

    func loadConfiguration() -> String? {
        loadConfigurationCallsCount += 1
        return loadConfigurationClosure.map({ $0() }) ?? loadConfigurationReturnValue
    }
}
```

Notice force unwrapping for `loadConfigurationReturnValue` value. This produced `Unexpectedly found nil while unwrapping an Optional value` error if we want to use mocked method with no value set:

```swift
let mock = BasicProtocolMock()
mock.loadConfiguration() // Error
```

This makes impossible to mock method, which returns `nil`.

ilyapuchka:
Resolved with #595

PR #595, opened 2018-04-10
Fix for AutoMockable template to handle methods optional return values

SergeyKuryanov:
This address issue: #594 

Template updated to add force unwrap for non-optional return values only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,10 @@
 
 ## Master
 
-### New Features
-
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ## Master
 
+### New Features
```
ilyapuchka: please put it into bug fix section

ilyapuchka:
👍

PR #599, opened 2018-04-13
[WIP] AutoCodable template

ilyapuchka:
- [x] Inline Coding Keys so that user only defines keys which do not match property names
- [ ] Support different key decoding strategies, not based on standard library strategy, but instead used by Sourcery to generate values for keys (compile-time vs run-time)
- [ ] Generate constructors in extensions (inlined or in separate file), so that structs still have member-wise initialisers
- [ ] Support custom coding methods in form of `decode<property name>`/`encode<property name>` instead of using annotations
- [ ] helper methods for safely decoding arrays, formatted dates etc.


Templates/Tests/Context/AutoCodable.swift
```diff
@@ -0,0 +1,29 @@
...
+    let int: Int
+
+    enum CodingKeys: String, CodingKey {
+        case int = "integer"
+
+// sourcery:inline:auto:CustomKeyDecodableStruct.CodingKeys.AutoCodable
+        case string
+        case bool
+// sourcery:end
+}
```
ilyapuchka: This should be fixed in #600

Templates/Tests/Context/AutoCodable.swift
```diff
@@ -0,0 +1,29 @@
...
+
+struct CustomKeyDecodableStruct: AutoDecodable {
+    let string: String
+    let bool: Bool
+    let int: Int
+
+    enum CodingKeys: String, CodingKey {
+        case int = "integer"
+
+// sourcery:inline:auto:CustomKeyDecodableStruct.CodingKeys.AutoCodable
```
ilyapuchka: #600 should keep indentation used in template

guides/Codable.md
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    func encodingContainer(_ encoder: Encoder) -> KeyedEncodingContainer<CodingKeys> {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        return container.nestedContainer(keyedBy: CodingKeys.self, forKey: .nested)
+    }
+}
+```
+
+  - method to encode a property. This is usefull when you need to manually encode a property:
```
Liquidsoul: Typo here: `useful`.

guides/Codable.md
```diff
@@ -0,0 +1,145 @@
...
+       try container.encode(String(myProperty), forKey: .myProperty)
+	}
+}
+```
+
+These methods may throw or not. 
+
+If you need to manually encode computed property and you defined custom encoding method for it, template will generate a coding key for it too, so you don't have to define it manually (though you may still need to define it if it needs custom raw value).
+
+  - method to encode any additional values. This is usefull when you need to encode computed properties or constant values:
```
Liquidsoul: Typo here: `useful`.

guides/Codable.md
```diff
@@ -0,0 +1,145 @@
...
+struct MyStruct: AutoDecodable {
+	let myProperty: Int
+
+	func encodeMyProperty(to container: inout KeyedEncodingContainer<CodingKeys>) {
+		try? container.decode(String(myProperty), forKey: .myProperty)
+	}
+	//or 
+	func encodeMyProperty(to encoder: Encoder) throws {
+		var container = encoder.container(keyedBy: CodingKeys.self)
+       try container.encode(String(myProperty), forKey: .myProperty)
```
Liquidsoul: Indentation seems off in this sample.

guides/Codable.md
```diff
@@ -0,0 +1,145 @@
...
+struct MyStruct: AutoDecodable {
+	let myProperty: Int
+
+	func encodeMyProperty(to container: inout KeyedEncodingContainer<CodingKeys>) {
+		try? container.decode(String(myProperty), forKey: .myProperty)
+	}
+	//or 
+	func encodeMyProperty(to encoder: Encoder) throws {
+		var container = encoder.container(keyedBy: CodingKeys.self)
+       		try container.encode(String(myProperty), forKey: .myProperty)
```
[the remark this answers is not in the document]
ilyapuchka: I blame github markdown renderer for that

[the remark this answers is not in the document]
Liquidsoul: It seems there are spaces at the beginning of this line, not just tabs.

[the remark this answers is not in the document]
ilyapuchka: should be fixed now

RomanTysiachnik:
Amazing!

PR #602, opened 2018-04-16
Insert auto-inlined code line before body end

### LINKED ISSUE #600: Auto inline annotations do not preserve code indentation
When inserted in code auto inline annotations do not properly indent `}` of containing type

ilyapuchka:
Resolves #600

### DIFF
--- a/SourceryTests/SourcerySpec.swift
+++ b/SourceryTests/SourcerySpec.swift
@@ -235,6 +235,54 @@ class SourcerySpecTests: QuickSpec {
                         expect(result).to(equal(expectedResult))
                     }
 
+                    it("insert generated code line before the end of type body") {
+                        update(code: """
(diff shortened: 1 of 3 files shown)


SourceryTests/SourcerySpec.swift
```diff
@@ -237,40 +237,60 @@ class SourcerySpecTests: QuickSpec {
 
                     it("insert generated code in multiple types") {
                         update(code: """
-                            class Foo {}
```
Liquidsoul: I would have let this test as it is, exposing the insertion of inline code, and would have just added a "Insert auto-inlined code line before body end" test to ensure that the existing insertion code still works.

ilyapuchka: done

PR #612, opened 2018-05-07
Cache base path configurable via yaml file

JimRoepcke:
I have added support for a `cacheBasePath` key to the yaml file. This path will be used in place of `~/Library/Caches` (which is still computed using `NSSearchPathForDirectoriesInDomains(.cachesDirectory, .userDomainMask, true)`) when computing the `cachesPath` for the `loadOrParse` function.

I have extracted the logic for calculating the default so that the default value can be placed in the `Configuration` when there is no yaml file.

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,3 +1,4 @@
 ---
 BUNDLE_DISABLE_SHARED_GEMS: "true"
+BUNDLE_PATH: "Vendor/bundle"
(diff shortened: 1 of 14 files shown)


ilyapuchka:
Its not needed to be committed, it is an intermediate file. And we'll update the docs with new release, if i have time in the evening i'll make it.

Ilya Puchka

Отправлено с iPhone

PR #616, opened 2018-05-14
Fixed AutoEquatable access level for == func

albsala:
AutoEquatable is currently generating == func's with the different access levels from public (`type.accessLevel`) which doesn't match the requirement of Equatable protocol:
```
public static func == (lhs: Self, rhs: Self) -> Bool
```
and is generating building errors:
```
Method '==' must be declared public because it matches a requirement in public protocol 'Equatable'
```
This PR replace this `type.accessLevel` with `public`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ---
 
+## Master
+
+### Bug fixes
+
+- Fixed AutoEquatable access level for `==` func
+
 ## 0.13.1
 
 ### New Features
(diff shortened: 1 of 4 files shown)


shrnjad:
Hard-coding the access level does not work for me as I get the following compile error:
`Function cannot be declared public because its parameter uses an internal type`

I think the problem is that my auto-equatable origin classes have internal access level. When I change back to the previous version of the template, it compiles fine:
`{{type.accessLevel}} func == (lhs: {{ type.name }}, rhs: {{ type.name }}) -> Bool {`

PR #617, opened 2018-05-14
AutoEquatable with annotations

albsala:
As I already did with AutoMockable (#518), I think it's a good idea to make AutoEquatable usable with Annotations.

This is a backward compatible change on AutoEquatable template for making it usable with annotations in addition to the current way (extending a protocol).


ilyapuchka:
@albsala can you please update guide so that it mentions this annotation?

albsala:
Hi @ilyapuchka, I've updated both AutoEquatable and AutoMockable since I didn't do it for the latter in #518. Thank you for your feedback.

PR #634, opened 2018-06-13
Fix extraction of inline annotations in comments like /** sourcery: skipDescription */

4brunu:
Hi, I have a project that was using an old version of Sourcery, 0.10.1.

In that version, Sourcery was able to extract annotations from comments with the following format:
```
/** sourcery: skipDescription */
```

Today I updated Sourcery and it started failing parsing this type of comments.

So I fixed and added a test case to it.

EDIT:Supporting this type of comment is important to me, because I use a tool that generates code with this kind of comments, that contains annotations for Sourcery.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 - Fixed AutoEquatable access level for `==` func
 - Fixed path of generated files when linked to Xcode project
+- Fixed  extraction of inline annotations in comments like /** sourcery: skipDescription */
 
 ## 0.13.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 - Fixed AutoEquatable access level for `==` func
 - Fixed path of generated files when linked to Xcode project
-- Fixed  extraction of inline annotations in comments like /** sourcery: skipDescription */
+- Fixed extraction of inline annotations in multi line comments with documentation style
 
 ## 0.13.1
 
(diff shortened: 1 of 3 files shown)


ilyapuchka:
Xcode will render this as documentation comments though...

4brunu:
Yes, but because this is generated code, it’s my only option...😕

4brunu:
Another thing that I would like to add to Sourcery is the ability to extract inline annotations in multi line comments like this
```
/** 
* Some comment
* sourcery: skipDescription 
* sourcery: skipEquality 
*/
```

I have a [branch](https://github.com/4brunu/Sourcery/tree/multi_line_comments_inline_annotation_extraction) with those changes already implemented, but I was not sure if it was a good idea to mix this two changes.

What do you think, should we address those things together or separately?
Thanks

ilyapuchka:
I personally think it will be a bad practice to mix documentation with annotations, so we shouldn't even support `///` comments as they are also for documentation purposes.

4brunu:
I understand your point, and I honestly can't argue with that.

I'm only using inline annotations in documentation comments, because I have no other option, since it's generated code.

If this PR could be accepted, it would be great for me, so I can continue to use Sourcery, if don't I understand.

4brunu:
I just added support for annotation in multi line comments and one more test.

krzysztofzablocki:
@ilyapuchka are you ok with the changes?

PR #654, opened 2018-07-30
Concat file content

### LINKED ISSUE #648: Feature Request: Append on "Per File" code generation
At the end of the [Writing Templates](https://cdn.rawgit.com/krzysztofzablocki/Sourcery/master/docs/writing-templates.html) section in the docs, there's a sourcery annotation (sourcery:file) that lets you specify a different output file to generate  to for each type instead of writing the output for all types that apply to a template into one giant file. This has been especially useful because I use the incremental single file compilation mode, and having a giant generated file that all my classes depend on means Xcode has to recompile all of my files that reference that giant file every time it changes.

I ran into a problem lately where I'd like to have two different templates write to the same file, so I can keep my number of files low, but wasn't able to do it because the second template's output overwrites the first one's. Would it be possible to make it so when two or more templates try to write to the same file in the same invocation of sourcery, the second one's output would get appended to the first instead of overwriting it?

ilyapuchka:
Resolves #648


ethanjdiamond:
Thanks ilya!

PR #655, opened 2018-07-31
Make AutoEquatable template react to origin access level

shrnjad:
Making an internal class auto-equatable resulted in compile errors before, because the access level public was hard-coded to the template and access levels would not match.


shrnjad:
Sorry, I did not realize, I was basicall reverting commit 4cd02ec832a96a252f2098d371856af62aeb257e.

But however, using the hard-coded public access level does not work in all cases either. I have also commented no the issue.

PR #663, opened 2018-08-13
Allow Swift templates (`.swifttemplate`) to define code blocks that have to be at the top level in the compiled `.swift` file.

ivanmisuno:
Currently, all code declarations from the processed `.swifttemplate` file are nested in the body of the function in the compiled `.swift` file like this:
```
extension TemplateContext {
    func generate() throws {
        \(contents)
    }
}
```

Such organization of the code prevents usage of some of the Swift language constructs that have to be at the top level of the Swift program, for example, protocol and type extensions.
The compiler would give "declaration is only valid at file scope" error if an extension was used in the template, for example.

However, when writing complex `.swifttemplate` templates it is sometimes desirable to be able to use them.

This commit adds support for a new escaping pattern `<%^` when parsing `.swifttemplate` files.
The code extracted from such blocks will be placed at the top level of the intermediate compiled `.swift` file, so that defining extensions becomes possible.

As an example, `Decorator.swifttemplate` is updated to define following extensions on `SourceryRuntime.Method` type instead of free functions:
```
fileprivate extension SourceryRuntime.Method {
    /// Constructs full method declaration string
    var methodDeclaration: String { /*...*/ }
    /// Constructs method call string passing in parameters with their local names
    var methodCall: String { /* ... */ )
}
```


ivanmisuno:
> One concern about this implementation - I think usage of <%^ %> tag should be limited to top level of templte code and should rais error if used anywhere else.

I also thought about limiting how this block could be used, but did not come up with any practically helpful ideas or easy enough solutions.

E.g., this will work fine, despite being poorly organized:
```
<% for type in types.all { -%>
<%^
let extensionPrefix = "prefix"
extension SourceryRuntime.Type {
    var extensionProperty: String { return "\(extensionPrefix)\(name.uppercased)" }
}
-%>
// MARK: - <%=type.extensionProperty%>
<% } -%>
```

There are many ways, of course, to break the top-level template block, e.g. calling the throwing function etc:
```
<%^
let extensionPrefix = try throwingFunction()
-%>
```

Practically, I don't see an easy solution here without actually parsing and interpreting the resulting Swift program - but that's what the compiler would do anyway.

On the other hand, I just wanted to give (myself) a little bit more flexibility writing templates, and I thought that the flexibility in this case can come with the additional cost of having to be more careful.

ilyapuchka:
I think just checking that <%^ is used before any other tags should be enough to achieve that. We can be even more strict and only allow it as a first entry in the template, requiring all the static text go after it.

ivanmisuno:
> I think just checking that <%^ is used before any other tags should be enough to achieve that. We can be even more strict and only allow it as a first entry in the template, requiring all the static text go after it.

That would be too restricting IMO.
`<%- include("") -%>`, for example, might want to have their own `<%^%>` blocks, in addition to having static text and other escaping blocks (as well as including other files) -- and these usually come first in templates.

That's how I'm trying to structure my own templates currently, more or less:

```
<%- include("_commonHeaders") -%>
<%- include("_mockGenerator") -%>
<%= MockGenerator(types: types).generate() -%>
```

I'm actually going away from the "template file" concept here, which allows to put some static text in between some processing instructions and then have a template engine parse and interpret it, and building my own Swift code generation engine (albeit a very simple one) on top of the framework that Sourcery provides.

There are definitely several paths forward at this moment, from making the `<%^%>` block more safe to use (but without restricting the flexibility), to re-thinking the concept behind swift templates. Maybe open an issue to collect the requirements for the "next version" of swift templates?

krzysztofzablocki:
I agree with @ivanmisuno on that being a bit too restrictive

anandabits:
I have been doing something similar to @ivanmisuno.  In fact, I have been naming the files I include with the `.swift` extension for improved editor support.  

The `<%^` fix in this PR would be a nice enhancement, but it is still not a complete solution as access control will not work correctly.  In the example given, `_commonHeaders` and `_mockGenerator` will be able to see each other's `fileprivate` declarations and be able to extend each other's types to gain access to `private` declarations.

The longer-term solution I would prefer is to have something along the lines of `includeFile` which _does not_ inline the contents of the file, but instead adds it to the list of source files given to `swiftc` when compiling the Swift template.  Ideally it would be possible to set up an Xcode target to edit (with full code completion, quick help, etc), build, and test sophisticated Swift templates.  This would significantly improve the developer experience of authoring non-trivial templates (IMO).

ilyapuchka:
@anandabits that would be good, but with current syntax of swift templates code completion will not be possible any way, to have it swift template should be a valid swift program. This is achievable but will make them less "template" and more a "plugin".

anandabits:
@ilyapuchka it would be interesting to see how something like that could come together someday, however it seems like something that might take a while to pull off.  

What I had in mind is that `includeFile` would expect the files to be valid Swift code, not template code.  This would allow the file paths to be passed directly to `swiftc`.  Templates using this feature would usually have a `.swifttemplate` that only uses `includeFile` and then calls a driver implemented in the `.swift` files (as in the example @ivanmisuno showed).

This would match what I am already doing.  The first line of my included `.swift` files is `// <%_` and the last line is `// %>`.  These are already valid Swift files, but they depend on symbols from SourceryRuntime and setting up a test driver with a live SourceryRuntime isn't something I have explored.

When this PR gets merged it will be possible to use protocols and extensions.  That will allow templates to be written as normal Swift programs while also isolating the dependency on SourceryRuntime from the logic of the template.  Doing this will make it much easier to set up an edit / build / test environment that does not depend on SourceryRuntime by using protocols that can be conformed to by other types in an edit / build / test target in Xcode (while excluding the `.swift` files that declare conformances for SourceryRuntime types).

I just looked at what would be involved in adding `includeFile`.  I could be missing something, but it seems like it would be pretty straightforward to implement and the improvement in developer experience it would enable is significant (IMO).  Swift files included by Swift templates would support the _entire_ language.

ivanmisuno:
@anandabits I like the idea of passing the paths to .swift files to the compiler. That's a more interesting approach to explore, compared to just inlining the code. That would solve code separation concerns, as well as bring other mentioned benefits, like the ability to set up a proper dev environment etc.
Would be great to have `SourceryRuntime` as a separate Pod to help with setting up the dev environment indeed.

ilyapuchka:
Is this still relevant since #666 was merged?

ivanmisuno:
Sorry, was traveling and couldn't answer faster. Yes, #666 looks great and makes this PR unnecessary.

PR #666, opened 2018-08-23
Add support for including `.swift` files in Swift templates

anandabits:
There are a number of Swift language features that are not currently supported by Swift templates.  Most notable are extensions and protocols which are not available because Swift template code does not currently live at the top level scope of the Swift file that is compiled.  The use of files and `private`  `fileprivate` access control to structure the code in larger templates is also not available.  Lifting these restrictions will significantly improve support for writing sophisticated templates using the full power of Swift.

This PR adds support for the full scope of Swift language features by allowing Swift templates to include entire `.swift` files.  The paths to these files are passed to the compiler along with the path to `main.swift`.  Usage is very simple:

In the `.swifttemplate` file:
```
<%- includeFile("Template.swift") -%><%= Generator.generate(for: types) %>
```

In `Template.swift`:
```swift
import SourceryRuntime

enum Generator {
    static func generate(for types: Types) -> String {
        // ...
    }
}

private extension Typed {
    // ...
}
```

Note: this PR overlaps significantly with the changes proposed by https://github.com/krzysztofzablocki/Sourcery/pull/663.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### New Features
 
+- You can now include entire Swift files in Swift templates
 - You can now use AutoEquatable with annotations
 - Content from multiple file annotations will now be concatenated instead of writing only the latest generated content.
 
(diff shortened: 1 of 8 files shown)


ilyapuchka:
👏 that looks great @anandabits 
@krzysztofzablocki mind taking a look?

Issue #639, opened 2018-06-25
Warning on parseEnumCase when using Swift 4.2


marcelofabri:
After enabling use with Swift 4.2/Xcode 10 (#638), I started getting this warning when running sourcery:

```
File.swift:parseEnumCase: Unable to extract enum body from ["key.nameoffset": 924, "key.accessibility": "source.lang.swift.accessibility.private", "key.length": 13, "key.name": "empty(_:)", "key.kind": "source.lang.swift.decl.enumelement", "key.namelength": 5, "key.offset": 924]
```

I haven't look at the code, but SwiftLint faced a similar issue because `key.name` now contains the parameters (https://github.com/realm/SwiftLint/issues/2231), so it might be related.

fassko:
I'm having the same issue with Xcode10/Swift4.2

PR #668, opened 2018-08-24
fix: get rid of (:_) for enum cases

krzysztofzablocki:
Can a fix for #639 be that simple?


ilyapuchka:
Is it because sourcekit output format changed?

krzysztofzablocki:
yeah they add `(:_)` to name

PR #669, opened 2018-08-30
Fix non existing revision dependency

mickamy:


ilyapuchka:
@mickamy sorry, what's the reason to update all dependencies?

PR #677, opened 2018-09-13
Parsing JSON string args & annotations into arrays and dictionaries

Landsi:
This PR enables you to pass a json string as an argument to sourcery in terminal or as an annotation in code and mak it 

We have a use-case of generating a type that mirrors most properties of another type, with an optional property name mapping, should it be needed. The changes in this PR enable me to do this:

In code:
```
// sourcery: propertyMapping='[{"from": "typeEnum", "to": "type"},{"from": "goalStatus", "to": "status"}]'
extension RTDatabaseGoal: AutoEntity { ... }
```

In the template:
```
// ...
{% for type in types.implementing.AutoEntity %}
    {% for variable in type.variables %}
        public var {% for mapping in type.annotations.propertyMapping where mapping.from == variable.name %}{{mapping.to}}{% empty %}{{variable.name}}{% endfor %}: {{variable.typeName}}{% if variable.isOptional %}?{% endif %}
    {% endfor %}
{% endfor %}
// ...
```

Helped me out in my case, so maybe you'll consider merging.

I'm rather new to Sourcery -> so if there might be a better way to achieve this, I'm open to suggestions.


Landsi:
If this is considered to be accepted, i should probably keep the project, Podfile and Xcode 10 changes out of this PR. Sorry about that.

krzysztofzablocki:
@Landsi if you were able to get Sourcery to build with Xcode 10, maybe you could make a separate PR to get that working ? a lot of people asked for that in #680 and it would be awesome if you could help out

Sourcery/Parsing/Utils/AnnotationsParser.swift
```diff
@@ -260,7 +260,21 @@ internal struct AnnotationsParser {
...
                         value = value.trimmingCharacters(in: .whitespaces)
                     }
-                    append(key: name, value: value as NSString, to: &annotations)
+
+                    guard let data = (value as String).data(using: .utf8) else {
+                        append(key: name, value: value as NSString, to: &annotations)
+                        return
+                    }
+
+                    if let jsonArray = try? JSONSerialization.jsonObject(with: data, options: .allowFragments) as? [[String: Any]],
```
ilyapuchka: why not to parse it just once and then do type casting? This would be more performant and compact. I also wonder if conditional typecasting to NSArray and NSDictionary is really needed, aren't they bridged automatically?

```swift
guard let data = (value as String).data(using: .utf8),
   let json = try? JSONSerialization.jsonObject(with: data, options: .allowFragments) else {
  append(key: name, value: value as NSString, to: &annotations)
  return
}
if let array = json as? [[String: Any]] {
  append(key: name, value: json as! NSArray, to: &annotations)
} else if let dict = json as? [String: Any] {
  append(key: name, value: json as! NSDictionary, to: &annotations)
} else {
  append(key: name, value: value as NSString, to: &annotations)
}
```

Landsi: true. much better.

Issue #678, opened 2018-09-13
Error in Xcode 10


hartbit:
I'm getting an error running Sourcery since I passed to Xcode 10 and updated Sourcery to 0.14.0:

```
No config file provided or it does not exist. Using command line arguments.
Scanning sources...
Found 676 types.
Loading templates...
Loaded 5 templates.
Generating code...
error: Error Domain=NSOSStatusErrorDomain Code=-1 "kCFStreamErrorHTTPParseFailure / kCFSocketError / kCFStreamErrorDomainCustom / kCSIdentityUnknownAuthorityErr / qErr / telGenericError / dsNoExtsMacsBug / kMovieLoadStateError / cdevGenErr: Could not parse the request/response. /  / custom to the kind of stream in question  /  / queue element not found during deletion /  / not a SysErr, just a placeholder  /  / General error; gray cdev w/o alert" UserInfo={terminationReason=__ObjC.Process.TerminationReason, error=dyld: Symbol not found: __T0BOWV
  Referenced from: /Users/david/Library/Caches/Sourcery/AutoCodable.swifttemplate/../SourceryRuntime.framework/Versions/A/SourceryRuntime
  Expected in: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx/libswiftCore.dylib
 in /Users/david/Library/Caches/Sourcery/AutoCodable.swifttemplate/../SourceryRuntime.framework/Versions/A/SourceryRuntime
}
```

bscothern:
Same here. I've deleted my `~/Library/Caches/Sourcery` and the error changed to this:

```
error: /var/folders/x9/bvxmg8bx42jd0hl20q0l6l1c0000gn/T/SwiftTemplate.build/main.swift:2:8: error: module compiled with Swift 4.1.2 cannot be imported by the Swift 4.2 compiler: /var/folders/x9/bvxmg8bx42jd0hl20q0l6l1c0000gn/T/SourceryRuntime.framework/Modules/SourceryRuntime.swiftmodule/x86_64.swiftmodule
import SourceryRuntime

```
I have also just tried building from source and get the error: "Redefinition of module 'CommonCrypto'".

ilyapuchka:
swift templates require to be built with the same version of swift as Sourcery. There was no release against swift 4.2 yet

hartbit:
[...]

Issue #672, opened 2018-09-07
Sourcery with Swift 4.2, Cocoapods, Xcode 10b6


daveanderson:
We recently modified our app (which has an internal .framework) to use Sourcery for the framework as well. 

The build phase script is doing a 

```
"${PODS_ROOT}"/Sourcery/bin/sourcery --config InternalKit/.sourcery.yml
```

And we're getting the error

```
Using configuration file at 'InternalKit/.sourcery.yml'
Scanning sources...
Found 266 types.
Loading templates...
Loaded 4 templates.
Generating code...
error: /var/folders/f8/jv7mx3552r58j521g47x_y400000gn/T/SwiftTemplate.build/main.swift:2:8: error: module compiled with Swift 4.1.2 cannot be imported by the Swift 4.2 compiler: /var/folders/f8/jv7mx3552r58j521g47x_y400000gn/T/SourceryRuntime.framework/Modules/SourceryRuntime.swiftmodule/x86_64.swiftmodule
import SourceryRuntime
```

xcode-select indicates

```
$ xcode-select -p
/Applications/Xcode-10.app/Contents/Developer
```

with cocoa pods

```
$ pod --version
1.6.0.beta.1
```
and Sourcery

```
$ Pods/Sourcery/bin/sourcery --version 
0.14.0
```

Is this a configuration defect on our end or have we hit some sort of a compatibility issue?

ilyapuchka:
As error says: `module compiled with Swift 4.1.2 cannot be imported by the Swift 4.2 compiler`
You can use Sourcery with Swift templates only with Swift 4.1, there was no release for Swift 4.2

daveanderson:
Thank you for the confirmation @ilyapuchka

PR #681, opened 2018-09-18
Use SPM instead of swiftc to build swift templates

Liquidsoul:
**Objectives**

- remove compatibility issues with different Swift versions than the one used to package a Sourcery release (like these ones #678 #672)
- introduce more extensibility to the swift template build process. We could imagine allowing the use of external packages in the templates in the future

**What this does**

This drops the use of `swiftc` and the (problematic) linking step to the embedded SourceryRuntime framework.

Because we cannot use the embedded framework, this introduces the "packaging" of the SourceryRuntime source code into the Sourcery binary. To do that, there is a new build phase in the Xcode project that takes all the SourceryRuntime files and write them into a swift file compiled by Sourcery in `SourcerySwift` target (`SourceryRuntime.content.generated.swift`). This code will then be used during the Swift template build process to write down all the files back to the disk where they'll be used to build the template.

The build process of a Swift template looks like this:
1. create the package folder hierarchy
2. synchronize the SourceryRuntime code into the package's `Sources` using the
embedded code in the Sourcery binary and `FolderSynchronizer`
3. write the manifest file `Package.swift`
4. delete `Sources/SwiftTemplate` folder if it exists to remove any file from
the previously built template
5. write the `main.swift` file in `Sources/SwiftTemplate`
6. add any included files in `Sources/SwiftTemplate`
7. build the template's binary using SPM

The hierarchy before building:
```
.
├── Package.swift
└── Sources
    ├── SourceryRuntime
    │   ├── ... (SourceryRuntime files)
    └── SwiftTemplate
        ├── main.swift
        └── ... (included files if any)
```

**Drawbacks**
This introduces some performance issues for the first swift template you'll build. This is because SPM needs to build the SourceryRuntime package a first time. Then, the SPM build cache will do its job until a new version of Sourcery is used or the temporary folder is deleted.
**_I did not run extensive benchmarks to see what the real impact of this is. I would love to have some help on this_** 😊

**SourceryRuntime packaging rationale**
Maybe you are wondering why I've introduced this SourceryRuntime packaging step, so I'll try here to expose the reasons why I did that.

Let's see what solutions I have thought of:

_1. Reference this repository as the package source for SourceryRuntime_

This is the first thing I tried. There are several problems with this:
- performance: SPM would have to clone the whole repository and just reference a single package in it. SPM may even need to build the whole package with a lot of things we do not want.
- offline availability: using a reference to a remote repository makes them hard to use if you just want to work on your code without relying on your internet connection. My point: if you have downloaded Sourcery, you should not be required to download anything else to run it.
- versions synchronisation: we need to make sure that the remote code is synchronised with the Sourcery version used. The main problem is more for the maintainers here because when you need to work on this part of the code, you would have to change the package reference to a local one.

_2. Extract SourceryRuntime into its own repository/package_

This have the same issues as in `1.` but even introduces the hassle of having to synchronise one more dependency for the maintainers. I think we do not want to introduce more maintenance work that is needed here.

_3. Package the code inside Sourcery's binary_

This has the advantages of ensuring that there are no difference between the binary and the code of SourceryRuntime used in Swift template.
The problem with that is this introduces a packaging step of the SourceryRuntime code which needs to be maintained. And also this will increase the size of the binary.

I thought that the option `3.` was the best tradeoff of these options. However, if you think of any better solution, please share it with me!

**Sidenotes**
The process does not remove the temporary folder anymore so that the SourceryRuntime code is not rebuilt for every template. In the future, we'll have to keep an eye on that because there may be some remaining files from one run to the next. Right now this should not be a problem because all template
related files are removed at the end of `build()`.


Scripts/PackageContent.sh
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env bash
```
ilyapuchka: wooow, what is that and what does it do? I personally won't be able to support such script as I have no idea what and how it does...

ilyapuchka:
@krzysztofzablocki @Liquidsoul TBH I'm not a fan of this approach, especially `PackageContent.sh` 😞 I think we can explore other options. Maybe we can provide a custom script to install Sourcery that will also automatically checkout and maybe even build SourceryRuntime in some cache folder and then use it to link in template files. I'm not familiar with latest state of SMP though and if it is something possible or not without performance implications. But I don't think we should proceed with this approach - swift templates are already complicated and they become even more complicated with that 😞

Liquidsoul:
To solve the script issue, I have just pushed a replacement of this bash script with a swift version of it that should be easier to maintain by any Swift developer.

`PackageContent.sh` (and now `package_content`) is really just a simple script that just reads and prints back Swift files into one Swift file with characters escaped.

I have written the bash script with as much documentation and helpers as possible. That's why it can be intimidating. But it just "globs" the files in a folder and write then back into one file with some Swift tokens, it just boils down to the `main` function.

My approach here is to avoid adding overhead to the user who uses swift templates. Adding an additional build step during install or before running sourcery is not something that I see as acceptable.

As I said, it is not possible to fetch "just" SourceryRuntime as it is right now. And isolating it into a separate repository would be too much of a hassle for the rest of the project. So using SPM to fetch the SourceryRuntime does not seem possible.

But even if we _did_ have a solution to fetch `SourceryRuntime` from a remote repository, this would introduce a strong dependency to the latest released version of Sourcery and would make development harder when introducing changes to `SourceryRuntime`.

The _only_ solution we have would be to keep the system as it is and wait for ABI stability.
But this would mean that compatibility issues will be still reported in the meantime, and this may also prevent any wider adoption of swift templates in the near future.

krzysztofzablocki:
this seems much easier to maintain with Swift script, thoughts @ilyapuchka ?

ilyapuchka:
swift helps, but I'm still in doubt about the whole approach

krzysztofzablocki:
I agree with @Liquidsoul that this is nicer solution than having to require people to perform extra manual steps to use swift templates, but I'm not familiar with SPM.

I wonder what other swift template users thing about this, @anandabits @ivanmisuno  what do you guys think ?

anandabits:
I like the idea of using SPM to build Swift templates.  The potential it gives us to allow Swift templates to depend on packages would be awesome!  I would need more time to think through the tradeoffs to form an opinion about implementation strategy.

My team currently commits Sourcery to our repository along with the templates we use.  All developers need to do is clone the repo and build.  Sourcery is mostly transparent in their workflow - the only noticeable differences are occasional issues with (incorrect) annotations as well as compiler errors after editing until a new build is run.  

The suggestion by @ilyapuchka would introduce an "install Sourcery" step into the workflow for new developers as well as the whole team every time we upgrade Sourcery.  This isn't a dealbreaker but I think it would be better to avoid it if possible.  We should certainly take this cost for users into account when evaluating the alternatives.  

My opinion is that it would be much better if Sourcery just detected a new or upgraded version is being used automatically and did whatever necessary to build and link the correct runtime.  This should absolutely support different versions of Sourcery used by different projects developed on the same computer.  It would be cool if it could work from a cloned repo offline but I don't think that's essential as long as a cache is retained after a new version of Sourcery is used.  I'm not concerned about this introducing additional build time for new or upgraded installs.

krzysztofzablocki:
@Liquidsoul will this PR change work well with different sourcery versions on same machine / switching branches in the same project? if so I think we don't really have a better alternative for now and I'd be up for accepting this

krzysztofzablocki:
Can you add that? Sounds like a good improvement and would avoid future issues that are going to happen otherwise

Liquidsoul:
@krzysztofzablocki I've added a commit that injects the version into `SwiftTemplate` and uses it to specify a build folder per Sourcery version.

PR #682, opened 2018-09-18
Fixes generation of codable for enums without associated value names.

kzaher:
This example failed to compile:

```
enum A: AutoCodable {
  case a(Int)
}
```

with unwrapping nil fatal error.

### DIFF
--- a/Templates/Templates/AutoCodable.swifttemplate
+++ b/Templates/Templates/AutoCodable.swifttemplate
@@ -87,7 +87,11 @@ func codingKeysFor(_ type: Type) -> (generated: [String], all: [String]) {
     } else if let enumType = type as? Enum {
         let casesKeys: [String]
         if enumType.hasAssociatedValues {
-            casesKeys = enumType.cases.map({ $0.name }) + enumType.cases.flatMap({ $0.associatedValues }).flatMap({ $0.localName })
+            casesKeys = enumType.cases.map({ $0.name }) + enumType.cases.flatMap({ enumCase in
+              enumCase.associatedValues.enumerated().map({ (index, associatedValue) in
+                associatedValue.localName ?? "\(enumCase.name)_\(index)"
+              })
+            })
         } else {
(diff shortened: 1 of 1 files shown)


krzysztofzablocki:
hey @kzaher would you mind adding changelog and update template tests to cover this use-case so we don't make regressions later on?

kzaher:
Hi @krzysztofzablocki ,

I can do that, but I couldn't find any tests for those templates. Could you point me where are these tests located?

ilyapuchka:
This was deliberately not supported by the template and is mentioned in the [docs](https://github.com/krzysztofzablocki/Sourcery/blob/master/guides/Codable.md#enums-with-assoicated-values). Using format `\(enumCase.name)_\($0)` is arbitrary and may quite likely not work for others.
If we want to support that in this template IMO using array for keyed by the name of the enum case will be a better option on is better alligned with format expected for cases with associated values. Then `case a(Int, ..., Int)` will encoded/decoded from/to 

```
{ 
  "a": [
    value1, 
    ..., 
    valueN 
  ] 
}
``` 

instead of proposed

```
{ 
  "a_0": ..., 
  "a_N": ...  
}
```

ilyapuchka:
This temple is designed primarily to handle general case, as you say, which is decoding/encoding in JSON format. If it does not suit your needs as it is you can adjust it to suit you but merging in any such change will affect other users of this template.

kzaher:
I've meant general case as in "handles all cases supported in the language without crashing with unwrapping nil exception".

> but merging in any such change will affect other users of this template.
This can't affect any user in negative way IMHO since every time somebody encountered this case he literally crashed with `unwrapping nil exception`.

But on the other hand it is your library, so if you want to crash with `unwrapping nil exception` it's your right to do so.

ilyapuchka:
The crash is unarguably a flaw in the template that should be fixed, so thanks for flagging it.

Issue #680, opened 2018-09-18
Release date for Swift 4.2 ?


Gruppio:
Hi, do you already have an idea about the release date for make it compatible with Swift 4.2 ? 
Thank you!

petaren:
Not a contributor. I tried getting it to compile and run with XCode 10. It seems like most/all of Sourcerys' dependencies have been updated, but I'm not able to compile due to other issues.

amichnia:
Same here, I had a lot of troubles with building in Xcode 10. From what I tried, these are some issues:
- StencilSwiftKit - array literals containing closed range literals (SwiftIdentifier.swift) is too complicated for new compiler
- libCommonCrypto - could be removed in general, CommonCrypto is now visible from swift without the need for custom module maps
- some dependencies fail with optimization turned on, -Onone seems to work for them, but not 100% sure. I don't recall particular one though, maybe xcproj
- SourceryRuntime - at least for me it failed due to group structure not matching folders structure

I still fail in CodeSign path though, will give it a better look, as I'm really looking forward to use it with XCode 10

krzysztofzablocki:
I'm completely swamped with commercial work, and I imagine a lot of our core team is in similar situation right now, so we could use some help here.

First step is to update dependencies that aren't upto date like SourceKitten, Stencil and StencilSwiftKit because when I did last update their respective dependencies were messed up, those things are resolved now I think so that would be the first thing someone needs to do. 

That should resolve some issues and allow us to figure out next steps

ilyapuchka:
0.15.0 with Xcode 10 and Swift 4.2 support is released

PR #683, opened 2018-09-20
Xcode 10 & Swift 4.2

Landsi:
This PR enables Xcode 10 support and paves the way for #680.

This will not build with Xcode 9.x any more, due to different handling of module maps in combination with cocoapods.

I ran a full `pod update` on this (yes i dared @krzysztofzablocki - nice plugin ;) )
I kept your scheme of specifying specific pod versions, although i believe it would be better to be a bit more lenient on the patch version at least, if not adopting semantic versioning altogether.

I updated the pods to already support swift 4.2 (i think, haven't really tried), but did not update the sourcery project itself.

2 tests failed after the update. i believe it's nothing major, but check out my changes in the specs before merging. now the tests succeed but my changes in the specs files might not be correct.

cheers, max


Landsi:
It seems like the dependency graph doesn't work for the swift package manager, because StencilSwiftKit is limiting some dependency versions for swift package manger, which the cocoapods do not.

will try to find a solution...

Podfile
```diff
@@ -26,15 +26,14 @@ target 'CodableContextTests' do
...
-  pod 'StencilSwiftKit', '2.4.0' 
-  pod 'Commander', '0.6.0'
-  pod 'PathKit', '0.8.0'
-  pod "xcproj", '4.2.0'
-  pod 'SourceKittenFramework', '0.21.1'
+  pod 'Stencil', '0.12.1'
+  pod 'StencilSwiftKit', '2.6.0'
+  pod 'Commander', '0.7.0'
+  pod 'PathKit', '0.9.2'
+  pod "xcproj", :git =>'git@github.com:tuist/xcodeproj.git', :tag => '4.3.1'
```
Liquidsoul: What is the reason you use a fork instead of `https://github.com/xcodeswift/xcproj.git` here? Because you're using the main repo in the [SPM manifest](https://github.com/krzysztofzablocki/Sourcery/pull/683/files#diff-37ca2dd15ca0f6b1b49e78db084ef5b9R20) 🤔

SourceryTests/Parsing/ComposerSpec.swift
```diff
@@ -50,8 +50,10 @@ class ParserComposerSpec: QuickSpec {
                         expect(barType.allVariables[0].definedInType).to(equal(barType))
                         expect(barType.allVariables[1].definedInType).to(equal(fooType))
                         expect(bazType.allVariables[0].definedInType).to(equal(bazType))
-                        expect(bazType.allVariables[1].definedInType).to(equal(fooType))
-                        expect(bazType.allVariables[2].definedInType).to(equal(barType))
+                        // TODO: seems like the inherited types are not always in the same order (any more)
+//                        expect(bazType.allVariables[1].definedInType).to(equal(fooType))
+//                        expect(bazType.allVariables[2].definedInType).to(equal(barType))
+                        expect(bazType.allVariables.map({$0.definedInType})).to(contain([fooType, barType]))
```
Liquidsoul: I suppose the order here is not critical. But maybe you could add another check on the `count` attribute to ensure that there are just theses types in `definedInType`?

SourceryTests/Generating/StencilTemplateSpec.swift
```diff
@@ -30,7 +30,9 @@ class StencilTemplateSpec: QuickSpec {
                 context("given array") {
                     it("doesnt modify the value") {
                         let result = generate("{% for key,value in type.MyClass.variables.2.annotations %}{{ value | toArray }}{% endfor %}")
-                        expect(result).to(equal("(\n    Hello,\n    beautiful,\n    World\n)"))
+                        // TODO: apparently something changed in the `toArray` filter?
+//                        expect(result).to(equal("(\n    Hello,\n    beautiful,\n    World\n)"))
+                        expect(result).to(equal("[Hello, beautiful, World]"))
```
Liquidsoul: 🤔 either the previous expectation was incorrect (which I doubt) or we have a major change of how loops on dictionaries are done in Stencil?

SourceryTests/Parsing/ComposerSpec.swift
ilyapuchka: @Liquidsoul https://github.com/stencilproject/Stencil/pull/240

Podfile
ilyapuchka: `xcodeswift/xcproj.git` redirects to `tuis/xcodeproj.git`

Landsi: it's a shame that they didn't release the 4.3.1 version properly with cocoapods..

PR #688, opened 2018-09-25
Add parsing for `lazy var`

### LINKED ISSUE #686: Proposal: Add support for knowing if a var is lazy
I think this can be very useful in many use cases where the usage of `lazy var` is used and may be nice to add it into a `Type`'s `attributes` or even with its own attribute `isLazy`. Maybe this can already be accomplished, if it does it should be given more visibility into the documentation.

Nice work and keep it up!

minuscorp:
Resolves #686 
This PR tries to add support for `lazy` attribute.

### DIFF
--- a/SourceryRuntime/Sources/Variable.swift
+++ b/SourceryRuntime/Sources/Variable.swift
@@ -53,7 +53,7 @@ public typealias SourceryVariable = Variable
         return attributes[Attribute.Identifier.final.name] != nil
     }
     
-    // Whether variable is lazy or not
+    /// Whether variable is lazy or not
     public var isLazy: Bool {
         return attributes[Attribute.Identifier.lazy.name] != nil
     }
--- a/SourceryRuntime/Sources/Variable.swift
+++ b/SourceryRuntime/Sources/Variable.swift
@@ -52,6 +52,11 @@ public typealias SourceryVariable = Variable
     public var isFinal: Bool {
         return attributes[Attribute.Identifier.final.name] != nil
     }
+    
+    // Whether variable is lazy or not
+    public var isLazy: Bool {
+        return attributes[Attribute.Identifier.lazy.name] != nil
+    }
 
     /// Reference to type name where the variable is defined,
     /// nil if defined outside of any `enum`, `struct`, `class` etc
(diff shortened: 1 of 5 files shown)


SourceryTests/Parsing/FileParser + VariableSpec.swift
```diff
@@ -115,6 +115,10 @@ class FileParserVariableSpec: QuickSpec {
                     expect(parse("var name: Int { didSet {} }")).to(equal(Variable(name: "name", typeName: TypeName("Int"), accessLevel: (read: .internal, write: .internal), isComputed: false)))
                 }
 
+                it("extracts lazy property correctly") {
```
ilyapuchka: please add the test instead in a `FileParserAttributesSpec` spec, there is a context for variables, the test will also become shorter. The test fails because you don't provide `description` parameter, so it makes it as `@lazy`. You'll see similar test for `final` attribute in `FileParserAttributesSpec`

minuscorp: Gonna jump into this!

minuscorp:
Also, I'd like to propose a computed property for it, like `isLazy`, I don't know how welcoming is this idea or how to implement it ATM.

krzysztofzablocki:
@minuscorp sure, just add the `isLazy` computed property and use the attributes check for it, nothing special is needed here

ilyapuchka:
@minuscorp computed property would be helpful, yes, look at `isFinal`, it's implemented the same way

minuscorp:
Done! 😄

SourceryRuntime/Sources/Variable.swift
```diff
@@ -52,6 +52,11 @@ public typealias SourceryVariable = Variable
     public var isFinal: Bool {
         return attributes[Attribute.Identifier.final.name] != nil
     }
+    
+    // Whether variable is lazy or not
+    public var isLazy: Bool {
```
SourceryBot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/krzysztofzablocki/Sourcery/blob/db9152f39b1f42a28afb27d81c5a13ed5e0126a0/SourceryRuntime/Sources/Variable.swift#L57"></span>Undocumented symbol.</td>
    </tr>
  </tbody>
</table>

minuscorp: 🤦‍♂️ :

ilyapuchka: It's fine =)

PR #690, opened 2018-10-01
Update Stencil

ilyapuchka:


SourceryRuntime/Sources/TemplateContext.swift
```diff
@@ -230,4 +230,7 @@ extension ProcessInfo {
         }
     }
 
+    public override func responds(to aSelector: Selector!) -> Bool {
+        return true
```
ilyapuchka: Now Stencil is smarter and on `NSObject`s calls this method before using KVC, so we have to override it to be able to access types by their names.

Issue #659, opened 2018-08-06
Auto inline into a file with several types introduce the annotations wherever.


minuscorp:
So I'm having issues with `// sourcery:inline:auto:{{ type.name }}` annotations. It seems that when more than one type is in the same file which gets all the requirements, Sourcery injects the code wherever (without any pattern that I could've found).

This is the example stencil:
```
{% for type in types.all where type.name|lowercase|!contains:"test" and type.implements.Action or type.implements.CompletableAction or type.implements.EmptyAction %}

    // sourcery:inline:auto:{{ type.name }}.AutoEquatableAction
    func someImportantFuncRequiredByTheProtocol() {
    }
   // sourcery:end
{% endfor %}
```
Imagine a file which has the following content:
```swift
class GetProvincesAction: Action {
}
class GetProvincesCompleteAction: CompletableAction {
}
class GetPersonalDataCompleteAction: CompletableAction {
}
```

It just messes up everything and inserts the code mixed. Just once you have moved the wrong placed annotations Sourcery does not mix them in further calls.

Any suggestion? I think this is a bug with the reading of the inline annotations 🤔

minuscorp:
I could try to give this a try if someone could give me some insight on how the process of embedding the `inline` annotations work.

jcislinsky:
+1

krzysztofzablocki:
duplicate of #601

PR #692, opened 2018-10-06
Do not cut out generated content when can't find location for inline annotation

### LINKED ISSUE #684: Proposal: Optional inline or file code generation from the same template
I'd like to get feedback on an idea I had regarding inline code gen (
https://cdn.rawgit.com/krzysztofzablocki/Sourcery/master/docs/writing-templates.html#inline-code-generation)

Couldn't we make inlining the generated code optional?

### Happy Case: Inline on both ends

In template:

```
{% for type in types.all %}
// sourcery:inline:{{ type.name }}.TemplateName
    var generatedStuff = 9876
// sourcery:end
{% endfor %}
```

In source code:

```swift
class MyType {
    var otherStuff = 123
// sourcery:inline:MyType.TemplateName
// sourcery:end
}
```

This will produce:

```swift
class MyType {
    var otherStuff = 123
// sourcery:inline:MyType.TemplateName
    var generatedStuff = 9876
// sourcery:end
}
```

### Sad path: no target :(

In template:

```
{% for type in types.all %}
// sourcery:inline:{{ type.name }}.TemplateName
    var generatedStuff = 9876
// sourcery:end
{% endfor %}
```

In source code:

```swift
class MyType {
    var otherStuff = 123 
}
```

**Note the absence of the inline comments in the code!**

This will produce:

```swift
class MyType {
    var otherStuff = 123
}
```

## Proposed solution

Given the same template as above and the code without inline comments, I'd want the generated code to end up in the `OUTPUTNAME.generated.swift` file just like the rest of the template outside the inline block's bounds.

- This way, users can opt in to have the generated code put inline or in separate files _from the same template._
- This also enables you to use a template to generate code one way for the app, and a different way for the test targets.

The 2nd case is the one that I'm currently craving for: to specify a location in my test code where the generated code should be put. Now I have to run the tool and copy & paste manually, which is no big deal, but this would be nice and sounds manageable to implement.

So, please give me your feedback on this! If it's overwhelmingly positive, I'll implement this later.

ilyapuchka:
Resolves #684 
By separating extracting annotated ranges and "cutting" then we now don't cut out ranges for which we can't find location to paste them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Support for parsing lazy vars into Variable's attributes.
 - Updated Stencil to 0.13.1 and SwiftStencilKit to 2.7.0
 - In Swift templates CLI arguments should now be accessed via `argument` instead of `arguments`, to be consistent with Stencil and JS templates.
+- Fixed missing generated code annotated with `inline` annotation when corresponding annotation in sources are missing. This generated code will be now present in `*.generated.swift` file.  
 
 ## 0.15.0
 
--- a/Sourcery/Parsing/Utils/InlineParser.swift
+++ b/Sourcery/Parsing/Utils/InlineParser.swift
@@ -17,10 +17,22 @@ internal enum TemplateAnnotationsParser {
     }
 
     static func parseAnnotations(_ annotation: String, contents: String, aggregate: Bool = false) -> (contents: String, annotatedRanges: [String: [NSRange]]) {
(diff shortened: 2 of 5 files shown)


minuscorp:
Maybe related to #659 ?

ilyapuchka:
@minuscorp I think it's a different issue cause by wrong range calculations. I couldn't reproduce it in tests so if you can submit a PR with failing example will be great

minuscorp:
I'll try to! Messy week this one...

PR #693, opened 2018-10-07
Allow file scope declarations (types, extensions) in swift templates

ilyapuchka:
Instead of using an extension on `TemplateContext` we now simply extract its properties into global variables with the same names. This way we don't need anymore a template code to be placed in an extension of `TemplateContext` and it is still accessible in the same way. The only downside of this is that if we add any properties to the `TemplateContext` type we need to update the template, but we haven't done this for ages, so should be good enough.
Also simplified a bit error handling as there were some unneeded jumping through the hoops.
This allows defining extensions or types without including separate swift files which was added in #666


anandabits:
@ilyapuchka can you add a test with a template that utilizes top-level only features?

PR #699, opened 2018-11-02
Update Autohashable for Swift 4.2

stephanecopin:
This PR updates the `AutoHashable` template so it uses the Swift 4.2 new `hash(into:)`'s method, rather than calculating a `hashValue`. Thanks to optional conformance defined of `Hashable` for `Optional` and `Collection`, this simplifies the template a bit. The annotation `skipHashing` allowing tweak which variable are excluded is still supported.
I've tried to keep the original structure the same, and just changed the implementation itself.

I've also included a commit to tentatively support inheritance for classes and `Hashable` - I'm not sure why it wasn't supported in the first place, so I'd definitely like another pair of eye to look at this.

This PR also makes a minor update that allows to use annotations to implement `AutoHashable` conformance.

This updated template has been in use in a production project for some time and no issues have been found with it so far. I would like it to have a first round of review before more extensive testing.

As a side note, since this doesn't touch the code but rather a template, I decided creating a PR directly is fine.

### DIFF
--- a/Templates/Tests/Context/AutoHashable.swift
+++ b/Templates/Tests/Context/AutoHashable.swift
@@ -134,7 +134,6 @@ class AutoHashableClass: AutoHashable {
     }
 }
 
-// Sourcery doesn't support inheritance for AutoHashable
 class AutoHashableClassInherited: AutoHashableClass {
     // Optional constants
     let middleName: String?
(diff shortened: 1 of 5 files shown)


ilyapuchka:
@stephanecopin thanks, that's a good idea, though maybe It would be better to make a different template, maybe `AutoHashableV2` (we don't have any smart versioning system 🤷‍♂️ ) Thoughts @krzysztofzablocki ?

krzysztofzablocki:
@ilyapuchka I don't think we need versioning for this, and this template update is a good improvement on the existing one.

My stance on the bundled templates is that if anyone disagrees with our take on the best way to implement specific template they should just copy them to their own project and modify as needed (this is very often required anyway because of different module imports that you might need in bigger projects).

I'd say we just keep updating the bundled templates to the best of our abilities

ilyapuchka:
My main concern is that when template starts to use a new language feature it will break generated code for those projects which for any reason are using the previous language version. Though keeping different versions of the same template will have maintenance cost and needed versions can always be retrieved from the history, so yeah.

krzysztofzablocki:
@ilyapuchka we can cut a bigger release number and that way it won't happen automatically for incremental pod updates, on that note maybe its time to make a 1.0 cut ?

ilyapuchka:
maybe, generics would be nice to have though, there are two PRs for that, but pretty big and incomplete, I think =/

krzysztofzablocki:
@ilyapuchka we can merge this one and make a new release in a while once those things are ready

stephanecopin:
@krzysztofzablocki @ilyapuchka that makes sense to me. My use case is also usually to copy the template in projects and either use them as-is or tweak them to our needs. Also, alternatively, I could easily update the template to use `#if swift(>=4.2)`? And keep the original <Swift 4.1 template in there in the `#else`, but maybe we wouldn't want to maintain the two templates. I'll let you guys decide

Templates/Tests/Context/AutoHashable.swift
```diff
@@ -153,3 +216,13 @@ class AutoHashableNSObject: NSObject, AutoHashable {
...
     }
 }
+
+// Should add super call
+class AutoHashableNSObjectInherited: AutoHashableNSObject {
+    let lastName: String
+
+    init(lastName: String) {
+        self.lastName = lastName
+            super.init(firstName: "")
```
ilyapuchka: indentation is incorrect here

PR #700, opened 2018-11-02
Add EnumsDefaults template

stephanecopin:
This PR adds a new template that generates helper functions allowing to use default values for Swift's enumerations with associated values, using sourcery's annotations system.
For example, with this template, consider the following enumeration:
```
enum AssociatedVariablesEnum {
	// sourcery: a = '"test"', b = "true", c = "0", d = "nil"
	case one(a: String, b: Bool, c: Int, d: Any?)
	// sourcery: 0 = '"test"'
	case two(String, Bool, Int, d: Any?)
	// sourcery: a = '"test"', d = "nil", c = "true", 1 = "true"
	case three(a: String, b: Bool, c: Int, d: Any?)
	// sourcery: a = '"test"', c = "true", 1 = "true", d = "DefaultImplementation()"
	case four(a: String, b: Bool, c: Int, d: Protocol)
}
```
may then be used in the following way:
```
let test1: AssociatedVariablesEnum = .one
let test2: AssociatedVariablesEnum = .two(d: nil)
let test3: AssociatedVariablesEnum = .three(true, 0, d: nil)
let test4: AssociatedVariablesEnum = .four(b: true, 0)
```
And all the other combinations of default parameters.
The variable name has to be used to specify its default value, and if the variable name is not specified then its parameter index can be specified.

Swift prevents by default from adding a static function that adds the default value:
```
enum Test {
    case test(String?)

    static func test(_ value: String? = nil) { // Will not compile
        return .test(value)
    }
}
```
However, with overloading it become possible to do it without an ambiguity:
```
enum Test {
    case test(String?)

    static func test() { // Works
        return .test(nil)
    }
}
```
Writing such overloads is possible to do manually for a small amount of default values, but as you can see from the size of the generated file, it can quickly get out of hand (and unmaintanable), hence how this template was born.

I'd love to get feedback (also the name can change, I had no inspiration for a short but meaningful name).

### DIFF
--- a/Templates/Templates.xcodeproj/project.pbxproj
+++ b/Templates/Templates.xcodeproj/project.pbxproj
@@ -43,6 +43,11 @@
 		6DC27BD81EC6592700B73CAF /* AutoHashable.expected in Resources */ = {isa = PBXBuildFile; fileRef = 6DC27BD71EC6592700B73CAF /* AutoHashable.expected */; };
 		B50435DF2157222E00F120DF /* AutoCodable.expected in Resources */ = {isa = PBXBuildFile; fileRef = 636D786B2083B8D700160CC4 /* AutoCodable.expected */; };
 		C03FD2E6F7645376C192094D /* Pods_TemplatesTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 26AD1250F60484521E03C14A /* Pods_TemplatesTests.framework */; };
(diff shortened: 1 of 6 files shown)


ilyapuchka:
That's a good idea, @stephanecopin, though I believe this will become unneeded as soon as [SE-0155](https://github.com/apple/swift-evolution/blob/master/proposals/0155-normalize-enum-case-representation.md) lands, probably in Swift 5 (?)
So knowing that I wouldn't proceed with including this template.

stephanecopin:
@ilyapuchka I see, yeah I agree then. I figured people could still have the need for it (I surely did at some point 😉), so feel free to close/merge/leave open for the time being.

PR #706, opened 2018-11-28
Remove SwiftTryCatch from the dependencies

f-meloni:
While i was trying to import Sourcery via SPM i've got this message 

screenshot 2018-11-28 at 15 42 23

I had a quick look and i saw that `SwiftTryCatch` was used only on `Sourcery.swift` and no one of the two blocks that i changed was actually throwing, then I think it was not even working as expected.
Then my proposal is just remove the dependency.
SPM can now import my forked branch as dependency


krzysztofzablocki:
SwiftTryCatch is for catching objective-c exceptions, those functions are throwing and you can't just remove it because it will break functionality we rely on

f-meloni:
Ok, thank you @krzysztofzablocki

PR #660, opened 2018-08-08
Add generics to types

truizlop:
As stated in issue #551, current type declaration only provides a boolean flag stating whether a type has a generic declaration or not. This PR parses the list of generic types associated to a type declaration as an `Array<String>`. They can be accessed during generation through the property `type.genericTypes`. Default value (when type is not generic) is an empty array.

Issue #87, opened 2016-12-27
Adding support for external source, e.g. Foundation or 3rd Party libraries


swizzlr:
Feature request: access Foundation types, standard library types, and even third party package types within stencils. This would make it a lot easier to generate boilerplate extensions. My current workaround is copy/pasting the swift interface to a file in the folder but not adding it to Xcode (hacky but works).

Not sure how we'd go about achieving this, could be a limit of SourceKit – maybe we could replicate the copy-paste hack by specifying a target, generating the interface, writing it to a temporary file and pointing SourceKit's type parser at it (which seems to happen at the Sema phase, pre-AST).

krzysztofzablocki:
We could introduce teaching phase to Sourcery, e.g. feed it foundation once and keep that data around, the problem with caching is invalidation, so we'd need to be careful when we require recaching those informations

ilyapuchka:
@krzysztofzablocki do you mean running Sourcery against swift libs source code somewhere locally, dumping result into file and bundling it with a binary to load it at runtime?

krzysztofzablocki:
that might be prefered for swift stdlib, but won't work for 3rd party so if we want to support 3rd party then we'd need to integrate more advanced caching

swizzlr:
Ideally SourceKitten/SourceKit can access Foundation/stdlib types natively. @jpsim is that possible?

jpsim:
> Ideally SourceKitten/SourceKit can access Foundation/stdlib types natively

SourceKitten can get data relevant to Foundation/stdlib just fine. What info would you like out of them? Information regarding source parsing obviously won't be available for compiled modules, however.

swizzlr:
@jpsim listing of all types + their publicly available methods, variables, etc. The stuff you see in the generated interface.

@krzysztofzablocki currently sourcery pulls this info from a given set of source files somehow; what sort of data is that? Perhaps @jpsim could advise on how we can retrieve similar data from the stdlib and Foundation?

krzysztofzablocki:
we just scan source code, this isn't the issue. 

The question we need to figure out first is how often we need to invalidate those informations from stdlib / foundation, because we should not be scanning this each time, the amount of code we might have in Swift / Foundation would add overhead to each run and that's not needed because it shouldn't change

swizzlr:
I'm assuming that object code exports a listing of available types to declare its interface – it's that info I think we ought to access.

bolismauro:
+1 for this

[...]

PR #707, opened 2018-12-06
[WIP] Introduce new generic parsing system

DenTelezhkin:
This PR is a logical continuation of #660, for which i'm very grateful to @truizlop for starting this work, which served as a stepping stone for a lot of things in current PR. Unfortunately though, PR's currently diverged in very different directions, and current differences between PR's or this PR and code base is just too much to sink in from a single branch, so I'm making a separate PR to clearly view proposed changes.

### Current generics state

Currently most of generic types are parsed into `TypeName` class by leveraging it's `array`(of `ArrayType`), `dictionary`(of `DictionaryType`) and `generic`(of `GenericType`) properties.

This approach has several downsides. First, concrete types are not completely resolved by Composer. For example, it seems currently impossible to get types from Sourcery, whose concrete generic type implements a specific protocol. 

Secondly, current implementation does not make a difference between template type and concrete type, so `Observable<T>` and `Observable<Int>` are considered two different types, even though in reality they are the same type, but in second case specialized by concrete type. 

And thirdly, it feels awkward to go to `typeName` property of type and then to `array` property to get it's generic information, it feels like this information should be available on a type itself.

Main goal of those changes is to introduce new way of parsing and composing generic types from SourceKit and improve Sourcery understanding of generic types. Before I dive into implementation details, here is proposed glossary, that will be used throughout this PR.

## Proposed glossary

When speaking about generic parameters of a type, it feels neccessary to draw a clear line between parameters, that are just placeholders for a type, and parameters, whose generic type is concrete. So, when using following example:

```swift
struct A<B:C>: D {}

let d : A<E>
```

`A` is a *generic struct*
`B` is a *generic type placeholder*
`C` is a *generic type constraint*
`D` is an *inherited type*

and

`E` is a *generic type parameter*

### Proposed data structures

*generic type placeholder* and *generic type constraints* are represented by `GenericTypePlaceholder` class with following properties:

```swift
    /// Generic placeholder type name
    public var placeholderName: TypeName

    /// Generic type parameter constraints
    public var constraints: [Type]
```

*generic type parameter* is represented by `GenericTypeParameter` class with following properties:

```swift
    /// Concrete type name of this generic parameter
    public var typeName: TypeName

    /// Concrete generic parameter type, if known
    public var type: Type?
```

`Class`, `Struct`, `Enum` types should have `genericTypeParameters` and `genericTypePlaceholders` properties, that each have an array of `GenericTypeParameter` and `GenericTypePlaceholder`. 

Additionally, `Method` class should have only `genericTypePlaceholders` property, since it's currently impossible to specialize method with concrete generic parameters.

`Type` is considered generic, if it has at least one `generic type parameter` or `generic type placeholder`.

### Current state

This PR is **work in progress** and is not ready to be merged. It's currently in the middle between current generic handling system and a new one. And while all tests pass(locally, we'll see what CI has to say about that), I believe that there are cases with generic types that are broken, and test suite just does not show it, cause tests are not deep enough to show problems.

In most cases *generic type placeholders* and *generic type parameters* are correctly parsed, here's example of one of the tests:

```swift
it("Extracts protocol information for generic type arguments") {
    let result = parse("protocol Test {}; class Foo<T:Test> : Bar<T> {}")
    let expected = Class(name: "Foo", genericTypePlaceholders: [
        GenericTypePlaceholder(placeholderName: TypeName("T"), constraints: [ Protocol(name: "Test") ])
        ])
    expected.inheritedTypes = [
        Type(name: "Bar", genericTypeParameters: [
            GenericTypeParameter(typeName: TypeName("T"), type: Type(name: "T"))]
        )]
      expect(result).to(equal([expected, Protocol(name: "Test")]))
}
```

As you can see, there is a clear difference between concrete and template generic type, and we are able to not only see `Test` requirement for generic type, but even properly understand that it's a protocol defined earlier. Also, `inheritedTypes` property now contains `[Type]` instead of `[String]`.

Note on inheritedTypes and SourceKit: even though SourceKit reports some generic information through `inheritedTypes` property, `where` clauses are currently lost - https://github.com/jpsim/SourceKitten/issues/554. We could parse those types completely ignoring SourceKit output, but since I expect SourceKit and SourceKitten to improve over time, I chose hybrid approach - `where` clauses are parsed by `Sourcery` codebase, and `inheritedTypes` are parsed from `SourceKitten` output. When SourceKit starts to report `where` clauses in `inheritedTypes` property, we can simply remove code, that currently parses `where` clauses manually.

Dictionary and array shorthand syntax is not implemented in new generic system, and is hacked to work similarly to way it was working before. There is a big amount of hacks thrown onto each other in `Composer` class to restore previous functionality with parsing generics into typeNames. Mostly it was needed because new generic system parses generic like `Array<Foo>` as a typeName `Array` instead of typeName `Array<Foo>`, as it was done before, thus breaking old code, and requiring more hacks on `Composer` level to restore previous functionality.

Also some changes are made to how `Composer` handles unknown types, I feel that it's not necessary to lose types even if they are unknown to us, so `Composer` now tries to not lose type information if it's available even partially.

Also, as a first time contributor to `Sourcery`, I may have made some stupid mistakes, that went unnoticed =) Obviously, on later stages this code would require review, fixes, and commits should be squashed to reasonable amount.

### Next steps

I want to gather feedback on general direction of this PR and make a decision on whether supporting current generic system is worth, since currently it takes huge amount of time and effort to backport all changes to generics to old code, and i'm not sure it's worth sinking that time in, if new system can cover more cases than the current one.

Here's a list of steps i'm proposing to make:

1. Remove `ArrayType`, `DictionaryType`, `GenericType` and their respective properties
2. Either [allow Sourcery to scan Foundation](https://github.com/krzysztofzablocki/Sourcery/issues/87), which is definitely out of scope of this proposal, or for now educate `FileParser` class about existance of `Array` and `Dictionary` generic types, which will allow to parse them as usual generic types without special handling they require now.
3. Add proper handling for dictionary and array shorthand syntax in FileParser. This will allow to cleanup huge amounts of hacks I made in `Composer` class that parse those types the old way.
4. Migration guide. Those will be breaking changes for users that rely on current generics handling system.

I would love to hear any thoughts on where this is heading, and adjust implementation to handle generics the best and most convenient way we can.


SourceryRuntime/Sources/Type.swift
```diff
@@ -44,7 +44,20 @@ import Foundation
...
+        return !genericTypeParameters.isEmpty || !genericTypePlaceholders.isEmpty
+    }
+
+    /// Whether type has been concretely specified
+    public var isConcreteGenericType: Bool {
+        return isGeneric && !genericTypeParameters.isEmpty
+    }
+
+    // Generic type placeholders
+    public var genericTypePlaceholders: [GenericTypePlaceholder]
```
SourceryBot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/krzysztofzablocki/Sourcery/blob/7e3932bd24a1073253ad00287b3aca685547c63d/SourceryRuntime/Sources/Type.swift#L57"></span>Undocumented symbol.</td>
    </tr>
  </tbody>
</table>

SourceryRuntime/Sources/TypeName.swift
```diff
@@ -264,6 +263,36 @@ public protocol Typed {
     // sourcery:end
 }
 
+// Describes Swift generic type placeholder
+@objcMembers public final class GenericTypePlaceholder: NSObject, SourceryModel {
```
SourceryBot: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/krzysztofzablocki/Sourcery/blob/7e3932bd24a1073253ad00287b3aca685547c63d/SourceryRuntime/Sources/TypeName.swift#L267"></span>Undocumented symbol.</td>
    </tr>
  </tbody>
</table>

truizlop:
Very nice work @DenHeadless, it looks like you have put a lot of thought into this and generalized very well the initial proposal I did. I presume the other PR can be disregarded after this one gets approved. It looks good to me, but as I don't have a full overview of the project and the direction it is going, I think @ilyapuchka or @krzysztofzablocki should take a look at it. I think this addition is necessary to bring more flexibility to Sourcery and I'd be happy to help to move it forward.

truizlop:
Hi! This has been sitting here for a very long time and I'd like to know the plans about moving it forward. I think having more information about generic types is a useful feature that everyone can benefit from. The work is already initiated but it would be nice to know how we can move it forward and add it to a release in the near future.

Any opinions? /cc @ilyapuchka @krzysztofzablocki @Liquidsoul

ilyapuchka:
Seems like a sensible approach (I didn't look into code as didn't work with it for quite a while, so it wouldn't help, let's hope tests will be sufficient to ensure correctness). But I have few concerns:

- this seems to be quite a big breaking change, isn't it?
- I'm not sure how exactly current `typeName.array` or `typeName.generic` are limiting (again, didn't work with a code for long time so I don't remember details) and why changing rather than extending these API is necessary
- changing from type name `Array<Foo>` to `Array` will seem to be such a big breaking change that would add unnecessary complexity to the templates which don't care about type details and only need to put in a correct type (if I understood the change correctly). It seems like such templates will need to explicitly check each type if it is generic and additionally access its generic data in order to reconstruct it, is this correct? If so how this can be avoided preferably making it non-breaking change?
- considering placeholders/parameters will the type be considered generic if it is defined as `class Foo: Bar<Int>`? There also can be a case where type have both parameters and placeholders - `class Foo<A>: Bar<A, Int>`. Same applied to generic typealiases.
- considering scanning Foundation I believe this is unneeded to support generics and probably shouldn't be even implemented by Sourcery itself (though it could maybe support providing interface information for external sources).

Regarding plans about moving this forward as @DenTelezhkin pointed this is not ready to be merged yet, so I'd say it depends on his and yours @truizlop (and whoever is up for it) initiative to push this further to the more or less complete state.

DenTelezhkin:
Hi guys! 
First off, i would like to apologize, as it seems that approach to those changes was not great. Small additive improvements turned out to be large rewrite with some huge breaking changes.

The main benefits of those is to give Sourcery and users more knowledge about generic types. Currently most of generic information is parsed only on one level, so nested generics, generic subclasses, properties that have generic types and a lot of other data structures simply do not have generic information available.

I believe we should restart this with a plan and series of smaller PRs to achieve desired goal. I still believe that breaking changes are worth it, but maybe they could be done in the end, and be more clear then. But anyway, it's important to discuss possible future breaking changes now, so that we all know what to expect.

We can try leaving brackets syntax as it is now(Array<Foo>), though it may not be easy or possible. The main things that concern me are as follows.

1. There is this notion of "known" and "unknown" types, described here https://cdn.rawgit.com/krzysztofzablocki/Sourcery/master/docs/writing-templates.html#what-are-em-known-em-and-em-unknown-em-types. I'm not sure i fully understand why Sourcery needs to care if type is known or unknown. Specifically in this PR this was already changed to never lose type information, if it's available, even though declaration of type may not be in scanned sources. This is critical for data-mining type information, since otherwise type information may not be complete.
2. Currently generic information lives in `TypeName` properties instead of `Type` properties. I'm not sure about this, since those should ideally be in `Type` to allow digging deeper into nested types, and `TypeName` could mostly serve for description purposes. This is one of the biggest breaking changes that this PR can bring. 
3. Removing types from brackets - changing `Array<Element>` to `Array`. This might have been unnecessary, but was done to correctly unwrap generics, for example for complicated cases where Element itself is nested - Array<Generic<Foo>>. It's possible that we can leave type name as `Array<Element>`, need to dig deeper.

Looking at 2 point though, i think this is an opportunity to simplify generic handling in Sourcery. For example `array` and `dictionary` getters currently have their own type - `ArrayType` and `DictionaryType`, which is an overkill, those getters might simply look into `Type` itself, and if it's an Array<Element> or Dictionary<Element>, to return this type, and don't return it if generic is different. Ideally, i think it can be possible to start parsing types by already knowing two default ones - Array and Dictionary, with shorthand syntax support. This way a lot of abstractions for Array and Dictionary can be scaled down. But in this PR i did not find an easy way to teach `FileParser` or `Composer` about that.

Also `TypeName` was a biggest problem for me while i was digging into parsing, because every change i made to support generic Type parsing needed the backport to old `TypeName` architecture, which was really difficult to pull off. That's why i thought that if we agree on dropping it, implementing second system would be much faster.

Foof. My posts in this thread are incredibly long and boring, sorry about that =) @ilyapuchka what do you think, maybe you can give some feedback on those three points or point us to things that may be useful in implementing this stuff?

ilyapuchka:
1. Unknown types are special for Sourcery because for these types we don't know their nature, either they are structs, classes or enums, generic or not, and anything else that is defined in the sources that are not scanned. We only have information from their extensions (see `//for unknown types we still store their extensions` in `Composer.swift`) in the scanned sources.

2. Whether generics information is in `TypeName` or in `Type` is an implementation detail that has only effect on user-facing APIs in a way they provide access this information, either through `typeName` or `type` property of variable or other "typed" instances. If you manage to ensure that `type` is always (even in case of unknown types) set on variables and other types, then moving these generic properties there should be sensible. The main reason to have it in `TypeName` was that `typeName` is never `nil` (in contrast to `type`), so we can provide information about some standard generic types like arrays, but it requires custom parsing anyway. Ideally, with a new implementation, there shouldn't be more breaking changes to existing APIs other than changing from `typeName` to `type`(ideally it should be backward compatible).

3. Yes, this should not change as there are a lot of cases when templates just need to repeat the type from the original code, just using `typeName` property, no matter how it was defined, rather than making complex logical decisions based on the type structure. I think in the majority of cases full generic information is not needed (otherwise it will be supported already) or can be easily reconstructed with annotations. That's why introducing support for generics shouldn't result in breaking changes for trivial templates which don't require this information. It can be an option to introduce a different property that would serve the purpose of `typeName`, but that would be another massive breaking change.

DenTelezhkin:
Hey, folks!

I finally managed to find time to rescope this PR to a smaller one - #764. I stayed away from breaking changes, but moved all terminology from this PR there. It's a first step, next steps would probably be better handling for `inheritedTypes`, which currently stays as `[String]`, because changes to that are too huge, and it's better to move them to separate PR. 

Hope this time it will be easier to understand what is going on and why those changes are being made. This time #764 is not a WIP and is actually ready for review, because it stands on it's own.

DenTelezhkin:
Closing in favor of #764

Issue #713, opened 2018-12-20
Installing via Mint, and then Homebrew still results in not being able to run .swifttemplate


gbrhaz:
I installed Sourcery 0.15.0 using Mint first, but since the AutoCodable template is a .swifttemplate, I decided to try with Homebrew instead.

Unfortunately, running 'mint uninstall sourcery' and then 'brew install sourcery' I still get the error 

warning: Skipping template AutoCodable.swifttemplate. Swift templates are not supported when using Sourcery built with Swift Package Manager yet. Please use only Stencil or EJS templates. See https://github.com/krzysztofzablocki/Sourcery/issues/244 for details.

I have also tried deleting all known instances of sourcery in /usr/local and ~/Library/Caches/Homebrew

ilyapuchka:
@gbrhaz as the error message says swift templates are not supported for SMP builds (both Homebrew and Mint use it build Sourcery binary from its sources). 

@Liquidsoul have this constraint been lifted with #681 ? If yes, I suppose we need to drop this check and update docs to reflect it.

Liquidsoul:
@ilyapuchka you are right, once a version with #681 is released, swift templates will be available when using SPM. I'll look into double-checking that fact and make the changes to drop the check alongside the docs update.

PR #715, opened 2018-12-24
SourceryFramework

ilyapuchka:
This PR introduces SourceryFramework that exposes parsing and generating functionality (generating part is really just two protocols then implemented by the tool). Motivation for that is that this framework can provide the foundation for other tools, especially parsing, which don't necessarily want to use templates and generate code, without the need to include such functionality in the tool. There were multiple requests in the past about such features which were shut down because they were considered out of the scope of Sourcery as a tool.
Mainly this change is about moving files around and changing minimum amount of internal types and properties to public to make the main tool compile. We may want to revisit public API later to expose more things publicly.


Liquidsoul:
I've pushed a missing change to the SPM manifest with SourceryFramework.
However, we still have some dependency issues with some source files which belong to multiple targets in the Xcode project:
- `Sourcery/Utils/Path+Extensions.swift`
- `Sourcery/Utils/Sha.swift`
- `Sourcery/Version.swift`

One solution I see would be to add a `SourceryUtils` target which would include these files. I'll try to work something out.

Note that the `Sha.swift` is also included in `SwiftTemplate` which prevents me to fix #713. So we need this PR to fix it.

Liquidsoul:
I've extracted the files into `SourceryUtils`, tell we what you think about this @ilyapuchka and @krzysztofzablocki

krzysztofzablocki:
@ilyapuchka you can merge then if you are happy with it 👍

PR #718, opened 2019-01-21
Use updated version of SwiftTryCatch

klundberg:
If I try to use Sourcery with Swift Package Manager, I get this error:

```
error: the package sourcery[https://github.com/krzysztofzablocki/Sourcery] @ 0.15.0 contains incompatible dependencies:
    swifttrycatch[https://github.com/seanparsons/SwiftTryCatch.git] @ e7074a72e4d4dc516f391bc4d4afd8ca6a845b4b
```

SwiftTryCatch has a released version now, so this should point to that version now, which should be compatible with swift 4.2.


ilyapuchka:
@klundberg I believe this should point to the latest master (6a177252cfa2ef649b0aa7b2d16ab221386ca51c), not to 1.0.0 release which was done on 2015

klundberg:
You're right! I just made the change.

Issue #537, opened 2018-03-08
Is there a way to specify indentation level for inline templates?


garnett:
I'm using the following custom template for init methods:
```stencil
{% for type in types.implementing.AutoInitable %}
    // sourcery:inline:auto:{{ type.name }}.AutoInitable
    // swiftlint:disable line_length
    {{ type.accessLevel }} init({% for variable in type.storedVariables %}{{ variable.name }}: {{ variable.typeName }}{% if not variable.name == type.storedVariables.last.name %}, {% endif %}{% endfor %}) {
        {% for variable in type.storedVariables %}
            self.{{ variable.name }} = {{ variable.name }}
        {% endfor %}
    }
    // swiftlint:enable line_length
    // sourcery:end
{% endfor %}
```

it works fine for non-nested definitions, but, if applied to nested models it has 1 level indentation:

Now:

```
class A {
  class B {
  init(..)
  }
  init(..)
}
```

Expected:

```
class A {
  class B {
    init(..)
  }
  init(..)
}
```

ilyapuchka:
Should be possible with new indent filter in Stencil (https://github.com/kylef/Stencil/pull/188), but we need to update to master version, or there should be a new release of Stencil

ilyapuchka:
You can workaround this like [here](https://gist.github.com/Liquidsoul/efa4f65af055acfde64afed6cda0007d):

```jinja
{% set spacing %}{% if type.parentName %}    {% endif %}{% endset %}
```

ilyapuchka:
@garnett master now uses version of stencil that contains `indent` filter

PR #731, opened 2019-03-07
Indent inline generated code to match the `// sourcery:inline:` annotation

filipe-lemos:
This allows specifying the indentation of the inline generated code, as requested in #537, by matching it with the indentation of the `// sourcery:inline:` annotation.

This should be particularly useful for nested types with inline generated code, allowing it to keep the correct indentation, for example:
```swift
class Foo: AutoName {
    // sourcery:inline:Foo.AutoName
    var name: String { "Foo" }
    // sourcery:end

    class Bar: AutoName {
        // sourcery:inline:Bar.AutoName
        var name: String { "Bar" }
        // sourcery:end
    }
}
```
```
{% for type in types.implementing.AutoName %}
// sourcery:inline:{{ type.name }}.AutoName
var name: String { "{{ type.name }}" }
// sourcery:end
{% endfor %}
```

### DIFF
--- a/SourceryFramework/Sources/Parsing/Utils/InlineParser.swift
+++ b/SourceryFramework/Sources/Parsing/Utils/InlineParser.swift
@@ -48,7 +48,6 @@ public enum TemplateAnnotationsParser {
             let endLineRange = result.range(at: 5)
 
             let indentation = bridged.substring(with: indentationRange)
-            print("indentation raw \(indentation) \(indentation.count)")
             let name = bridged.substring(with: nameRange)
             let range = NSRange(
                 location: startLineRange.location,
--- a/SourceryFramework/Sources/Parsing/Utils/InlineParser.swift
+++ b/SourceryFramework/Sources/Parsing/Utils/InlineParser.swift
@@ -7,16 +7,18 @@ import Foundation
 
 public enum TemplateAnnotationsParser {
 
+    public typealias AnnotatedRanges = [String: [(range: NSRange, indentation: String)]]
+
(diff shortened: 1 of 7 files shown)


SourceryFramework/Sources/Parsing/Utils/InlineParser.swift
```diff
@@ -28,33 +30,36 @@ public enum TemplateAnnotationsParser {
...
-            let nameRange = result.range(at: 2)
-            let startLineRange = result.range(at: 3)
-            let endLineRange = result.range(at: 4)
+            let indentationRange = result.range(at: 2)
+            let nameRange = result.range(at: 3)
+            let startLineRange = result.range(at: 4)
+            let endLineRange = result.range(at: 5)
 
+            let indentation = bridged.substring(with: indentationRange)
+            print("indentation raw \(indentation) \(indentation.count)")
```
krzysztofzablocki: can we remove this debug print statement?

filipe-lemos: Of course! Oops... 😅

macduy:
This is a great change! Does it work with `// sourcery:inline:auto:...`?

filipe-lemos:
> This is a great change! Does it work with `// sourcery:inline:auto:...`?

The `:auto` is added automatically with no inline level, so that behaviour is kept the exact same. You can, after its created, indent the opening `:auto` comment it will indent the code block.

mr-fixit:
Thanks, @filipe-lemos !

PR #732, opened 2019-03-12
Update AutoMockable.stencil

mpawlo:
Fix inconsistent naming of Invocation variable and call on it which has been cause of build errors.


krzysztofzablocki:
thanks for contributing! 
could you add the changelog entry?

PR #751, opened 2019-04-16
fix ReceivedInvocations's type

Cookiezby:
Recently I am using Sourcery to refactoring my project, it is awesome project!  I also use the AutoMockable to help me generate the mock class for unit test. Then I find a tiny problem.

```swift
//sourcery:AutoMockable
protocol Test {
    func send(message: String?)
}
```
the generated code is
```swift
class TestMock: Test {
    //MARK: - send

    var sendMessageCallsCount = 0
    var sendMessageCalled: Bool {
        return sendMessageCallsCount > 0
    }
    var sendMessageReceivedMessage: String?
    var sendMessageReceivedInvocations: [String] = []
    var sendMessageClosure: ((String?) -> Void)?

    func send(message: String?) {
        sendMessageCallsCount += 1
        sendMessageReceivedMessage = message
        sendMessageReceivedInvocations.append(message) // this will cause compile error
        sendMessageClosure?(message)
    }
}
```
So I edit the AutoMockable.stencil, and It works, hope my solution is the right way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 ## Master
 
 ### Bug fixes
-
+- Fix ReceivedInvocations's type for the method which have only one parameter in AutoMockable.stencil
 - Fix missing folder error that could happen when running a Swift template with existing cache
 - Don't add indentation to empty line when using inline generated code.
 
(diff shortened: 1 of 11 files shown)


Cookiezby:
I found use typeName directly will have some problem with the closure, so I check the type, if it is optional, add a "?" at the end of the type.

krzysztofzablocki:
hey @Cookiezby, thanks for contributing, could you add the case this fixes to TemplateTests projects and update the expected files to verify that this works correctly?

Cookiezby:
Hi @krzysztofzablocki , thank for your response, I will check it.

Cookiezby:
@krzysztofzablocki  Hi, I check the template project, and found the AutoMockable.expected is little different according to the AutoMockable.stencil. The name of the parameter name XXXInvacations seem to be XXXReceivedInvacations, so I fix the name in expected file.

ilyapuchka:
Thanks @Cookiezby !

PR #752, opened 2019-04-16
Fix annotations EnumCase not supporting commas

stephanecopin:
There was an issue when parsing annotation cases where basically the parser was trying to split all individual associated values using `,`, ignoring the fact that these commas may very well be in comments.

For example, the following `enum`:
```
enum Foo {
    case optionA(/* sourcery = first, second = "value" */ Int) 
}
```
The associated value would be parsed as:
Screenshot 2019-04-16 11 35 35

This PR address the issue by adding a method:
```swift
components(separatedBy delimiter: String, excludingDelimiterBetween between: (open: [String], close: [String]))
```
that the previous one calls in the following way:
```swift
func components(separatedBy delimiter: String, excludingDelimiterBetween between: (open: String, close: String)) -> [String] {
    return self.components(separatedBy: delimiter, excludingDelimiterBetween: (between.open.map { String($0) }, between.close.map { String($0) }))
}
```
So that we can optionally specify (`//`, `\n`) and (`/*` and `*/`) as delimiters to ignore (note that the documentation comments `/**` and `///` are then automatically included).

This change allowed me to update the method to a version that I hope is easier to understand, reverting it to [a state]((https://github.com/krzysztofzablocki/Sourcery/blob/25b1c50f27a134beb19f4bfbb9a593e2bcdd2699/Sourcery/Parsing/Utils/Extensions.swift)) that's close to where it was before closure type detection was introduced in #131.

This PR passes all the unit tests (on my machine), but considering the method it changes is used in a lot of places and I haven't had time to test it on a real-world project, I'm creating the PR as a draft. This should also give you the opportunity to give me feedback.

### DIFF
--- a/Sourcery.xcodeproj/project.pbxproj
+++ b/Sourcery.xcodeproj/project.pbxproj
@@ -706,7 +706,10 @@
 				003958827841F12BBF99EB67 /* Frameworks */,
 				BF6C9BBD135B544916C5F0CD /* Pods */,
 			);
+			indentWidth = 4;
 			sourceTree = "<group>";
+			tabWidth = 4;
+			usesTabs = 0;
 		};
 		CD754C8C1D853F000082B512 /* Products */ = {
 			isa = PBXGroup;
--- a/SourceryFramework/Sources/Parsing/FileParser.swift
+++ b/SourceryFramework/Sources/Parsing/FileParser.swift
@@ -655,7 +655,7 @@ extension FileParser {
     fileprivate func parseEnumAssociatedValues(_ body: String) -> [AssociatedValue] {
         guard !body.isEmpty else { return [AssociatedValue(localName: nil, externalName: nil, typeName: TypeName("()"))] }
 
-        let items = body.commaSeparated()
+        let items = body.commaSeparated(ignoringComments: true)
         return items
             .map({ $0.trimmingCharacters(in: .whitespacesAndNewlines) })
             .enumerated()
--- a/SourceryFramework/Sources/Parsing/FileParser.swift
+++ b/SourceryFramework/Sources/Parsing/FileParser.swift
@@ -655,7 +655,10 @@ extension FileParser {
     fileprivate func parseEnumAssociatedValues(_ body: String) -> [AssociatedValue] {
         guard !body.isEmpty else { return [AssociatedValue(localName: nil, externalName: nil, typeName: TypeName("()"))] }
 
-        let items = body.commaSeparated(ignoringComments: true)
+        let items = body.components(
+            separatedBy: ",",
+            excludingDelimiterBetween: (["<", "[", "(", "{", "/*", "//"], ["}", ")", "]", ">", "*/", "\n"])
+        )
         return items
             .map({ $0.trimmingCharacters(in: .whitespacesAndNewlines) })
             .enumerated()
--- a/SourceryRuntime/Sources/Extensions.swift
+++ b/SourceryRuntime/Sources/Extensions.swift
@@ -112,13 +112,8 @@ public extension String {
 
     /// :nodoc:
     /// Returns components separated with a comma respecting nested types
-    func commaSeparated(ignoringComments: Bool = false) -> [String] {
-        var (open, close) = (["<", "[", "(", "{"], ["}", ")", "]", ">"])
-        if ignoringComments {
-            open += ["/*", "//"]
-            close += ["*/", "\n"]
-        }
-        return components(separatedBy: ",", excludingDelimiterBetween: (open, close))
+    func commaSeparated() -> [String] {
+        return components(separatedBy: ",", excludingDelimiterBetween: ("<[({", "})]>"))
     }
 
     /// :nodoc:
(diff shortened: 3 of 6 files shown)


krzysztofzablocki:
thanks for contributing, I don't see anything wrong with it, @ilyapuchka since you did the closure deteciton that stephane is talking about mind taking a look at the diff?

stephanecopin:
@krzysztofzablocki Thanks! I've just tested it on a fairly large projects using a bunch of custom templates, and there were no issues at all. Considering all unit tests passes as well, moving this to ready for review.

SourceryRuntime/Sources/Extensions.swift
```diff
@@ -130,47 +135,46 @@ public extension String {
...
         var item = ""
         var items = [String]()
         var matchedDelimiter = (alreadyMatched: "", leftToMatch: delimiter)
 
-        for char in self {
-            if between.open.contains(char) {
-                if !(boundingCharactersCount == 0 && String(char) == delimiter) {
+        var i = self.startIndex
+        while i < self.endIndex {
+            var increaseOffset = 1
```
ilyapuchka: please rename this to `offset` or `indexOffset`

SourceryRuntime/Sources/Extensions.swift
```diff
@@ -130,47 +135,46 @@ public extension String {
...
+            defer {
+                i = self.index(i, offsetBy: increaseOffset)
+            }
+            let current = self[i..<(self.index(i, offsetBy: delimiter.count, limitedBy: self.endIndex) ?? self.endIndex)]
```
ilyapuchka: what does `current` mean in this context?

SourceryRuntime/Sources/Extensions.swift
```diff
@@ -130,47 +135,46 @@ public extension String {
...
+            if let index = between.open.index(where: { String(self[i...]).starts(with: $0) }) {
```
ilyapuchka: This seems to find the opening delimiter that string suffix starts with, but the index is then not being used for anything but getting this delimiter, so I guess it could be access here with `between.open.first(where:)` which will make the intention more clear.
Also `String(self[i...])` should be calculated only once.

SourceryRuntime/Sources/Extensions.swift
```diff
@@ -130,47 +135,46 @@ public extension String {
...
+                if !(boundingCharactersCount == 0 && String(self[i]) == delimiter) {
                     boundingCharactersCount += 1
                 }
-            } else if between.close.contains(char) {
+                increaseOffset = between.open[index].count
+            } else if let index = between.close.index(where: { String(self[i...]).starts(with: $0) }) {
```
ilyapuchka: same here

SourceryRuntime/Sources/Extensions.swift
```diff
@@ -130,47 +135,46 @@ public extension String {
...
                 // do not count `->`
-                if !(char == ">" && item.last == "-") {
+                if !(self[i] == ">" && item.last == "-") {
```
ilyapuchka: please extract `self[i]` into local variable

SourceryTests/Parsing/FileParserSpec.swift
```diff
@@ -369,13 +373,13 @@ class FileParserSpec: QuickSpec {
...
                     it("extracts associated value annotations properly") {
-                        let result = parse("enum Foo {\n case optionA(\n// sourcery: annotation\nInt)\n case optionB }")
+                        let result = parse("enum Foo {\n case optionA(\n// sourcery: first\n// sourcery: second, third = \"value\"\nInt)\n case optionB }")
                         expect(result)
                             .to(equal([
                                 Enum(name: "Foo",
                                      cases: [
                                         EnumCase(name: "optionA", associatedValues: [
-                                            AssociatedValue(name: nil, typeName: TypeName("Int"), annotations: ["annotation": NSNumber(value: true)])
+																					AssociatedValue(name: nil, typeName: TypeName("Int"), annotations: ["first": NSNumber(value: true), "second": NSNumber(value: true), "third": "value" as NSString])
```
ilyapuchka: please fix indentation here

SourceryRuntime/Sources/Extensions.swift
stephanecopin: `current` represents the current "part" of the string being scanned (since it can be more than more character). I was trying to find an equivalence of `char` but for more than one character, but that's the best I could come up with. If you do feel free to say an d I'll update 😉

SourceryRuntime/Sources/Extensions.swift
stephanecopin: Makes sense to me. Will update

SourceryRuntime/Sources/Extensions.swift
stephanecopin: I can do that, but I'm unsure if this will contribute to having the code more readable - `self[i]` is already pretty straightforward IMHO

SourceryRuntime/Sources/Extensions.swift
ilyapuchka: `currentlyScanned` maybe? Definitely longer, but better be clear than short.

stephanecopin:
@ilyapuchka I've just updated my PR with your suggested changes, thanks for the feedback 🙂 
Hopefully the tests passed, they're not passing on my machine anymore, probably because of some Xcode vesion issue

stephanecopin:
hey @ilyapuchka, you might have missed my update but if this looks good it'd be awesome if it gets merged. Just rebased on top of `master`

Sourcery.xcodeproj/project.pbxproj
```diff
@@ -706,7 +706,10 @@
 				003958827841F12BBF99EB67 /* Frameworks */,
 				BF6C9BBD135B544916C5F0CD /* Pods */,
 			);
+			indentWidth = 4;
```
ilyapuchka: is this needed?

SourceryRuntime/Sources/Extensions.swift
```diff
@@ -112,8 +112,13 @@ public extension String {
 
     /// :nodoc:
     /// Returns components separated with a comma respecting nested types
-    func commaSeparated() -> [String] {
-        return components(separatedBy: ",", excludingDelimiterBetween: ("<[({", "})]>"))
+    func commaSeparated(ignoringComments: Bool = false) -> [String] {
```
ilyapuchka: rather than introducing a boolean parameter that is only applicable when dealing with enums it might be better to use `componets(separatedBy: excludingDelimiterBetween:)` directly in the `parseEnumAssociatedValues` method.

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -655,7 +655,7 @@ extension FileParser {
     fileprivate func parseEnumAssociatedValues(_ body: String) -> [AssociatedValue] {
         guard !body.isEmpty else { return [AssociatedValue(localName: nil, externalName: nil, typeName: TypeName("()"))] }
 
-        let items = body.commaSeparated()
+        let items = body.commaSeparated(ignoringComments: true)
```
ilyapuchka:
```suggestion
        let items = body.components(
            separatedBy: ",", 
            excludingDelimiterBetween: (["<", "[", "(", "{", "/*", "//"], ["}", ")", "]", ">", "*/", "\n"])
        )
```

ilyapuchka:
I'd wait to merge this until #760

Sourcery.xcodeproj/project.pbxproj
stephanecopin: @ilyapuchka I think so, my personal settings are different from Sourcery's (i.e. 2 indent width; tabs, but let's not discuss that 😛), and I think contributors shouldn't have to update their preferences just to contribute - it should be part of the project itself. What do you think?
cc @krzysztofzablocki

SourceryRuntime/Sources/Extensions.swift
stephanecopin: @ilyapuchka works for me, I'll update

stephanecopin:
@ilyapuchka updated the PR and rebased, hopefully should be ready for merge now 😉

Sourcery.xcodeproj/project.pbxproj
krzysztofzablocki: that's fair, anything that makes it easier to contribute

andynovak12:
@ilyapuchka any update on this? This fix would help me migrate my project to Xcode 10.2

PR #756, opened 2019-04-30
Fix build with Xcode 10.2

Liquidsoul:
Update Nimble to 8.0.1 to fix build with Xcode 10.2.

This **does not** migrate the code to Swift 5.

PR #760, opened 2019-05-16
Update project to run with Xcode 10.2

DenTelezhkin:
Hi!

This PR is aimed at updating project to Xcode 10.2 and does several things:

* Branch off master to include fix from #758
* Update Nimble, similar to #756
* Switches to Xcode 10.2.1 for CircleCI
* Fixes several tests, mostly for two failures - `.internal` access is now being reported for extensions on types, and String interpolation now reports compile error differently in `rethrows template parsing errors` test
* Update all projects to Xcode 10.2 recommended settings
* Fixes several warnings in unit tests
* Adds minimum deployment target for SPM package - macOS 10.11. Without this change SPM build fails because of NSKeyedArchiver API that was introduced in macOS 10.11(SPM chooses 10.10 by default) - for reference https://github.com/Alamofire/Alamofire/pull/2737 and https://github.com/Alamofire/Alamofire/pull/2736.

Overall changes seem backwards-compatible and non-breaking, however not 100% sure about that. @Liquidsoul maybe you can look into those and give your feedback?

### DIFF
--- a/Sourcery.xcodeproj/project.pbxproj
+++ b/Sourcery.xcodeproj/project.pbxproj
@@ -1108,7 +1108,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0920;
-				LastUpgradeCheck = 0930;
+				LastUpgradeCheck = 1020;
 				ORGANIZATIONNAME = Pixle;
 				TargetAttributes = {
 					0054C50322496806007F9AA7 = {
@@ -1153,6 +1153,7 @@
 			developmentRegion = English;
 			hasScannedForEncodings = 0;
 			knownRegions = (
+				English,
 				en,
 				Base,
 			);
@@ -1705,7 +1706,7 @@
 				CLANG_ENABLE_OBJC_WEAK = YES;
 				CLANG_WARN_DOCUMENTATION_COMMENTS = YES;
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
-				CODE_SIGN_IDENTITY = "Mac Developer";
+				CODE_SIGN_IDENTITY = "";
 				CODE_SIGN_STYLE = Automatic;
 				COMBINE_HIDPI_IMAGES = YES;
 				CURRENT_PROJECT_VERSION = 1;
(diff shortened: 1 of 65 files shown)


DenTelezhkin:
Not sure why CI build was not triggered 🤔 Possibly related to the fact that I enabled Circle CI on personal account, where build already did pass? https://circleci.com/gh/DenTelezhkin/Sourcery/7

Sourcery.xcodeproj/project.pbxproj
```diff
@@ -1153,6 +1153,7 @@
 			developmentRegion = English;
 			hasScannedForEncodings = 0;
 			knownRegions = (
+				English,
```
Liquidsoul: I don't think this has an impact but this could be removed I think 🤔

Sourcery.xcodeproj/project.pbxproj
```diff
@@ -1705,7 +1706,7 @@
 				CLANG_ENABLE_OBJC_WEAK = YES;
 				CLANG_WARN_DOCUMENTATION_COMMENTS = YES;
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
-				CODE_SIGN_IDENTITY = "Mac Developer";
+				CODE_SIGN_IDENTITY = "";
```
Liquidsoul: Is this change necessary? Because this may break the release process.
If this is necessary, I'll see if this breaks during the next release.

DenTelezhkin: Not sure, this was changed by Xcode, however when I tried reverting this change, this is the warning I see:

```
TryCatch will not be code signed because its settings don't specify a development team. (in target 'TryCatch')
```

I would roll with Xcode recommendation, but up to you, tell me what you think, and I'll add a change with next commit.

Sourcery.xcodeproj/project.pbxproj
DenTelezhkin: Done.

Templates/Templates.xcodeproj/project.pbxproj
```diff
@@ -376,6 +376,7 @@
 			developmentRegion = English;
 			hasScannedForEncodings = 0;
 			knownRegions = (
+				English,
```
Liquidsoul: Let's revert this localization change too 😉

DenTelezhkin: Done

DenTelezhkin:
When this PR is ready to be merged, ping me, I'll squash irrelevant commits 😉

PR #762, opened 2019-05-18
Updating Nimble version

### LINKED ISSUE #761: Does not build test target on Xcode 10.2
There is a build error in the Nimble test dependency.

https://github.com/Quick/Nimble/issues/625

This was fixed and so the pod should be updated.

joesus:
Fixes #761


joesus:
Duplicate of #760

PR #764, opened 2019-06-02
Reintroduce new generic parsing system

DenTelezhkin:
This PR is a restart of #707, and a first step towards better generics handling by Sourcery.

Changes are intentionally smaller in scope than in the original, and are almost backwards-compatible. In fact, the only truly breaking change I see is `Type.isGeneric` property that became computed (previously it was passed in initializer), but it's unlikely that this change will break someone's usage of the framework.

This PR follows glossary and terminology proposed in #707, so it's better to read full explanation provided in that PR.

Brief summary of changes that were implemented here:
* Introduce `Type.genericTypePlaceholders` and `Type.genericTypeParameters` properties, fill them on FileParser and Composer level
* Introduce `Method.genericTypeParameters` property
* Preserve `Type` knowledge wherever possible, even if types are unknown - in scope of objects, that are related to generics
* Implement type specialization(for example `Observable<Element>` can be a template class, and `Observable<Int>` can be a specialized version of that class, that contains both generic type placeholder `Element` and generic type parameter `Int`)

Even though this PR is only the first step, and does not finish generics parsing fundamentally, it was designed to be a good incremental improvement to not turn this PR into breaking monster it was previously.
 
/cc @ilyapuchka @truizlop

### DIFF
--- a/SourceryFramework/Sources/Parsing/Composer.swift
+++ b/SourceryFramework/Sources/Parsing/Composer.swift
@@ -277,7 +277,7 @@ public struct Composer {
 
     private static func resolveProtocolCompositionTypes(_ protocolComposition: ProtocolComposition, resolve: TypeResolver) {
         let composedTypes = protocolComposition.composedTypeNames.compactMap { typeName in
-            resolve(typeName, protocolComposition)
+            resolve(typeName, protocolComposition, nil)
         }
 
         protocolComposition.composedTypes = composedTypes
--- a/SourceryFramework/Sources/Parsing/Composer.swift
+++ b/SourceryFramework/Sources/Parsing/Composer.swift
@@ -88,6 +88,7 @@ public struct Composer {
             }
         }
 
+        updateTypeRelationships(types: Array(unique.values))
         let array = Array(unique.values)
         DispatchQueue.concurrentPerform(iterations: array.count) { (counter) in
             let type = array[counter]
@@ -120,7 +121,7 @@ public struct Composer {
 
         Log.benchmark("\tresolution took \(currentTimestamp() - resolutionStart)")
 
-        updateTypeRelationships(types: Array(unique.values))
+        
         return (
             types: unique.values.sorted { $0.name < $1.name },
             functions: functions.sorted { $0.name < $1.name }
@@ -149,14 +150,14 @@ public struct Composer {
         if let array = parseArrayType(lookupName) {
             lookupName.array = array
             array.elementType = resolveTypeWithName(array.elementTypeName, array.elementType)
-            lookupName.generic = GenericType(name: lookupName.name, typeParameters: [
+            lookupName.generic = GenericType(name: "Array", typeParameters: [
                 GenericTypeParameter(typeName: array.elementTypeName, type: array.elementType)
                 ])
         } else if let dictionary = parseDictionaryType(lookupName) {
             lookupName.dictionary = dictionary
             dictionary.valueType = resolveTypeWithName(dictionary.valueTypeName, dictionary.valueType)
             dictionary.keyType = resolveTypeWithName(dictionary.keyTypeName, dictionary.keyType)
-            lookupName.generic = GenericType(name: lookupName.name, typeParameters: [
+            lookupName.generic = GenericType(name: "Dictionary", typeParameters: [
                 GenericTypeParameter(typeName: dictionary.keyTypeName, type: dictionary.keyType),
                 GenericTypeParameter(typeName: dictionary.valueTypeName, type: dictionary.valueType)
                 ])
--- a/SourceryFramework/Sources/Parsing/Composer.swift
+++ b/SourceryFramework/Sources/Parsing/Composer.swift
@@ -548,11 +548,11 @@ public struct Composer {
                     .map({ $0.trimmingCharacters(in: .whitespacesAndNewlines) })
 
                 if components.count == 1 {
-                    return MethodParameter(argumentLabel: nil, typeName: TypeName(components[0]))
+                    return MethodParameter(argumentLabel: nil, typeName: TypeName(components[0]), type: Type(name: components[0]))
                 } else {
                     let name = components[0].trimmingPrefix("_").stripped()
                     let typeName = components[1]
-                    return MethodParameter(argumentLabel: nil, name: name, typeName: TypeName(typeName))
+                    return MethodParameter(argumentLabel: nil, name: name, typeName: TypeName(typeName), type: Type(name: typeName))
                 }
             })
 
--- a/SourceryFramework/Sources/Parsing/Composer.swift
+++ b/SourceryFramework/Sources/Parsing/Composer.swift
@@ -396,7 +396,7 @@ public struct Composer {
         }
     }
 
-    private func specializeType(_ type: Type?, with genericParameters: [GenericTypeParameter]) -> Type? {
+    private static func specializeType(_ type: Type?, with genericParameters: [GenericTypeParameter]) -> Type? {
         guard let type = type else { return nil }
         // This hack with keyed archivers serves only one goal - copy type to specialize it with generic parameters
         // We don't want to change original type, it should stay unspecialized
--- a/SourceryFramework/Sources/Parsing/Composer.swift
+++ b/SourceryFramework/Sources/Parsing/Composer.swift
@@ -398,8 +398,10 @@ public struct Composer {
 
     private static func specializeType(_ type: Type?, with genericParameters: [GenericTypeParameter]) -> Type? {
         guard let type = type else { return nil }
+
         // This hack with keyed archivers serves only one goal - copy type to specialize it with generic parameters
         // We don't want to change original type, it should stay unspecialized
+        // TODO: Introduce our custom Copyable protocol and generate code for all subclasses of Type that would use plain init
         let data = NSKeyedArchiver.archivedData(withRootObject: type)
         let typeCopy = NSKeyedUnarchiver.unarchiveObject(with: data) as? Type
         typeCopy?.genericTypeParameters = genericParameters
--- a/SourceryFramework/Sources/Parsing/Composer.swift
+++ b/SourceryFramework/Sources/Parsing/Composer.swift
@@ -92,7 +92,7 @@ public struct Composer {
         DispatchQueue.concurrentPerform(iterations: array.count) { (counter) in
             let type = array[counter]
             type.typealiases.forEach { (_, alias) in
-                alias.type = resolveType(alias.typeName, alias.type, type)
+                alias.type = resolveType(alias.typeName, alias.type, type) ?? alias.type
             }
             type.variables.forEach {
                 resolveVariableTypes($0, of: type, resolve: resolveType)
@@ -142,7 +142,7 @@ public struct Composer {
         let lookupName = typeName.actualTypeName ?? typeName
 
         let resolveTypeWithName = { (typeName: TypeName, presumedType: Type?) -> Type? in
-            return self.resolveType(typeName: typeName, presumedType: presumedType, containingType: containingType, unique: unique, modules: modules, typealiases: typealiases)
+            return self.resolveType(typeName: typeName, presumedType: presumedType, containingType: containingType, unique: unique, modules: modules, typealiases: typealiases) ?? presumedType
         }
 
         // should we also set these types on lookupName?
@@ -201,19 +201,19 @@ public struct Composer {
     typealias TypeResolver = (TypeName, Type?, Type?) -> Type?
 
     private static func resolveVariableTypes(_ variable: Variable, of type: Type, resolve: TypeResolver) {
-        variable.type = resolve(variable.typeName, variable.type, type)
+        variable.type = resolve(variable.typeName, variable.type, type) ?? variable.type
 
         /// The actual `definedInType` is assigned in `uniqueTypes` but we still
         /// need to resolve the type to correctly parse typealiases
         /// @see https://github.com/krzysztofzablocki/Sourcery/pull/374
         if let definedInTypeName = variable.definedInTypeName {
-            _ = resolve(definedInTypeName, variable.type, type)
+            _ = resolve(definedInTypeName, variable.definedInType, type)
         }
     }
 
     private static func resolveSubscriptTypes(_ subscript: Subscript, of type: Type, resolve: TypeResolver) {
         `subscript`.parameters.forEach { (parameter) in
-            parameter.type = resolve(parameter.typeName, parameter.type, type)
+            parameter.type = resolve(parameter.typeName, parameter.type, type) ?? parameter.type
         }
 
         `subscript`.returnType = resolve(`subscript`.returnTypeName, `subscript`.returnType, type)
@@ -224,7 +224,7 @@ public struct Composer {
 
     private static func resolveMethodTypes(_ method: SourceryMethod, of type: Type?, resolve: TypeResolver) {
         method.parameters.forEach { parameter in
-            parameter.type = resolve(parameter.typeName, parameter.type, type)
+            parameter.type = resolve(parameter.typeName, parameter.type, type) ?? parameter.type
         }
 
         /// The actual `definedInType` is assigned in `uniqueTypes` but we still
@@ -254,7 +254,7 @@ public struct Composer {
     private static func resolveEnumTypes(_ enumeration: Enum, types: [String: Type], resolve: TypeResolver) {
         enumeration.cases.forEach { enumCase in
             enumCase.associatedValues.forEach { associatedValue in
-                associatedValue.type = resolve(associatedValue.typeName, associatedValue.type, enumeration)
+                associatedValue.type = resolve(associatedValue.typeName, associatedValue.type, enumeration) ?? associatedValue.type
             }
         }
 
--- a/SourceryFramework/Sources/Parsing/Composer.swift
+++ b/SourceryFramework/Sources/Parsing/Composer.swift
@@ -77,15 +77,22 @@ public struct Composer {
 
         let resolutionStart = currentTimestamp()
 
-        let resolveType = { (typeName: TypeName, containingType: Type?) -> Type? in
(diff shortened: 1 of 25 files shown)


SourceryFramework/Sources/Parsing/Composer.swift
```diff
@@ -156,44 +174,53 @@ public struct Composer {
         typeName.generic = lookupName.generic
 
         let resolvedTypeName = lookupName.generic?.name ?? lookupName.unwrappedTypeName
-        return unique[resolvedTypeName] ?? typeFromModule(resolvedTypeName, modules: modules)
+        
+        let type = unique[resolvedTypeName] ?? typeFromModule(resolvedTypeName, modules: modules)
+        if let knownPresumedType = presumedType, knownPresumedType.isConcreteGenericType {
+            let resolvedGenericTypeParameters = knownPresumedType.genericTypeParameters.map {
```
ilyapuchka: I'd maybe move this map to `specializeType` as it's part of specialisation process anyway

SourceryFramework/Sources/Parsing/Composer.swift
```diff
@@ -348,6 +375,16 @@ public struct Composer {
             return nil
         }
     }
+    
+    private func specializeType(_ type: Type?, with genericParameters: [GenericTypeParameter]) -> Type? {
+        guard let type = type else { return nil }
+        // This hack with keyed archivers serves only one goal - copy type to specialize it with generic parameters
+        // We don't want to change original type, it should stay unspecialized
+        let data = NSKeyedArchiver.archivedData(withRootObject: type)
+        let typeCopy = NSKeyedUnarchiver.unarchiveObject(with: data) as? Type
```
ilyapuchka: I guess this is done for brevity and not checking for all the possible types of `Type`, but this will probably come with a performance hit? Maybe would be better to introduce our custom Copyable protocol and generate code for all subclasses of `Type` that would use plain `init`.

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -152,7 +152,15 @@ public final class FileParser {
                 return nil
             }
 
-            type.isGeneric = isGeneric(source: source)
+            if let suffix = extract(.nameSuffixUpToBody, from: source) {
+                let declaration = extractGenericsDeclaration(source: suffix)
```
ilyapuchka:
```suggestion
let genericDeclaration = extractGenericsDeclaration(source: suffix)
```

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -152,7 +152,15 @@ public final class FileParser {
...
+                if !declaration.isEmpty {
```
ilyapuchka:
```suggestion
                if !genericDeclaration.isEmpty {
```

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -273,6 +281,21 @@ public final class FileParser {
 
 // MARK: - Details parsing
 extension FileParser {
+    
+    fileprivate func parseConcreteType(source: String) -> Type? {
+        guard !source.isEmpty else { return nil }
```
ilyapuchka:
```suggestion
        guard !source.isEmpty else { return nil }

```

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -273,6 +281,21 @@ public final class FileParser {
...
+        let declaration = extractGenericsDeclaration(source: source)
```
ilyapuchka:
```suggestion
        let genericDeclaration = extractGenericsDeclaration(source: source)
```

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -152,7 +152,15 @@ public final class FileParser {
                 return nil
             }
 
-            type.isGeneric = isGeneric(source: source)
+            if let suffix = extract(.nameSuffixUpToBody, from: source) {
+                let declaration = extractGenericsDeclaration(source: suffix)
+                if !declaration.isEmpty {
+                    // This type is generic, parse it's requirements
```
ilyapuchka:
```suggestion
```

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -273,6 +281,21 @@ public final class FileParser {
 
 // MARK: - Details parsing
 extension FileParser {
+    
+    fileprivate func parseConcreteType(source: String) -> Type? {
+        guard !source.isEmpty else { return nil }
+        let declaration = extractGenericsDeclaration(source: source)
+        if declaration.isEmpty {
+            // This is not generic type.
```
ilyapuchka:
```suggestion
```

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -273,6 +281,21 @@ public final class FileParser {
...
+            return Type(name: source)
+        }
+        let typeName = extractGenericTypeName(source: source)
+        guard declaration.contains("<") else {
```
ilyapuchka:
```suggestion
        guard genericDeclaration.contains("<") else {
```

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -273,6 +281,21 @@ public final class FileParser {
...
+            return Type(name: typeName,
+                        genericTypeParameters: extractGenericTypeParameters(source: declaration))
```
ilyapuchka:
```suggestion
                        genericTypeParameters: extractGenericTypeParameters(source: genericDeclaration))
```

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -318,6 +336,144 @@ extension FileParser {
 
 }
 
+// MARK: - Generics
+extension FileParser {
+    fileprivate func parseGenericPlaceholders(source: [String: SourceKitRepresentable], definedIn: Any?) {
+        guard let type = definedIn as? Type else { return }
+        guard let substructure = source[SwiftDocKey.substructure.rawValue] as? [[String: SourceKitRepresentable]] else { return }
+        substructure.forEach { dict in
```
ilyapuchka:
```suggestion

        substructure.forEach { dict in
```

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -318,6 +336,144 @@ extension FileParser {
...
+            if (dict[SwiftDocKey.kind.rawValue] as? String) == SwiftDeclarationKind.genericTypeParam.rawValue {
+                let name = dict[SwiftDocKey.name.rawValue] as? String ?? ""
+                let inheritedTypes = (dict[SwiftDocKey.inheritedtypes.rawValue] as? [[String: SourceKitRepresentable]])?.compactMap { $0[SwiftDocKey.name.rawValue] as? String } ?? []
+                var inheritedTypesParsed: [Type] = []
+                inheritedTypes.forEach { constraintString in
+                    let constraints = constraintString.components(separatedBy: "&")
+                    inheritedTypesParsed.append(contentsOf: constraints.map { recursivelyParseGenericPlaceholderDeclaration(source: $0) })
```
ilyapuchka:
```suggestion
                    inheritedTypesParsed.append(contentsOf: constraints.map { 
                        recursivelyParseGenericPlaceholderDeclaration(source: $0) 
                    })
```

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -318,6 +336,144 @@ extension FileParser {
...
+            .map { parseGenericTypePlaceholder(source: $0) }
+    }
+    
+    fileprivate func parseGenericTypePlaceholder(source: String) -> GenericTypePlaceholder {
+        if let constraintStart = source.index(of: ":") {
+            let name = String(source[source.startIndex ..< constraintStart]).trimmingCharacters(in: .whitespacesAndNewlines)
+            let constraints = String(source[source.index(after: constraintStart) ..< source.endIndex])
+                .components(separatedBy: "&")
+                .map { constraint in
+                    recursivelyParseGenericPlaceholderDeclaration(source: constraint)
```
ilyapuchka: do we need to do that for each type? I think it's only allowed to use concrete type as a first type of composition and everything else would be protocols, i.e. `ViewController<A> & Presentable & Dismissible`. Probably would be a bit faster if we only ever do it on first element.

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -318,6 +336,144 @@ extension FileParser {
...
+            return GenericTypePlaceholder(placeholderName: TypeName(name), constraints: constraints)
+        } else if source.contains("<") {
+            return GenericTypePlaceholder(placeholderName: TypeName(extractGenericTypeName(source: source)))
+        } else {
+            return GenericTypePlaceholder(placeholderName: TypeName(source.trimmingCharacters(in: .whitespacesAndNewlines)))
+        }
+    }
+    
+    fileprivate func findGeneric(withName name: String, in array: [GenericTypePlaceholder]) -> GenericTypePlaceholder? {
+        let filtered = array.filter { generic in generic.placeholderName.name == name }
```
ilyapuchka: `filter` and `first` can now be replaced with `first(where:)`

SourceryTests/Parsing/ComposerSpec.swift
```diff
@@ -396,6 +403,32 @@ class ParserComposerSpec: QuickSpec {
...
 
+                context("given generic inheritance") {
+                    it("resolves inherited type for subclasses") {
+                        let result = parse("""
+                            class Generic<T> {}
+                            class Specialized : Generic<Int> {}
+                        """)
+                        let generic = Class(name: "Generic", genericTypePlaceholders: [ GenericTypePlaceholder(placeholderName: TypeName("T")) ])
+                        let specialized = Class(name: "Specialized", inheritedTypes: ["Generic<Int>"])
+                        expect(result).to(equal([ generic, specialized]))
```
ilyapuchka:
```suggestion

                        expect(result).to(equal([generic, specialized]))
```

SourceryTests/Parsing/ComposerSpec.swift
```diff
@@ -396,6 +403,32 @@ class ParserComposerSpec: QuickSpec {
...
                 }
 
+                context("given generic inheritance") {
+                    it("resolves inherited type for subclasses") {
+                        let result = parse("""
+                            class Generic<T> {}
+                            class Specialized : Generic<Int> {}
+                        """)
+                        let generic = Class(name: "Generic", genericTypePlaceholders: [ GenericTypePlaceholder(placeholderName: TypeName("T")) ])
+                        let specialized = Class(name: "Specialized", inheritedTypes: ["Generic<Int>"])
```
ilyapuchka: Can we also validate that `inherits` property is populated as expected and contains `Generic` type? Or this is not the case?

SourceryTests/Parsing/ComposerSpec.swift
```diff
@@ -396,6 +403,32 @@ class ParserComposerSpec: QuickSpec {
...
+                            class Generic<T> {}
+                            class Specialized<U: Generic<Int>> {}
+                        """)
+                        let generic = Class(name: "Generic", genericTypePlaceholders: [ GenericTypePlaceholder(placeholderName: TypeName("T")) ])
+                        let specializedGeneric = Class(name: "Generic",
+                                                       genericTypePlaceholders: [GenericTypePlaceholder(placeholderName: TypeName("T"))],
+                                                       genericTypeParameters: [GenericTypeParameter(typeName: TypeName("Int"), type: Type(name: "Int"))])
+                        let placeholder = GenericTypePlaceholder(placeholderName: TypeName("U"), constraints: [specializedGeneric])
+                        let specialized = Class(name: "Specialized", genericTypePlaceholders: [placeholder])
+                        expect(result).to(equal([ generic, specialized]))
```
ilyapuchka:
```suggestion
                        expect(result).to(equal([generic, specialized]))
```

SourceryTests/Parsing/ComposerSpec.swift
```diff
@@ -396,6 +403,32 @@ class ParserComposerSpec: QuickSpec {
...
+                            class Specialized : Generic<Int> {}
+                        """)
+                        let generic = Class(name: "Generic", genericTypePlaceholders: [ GenericTypePlaceholder(placeholderName: TypeName("T")) ])
+                        let specialized = Class(name: "Specialized", inheritedTypes: ["Generic<Int>"])
+                        expect(result).to(equal([ generic, specialized]))
+                    }
+
+                    it("resolves inherited types for generic placeholders") {
+                        let result = parse("""
+                            class Generic<T> {}
```
ilyapuchka: can we add a test where there is a constraint here as well, i.e. `T: Bar` where `protocol Bar {}`

SourceryTests/Parsing/FileParser+SubscriptsSpec.swift
```diff
@@ -39,7 +39,7 @@ class FileParserSubscriptsSpec: QuickSpec {
                     expect(subscripts?[1]).to(equal(
                         Subscript(
                             parameters: [
-                                MethodParameter(argumentLabel: "b", name: "b", typeName: TypeName("Int"))
+                                MethodParameter(argumentLabel: "b", name: "b", typeName: TypeName("Int"), type: Type(name: "Int"))
```
ilyapuchka: now when we have standard library types inferred does it still work properly when we have an extension of this type. It used to be the only case when we would set `type`, can you make sure we have a test for that and it still works?

SourceryFramework/Sources/Parsing/Composer.swift
DenTelezhkin: The reason I did not move this to `specializeType` is that we would need to pass `resolveTypeWithName` there, which actually makes this method to not only specialize, but resolve types too. In current form `specializeType` does one thing and one thing only - takes existing type, and creates a new one with specified generic type parameters, and does not search for underlying types.

SourceryFramework/Sources/Parsing/Composer.swift
DenTelezhkin: I agree, but i think this would be a bigger change, that needs to be resolved in a separate pull request. I'm adding a TODO item to those lines to address your comment.

SourceryFramework/Sources/Parsing/FileParser.swift
DenTelezhkin: Unfortunately, we do. Unit test in FileParserSpec "extracts generics correctly" on line 270 shows specific example where this is needed.

SourceryTests/Parsing/ComposerSpec.swift
DenTelezhkin: Currently, this property is not populated. Reason is in `Composer.updateTypeRelationships(types:)` method `typesByName` contains "Generic" and "Specialized", however `type.inheritedTypes` contains "Generic<Int>", and therefore we cannot find "Generic", since we are searching for "Generic<Int>". 

I have a feeling it's a part of a bigger problem with generic names, but not sure how to resolve this one in scope of current pull request. Would love your advice on this.

SourceryTests/Parsing/FileParser+SubscriptsSpec.swift
DenTelezhkin: Sorry, I don't seem to fully understand what test do we need here. Can you expand on this a little bit?

DenTelezhkin:
Sorry for dragging this since forever. Was burned out on my daily job and couldn't force myself to finish this. If anyone wishes to continue this PR further in more reasonable time, please do that.

In the meantime, I rebased on master, and addressed code review comments, some of them fixed, some require additional discussion or separate PR.

SourceryTests/Parsing/FileParser+SubscriptsSpec.swift
ilyapuchka: I probably meant something like what if we have an extension of Int in the scanned code? Will this type reference contain all symbols from this extension? Meaning do we set this type property to the fully resolved type?

SourceryFramework/Sources/Parsing/Composer.swift
ilyapuchka: The problem is that I guess this may affect performance of processing source code for everyone even if generic types are not used after all. I can't tell for sure though, but we could check on the performance tests that that we should have? If there is a negative effect but we want to fix it separately maybe it will make sense to introduce "experimental" feature flag and do not parse generics unless it is enabled? Also considering that (I guess) generics support is still not complete for some things.

SourceryTests/Parsing/FileParser+SubscriptsSpec.swift
DenTelezhkin: Got it, expanded unit test a little bit to have an extension on Int, and it's set for subscripts in more places then before(even though Int is reported as extension, because it's not seen anywhere else).

SourceryFramework/Sources/Parsing/Composer.swift
DenTelezhkin: I tried to run perfomance tests in `Sourcery+PerformanceSpec`, but they seem to be broken in a lot of ways, not sure how to properly fix and run them.

I tried to see perfomance difference on one of my fairly large projects, but got hit with some issues with indirect enums, that crash Sourcery on my project when used. Unfortunately, I rebased on master, so all features currently in master are in Sourcery build i tried.

Experimental flag seems a good idea to me, since there are some other things that could be done with generics, however I have no idea how to properly implement a feature flag for Sourcery. 

if you can try to run perfomance tests on Sourcery and/or on one of bigger project you have that use Sourcery, that would be great.

DenTelezhkin:
Hi all!

Clearly, this PR did not get enough interest(even from me), so those improvements do not seem like anything worth pursuing. If somebody wants to continue working with those changes, you are welcome to do so, I will not delete fork in the foreseeable future, but I will not pursue this further. Thanks to everyone who reviewed and was interested in this PR.

Issue #759, opened 2019-05-15
Sourcery 0.16.X fails to generate cases


steverichey:
I've tried 0.16.0 via Homebrew, the 0.16.1 binary from the releases tab, and building from source (after adding some #available checks to allow Sourcery to compile). In all cases, templates generated empty cases inside an enum, e.g.:

    enum SomeUniqueEnum: String {
        case
        case
    }

And so on. I have one stencil file that generates enums for a number of types in my codebase to fulfill protocol conformance. I was eventually able to resolve this issue by reverting to 0.15.0. I am using Xcode 10.2.1 and macOS 10.14.4.

ilyapuchka:
@steverichey can you please submit a pull request with a test that fails to demonstrate this issue? or at least with a template code that results in it (not full template please, only relevant pieces)?

krzysztofzablocki:
this template is incorrect, you are mixing things up, you either:

1. wanted to use global `type.` subscript that is a sibling of `types.` but you overwrote it in your for loop so you are trying to access something that doesn't exist in this context
2. wanted to use containedTypes of your parent type, but forgot that part

Either way fix is either not overriding the global argument or simply using `type.containedType.DefinedCases.cases.` in the inner for loop

PR #777, opened 2019-07-22
Add test to demonstrate case generation bug, re #759

steverichey:
Per request from @ilyapuchka in #759, this demonstrates the issue where Sourcery will generate several empty case statements when only a few valid case statements are expected. I don't know enough about Sourcery to fix this issue myself, but I hope this helps resolve it.

Basically, I'm getting:

```swift
extension ExampleStruct {
    enum GeneratedCases {
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
        case
    }
}
```

When I expect:
```swift
extension ExampleStruct {
    enum GeneratedCases {
        case one
        case two
    }
}
```

### DIFF
--- a/SourceryTests/Generating/StencilTemplateSpec.swift
+++ b/SourceryTests/Generating/StencilTemplateSpec.swift
@@ -220,6 +220,29 @@ class StencilTemplateSpec: QuickSpec {
                 expect(result).to(equal(expectedResult))
             }
 
+            it("generates cases") {
+                var outputDir = Path("/tmp")
+                outputDir = Stubs.cleanTemporarySourceryDir()
+
+                let templatePath = Stubs.templateDirectory + Path("CaseGenerator.stencil")
+                let expectedResult = """
+                // Generated using Sourcery Major.Minor.Patch — https://github.com/krzysztofzablocki/Sourcery
+                // DO NOT EDIT
+
+                extension ExampleStruct {
+                    enum GeneratedCases {
+                        case one
+                        case two
+                    }
+                }
+                """
+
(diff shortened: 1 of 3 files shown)


Liquidsoul:
Hi @steverichey 
Thank you a lot for the time you spent on this!

I've had a look at your test and noticed that you tried to use subscripting to access the contained type `DefinedCases` and I don't think that this is currently supported 🤔
But maybe I'm wrong and this is something that should be fixed 😅 

In any cases, as a workaround, you could change your template to this:
```
{% for type in types.implementing.ShouldGenerateCases %}
extension {{type.name}} {
    enum GeneratedCases {
{% for containedType in type.containedTypes %}
{% if containedType.localName == "DefinedCases" %}
    {% for c in containedType.first.cases %}
    case {{c.name}}
    {% endfor %}
    }
{% endif %}
{% endfor %}
}

{% endfor %}
```

This would generate the cases the way you want.

ilyapuchka:
the subscript should work actually, maybe the name is wrong for the nested type, I don't remember if we add a parent type to it =/

ilyapuchka:
btw Stencil supports `where` in for loops pretty much the same way as in Swift

krzysztofzablocki:
this template is incorrect, you are mixing things up, you either:

1. wanted to use global `type.` subscript that is a sibling of `types.` but you overwrote it in your for loop so you are trying to access something that doesn't exist in this context
2. wanted to use containedTypes of your parent type, but forgot that part

Either way fix is either not overriding the global argument or simply using `type.containedType.DefinedCases.cases.` in the inner for loop

PR #779, opened 2019-08-30
Improve performance by parallelizing and changing cache logic

krzysztofzablocki:
1. Switches cache to rely on modified date instead of sha256
2. Parallelizes combine phase

Does 2. step seem safe? looking at the code I think it is as we only mutate unique table before I run multi-threaded logic and it only mutates a single type per thread so no locks are required

For NYT project this branch yields 5-10x speed improvement for codebase which is huge.

### DIFF
--- a/SourceryFramework/Sources/Parsing/Composer.swift
+++ b/SourceryFramework/Sources/Parsing/Composer.swift
@@ -74,7 +74,7 @@ public struct Composer {
             unique[current.name] = current
         }
 
-        let resolutionStart = CACurrentMediaTime()
+        let resolutionStart = currentTimestamp()
 
         let resolveType = { (typeName: TypeName, containingType: Type?) -> Type? in
             return self.resolveType(typeName: typeName, containingType: containingType, unique: unique, modules: modules, typealiases: typealiases)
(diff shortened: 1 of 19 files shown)


SourceryFramework/Sources/Parsing/Composer.swift
```diff
@@ -76,11 +74,15 @@ public struct Composer {
...
 
+        let resolutionStart = CACurrentMediaTime()
+
         let resolveType = { (typeName: TypeName, containingType: Type?) -> Type? in
             return self.resolveType(typeName: typeName, containingType: containingType, unique: unique, modules: modules, typealiases: typealiases)
         }
 
-        for (_, type) in unique {
+        let array = Array(unique.values)
+        DispatchQueue.concurrentPerform(iterations: array.count) { (counter) in
```
ilyapuchka: TIL

AliSoftware: 🤯

krzysztofzablocki: we have our own parallel map in the main Sourcery target for parallel `flatMap` / `map` but for just multi-threaded apply this is good.

PR #782, opened 2019-09-14
Sync .swift-version as 5.0

kateinoigakukun:
`Package.swift` indicates that we are using swift 5.0, but `.swift-version` is 4.2.


krzysztofzablocki:
thanks!

Issue #785, opened 2019-09-20
Using sourcery with --watch does not print errors thrown from Generator.generate


dimitar-lazarov:
What version of Swift are you using (swift --version)?
    - Swift version 5.0.1

What did you do?
    - run sourcery with `--watch` option

What did you expect to see?
    - When `Generator.generate` method throws an error, it should be printed in the console

What did you see instead?
    - no errors

This works fine without the `--watch` option. The error is silenced here:
https://github.com/krzysztofzablocki/Sourcery/blob/6770e4d58a7cc16a55c7efc882f2d9f07c553547/Sourcery/Sourcery.swift#L469

PR #786, opened 2019-09-21
Log errors when running with `--watch` option Re #785

dimitar-lazarov:
This is an improvement for better logging when running with `--watch` option. https://github.com/krzysztofzablocki/Sourcery/issues/785


krzysztofzablocki:
thanks for contributing!

PR #791, opened 2019-10-21
Fix SwiftPM build on case-sensitive file systems

glyuck:
Rename from `Sources/sourcery` to `Sources/Sourcery`.
It looks like developers just forgot to rename file in this commit: https://github.com/krzysztofzablocki/Sourcery/commit/ae9f98638e55de510cad9c1aa3e71aca21f864c1#diff-37ca2dd15ca0f6b1b49e78db084ef5b9

Here is the SwiftPM error: `error: Source files for target Sourcery should be located under 'Sources/Sourcery', or a custom sources path can be set with the 'path' property in Package.swift`


krzysztofzablocki:
thanks for contributing

PR #794, opened 2019-10-30
HUGE performance boost to 'parse' when have a large complex 'excludes' list.

mishagray:
Removed the recursive computation of the excludeList from the outside the enumeration().

The old logic would recompute the entire exclusion path list for every file being checked in the path.
Now the exclusion list is computed first, and then it's reused for the enumeration.

Also found a definite improvement in using a 'Set' over a simple 'array'.

This results in a HUGE runtime performance for repos with large number of excluded files (we saw 95 seconds runtime -> 3.5 seconds).


krzysztofzablocki:
makes total sense, thanks for contributing!

PR #800, opened 2019-12-18
Strip out default values from type names in AssociatedValues

chrisfsampaio:
When parsing the following Enum:
```swift
enum Foo {
    case bar(Int = 1)
}
```
`typeName.name` in the `AssociatedValue` is set to `Int = 1`, which I'm assuming is semantically incorrect as the default value has nothing to do with a type name.

We can store the default value somewhere else (maybe another property of `AssociatedValue`?) in another PR if needs be.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 - Added `optional` filter for variables
 - Added `json` filter to output raw JSON objects 
+- Added `.defaultValue` to `AssociatedValue`
 
 ### Internal Changes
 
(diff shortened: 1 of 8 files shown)


krzysztofzablocki:
sweet, thanks for contributing!

chrisfsampaio:
@krzysztofzablocki no problem! Will update 👍

chrisfsampaio:
@krzysztofzablocki Updated!

krzysztofzablocki:
great job!

Issue #806, opened 2020-02-07
[Feature Request] Support for Free Functions


maxmamis:
Currently, it appears Sourcery has no support for free functions (ie, those not contained as members of another type). I have a use-case that requires this functionality. I've started looking into what it would take to add support, but before I go too far I'd like to ask, has any work or thought already been put into this idea? 

Obviously any `types.inherits` or `types.based` workflows would be impossible, but querying based on source annotations theoretically should be able to work.

maxmamis:
I spent some time digging a bit further into the implementation. Without much effort I was able to parse free functions as `SourceryMethod` instances and expose them to templates as a top-level parameter `functions`. I was unable to add them to `types`, as `SourceryMethod` does not conform to `Type`, and it's unclear to me if that would be semantically correct anyway (since a function is not really a type).

At this point, I feel like I may be able to accomplish my goals using a custom Sourcery build, but I would love to clean up this code and contribute it back to the project - but I would like some architectural input from the maintainers before I start 😄

PR #807, opened 2020-02-07
Add support for free functions

maxmamis:
As mentioned in https://github.com/krzysztofzablocki/Sourcery/issues/806, I have a use-case that requires working with free functions. I decided to see what would be involved in making this work and it turns out it's not too bad. 

This is a naive implementation meant as an aid for discussion, it's not production-ready yet. Notably I didn't add tests yet, or test support for templating languages other than Stencil.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Added `optional` filter for variables
 - Added `json` filter to output raw JSON objects 
 - Added `.defaultValue` to `AssociatedValue`
+- Added support for parsing free functions
 
 ### Internal Changes
 
--- a/Sourcery/Sourcery.swift
+++ b/Sourcery/Sourcery.swift
@@ -308,11 +308,11 @@ extension Sourcery {
         let uniqueTypeStart = currentTimestamp()
 
         //! All files have been scanned, time to join extensions with base class
-        let types = Composer.uniqueTypes(parserResult)
+        let (types, functions) = Composer.uniqueTypesAndFunctions(parserResult)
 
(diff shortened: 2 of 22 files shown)


krzysztofzablocki:
@maxmamis this looks great and honestly that's where I'd put those functions, as top level object and sibling of `types`, is there anything missing here functionality-wise, other than proper tests ?

maxmamis:
I believe this is complete, though I haven't tested it with JS templates yet. I also was seeing an issue where I didn't have full metadata for return types, but I'm not sure whether that's a separate issue

maxmamis:
thanks for the feedback! I'll add/fix the tests as soon as I get a chance

krzysztofzablocki:
@maxmamis if metadata is missing that would suggest to me you need to modify the unifying phase in `Composer.swift`, shouldn't be complicated as we already do this for normal types

maxmamis:
Thanks, I'll take a look there

maxmamis:
@krzysztofzablocki Updated `Composer.swift` to fix the issue I was seeing and added a few tests for free functions — wasn't sure how far down the rabbit hole I should go (should I test that it parses every kind of parameter and return type properly?), please let me know if you'd like to see some more. Otherwise ready for a full review!

krzysztofzablocki:
@maxmamis this looks good, maybe you could write a very simple template for EJS / Swift and add a test to each to prove that functions are processed correctly in those scenarios? the template could simply read the free function and generate another one from it

maxmamis:
@krzysztofzablocki Added!

krzysztofzablocki:
great job, thanks for contributing 👏

PR #808, opened 2020-02-22
Add support for protocol compositions

### LINKED ISSUE #541: Deconstruct types composition
When using markers like `typealias AutoCodable = AutoDecodable & AutoEncodable`, type that implements `AutoCodable` should also contain `AutoDecodable` and `AutoEncodable` in its `based` and `implements` collections. Currently it only contains `AutoCodable`.

erichoracek:
Fixes part of #541 (for named protocol compositions declared as a typealias) by adding a new `ProtocolComposition` `Type` subclass as described [here](https://github.com/krzysztofzablocki/Sourcery/issues/541#issuecomment-373544710). 

The reason that this type is not named `TypeComposition` is that they are called "Protocol Compositions" in the [Swift reference manual](https://docs.swift.org/swift-book/ReferenceManual/Types.html#ID454).

A `ProtocolComposition` has an array of `composedTypes`, which reference each of the types composed between ampersands in its declaration. The `kind` of a `ProtocolComposition` is `"protocolComposition"`. 

`ProtocolCompositions` are also made available on `Types` as `protocolCompositions`.


Rakefile
```diff
@@ -18,11 +18,11 @@ def version_select
 end
 
 def xcpretty(cmd)
-  if `which xcpretty` && $?.success?
-    sh "set -o pipefail && #{cmd} | xcpretty -c"
-  else
+  # if `which xcpretty` && $?.success?
```
krzysztofzablocki: forgot to clean this out?

erichoracek: @krzysztofzablocki I was trying to get to the root cause of the failing tests—looks like I was missing an `import Foundation` in `ProtocolComposition.swift` which was causing the SwiftTemplate tests to fail to compile. Should be fixed now!

PR #813, opened 2020-03-10
Add support for indirect enum cases

jacobdweightman:


SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -647,6 +647,11 @@ extension FileParser {
             return nil
         }
 
+        var isIndirect: Bool = false
+        if let attributes = source["key.attributes"] as? [[String: SourceKitRepresentable]], let attribute = attributes[0]["key.attribute"] as? String, SwiftDeclarationAttributeKind(rawValue: attribute) == .indirect {
```
jacobdweightman: I have to admit this isn't super clear code, but I couldn't see what the preferred way of digging into these `SourceKitRepresentable`s is so I thought I would call it out here. What might be a cleaner way to extract the attribute from `source`?

krzysztofzablocki: I think this is fine for this field, but shouldn't you guard against empty array? either use `first` or add a check before indexing

jacobdweightman: Good catch! I've amended to use `.first?` in place of `[0]`.

krzysztofzablocki:
thanks for contributing :)

jacobdweightman:
@krzysztofzablocki thanks for your feedback and quick action on this MR! I see that there hasn't been a release since September of last year—do you think it would be possible to make a release with this feature soon?

PR #814, opened 2020-03-14
Add support for accessing all typealiases, parse global typealiases

### LINKED ISSUE #235: Make typealiases accessible for types and in global scope
With typealiases becoming more stable we can make them accessible in templates. This will allow some nice tricks like using typealias value instead of annotations to generate typed things.
`Type` should provide typealiases mapped to their names via`typealias` property and array of typealiases for iteration via `typealiases` property. The same for global typealiases.

### LINKED ISSUE #765: Global typealias types are ignored, unless extended
If I have a Swift code like this:

```
typealias UserResult = Result<User, AccountError>
```

this type `UserResult` does not appear in `types.all`. 
If I do something like this:

```
extension UserResult {
...
}
```

then sourcery will add it into `types.all` but `type.name` will be `Result<User, AccountError>` instead of `UserResult` what I expected.

I read through many issues mentioning `typealias` here, but it's not clear to me is this correct behavior or a bug?

erichoracek:
Supports accessing all typealiases from `TemplateContext`, including both global and local typealiases as well as typealiases that point at other typealiases.

Added to `TemplateContext` as both `typesaliasesByName` and `typealiases`.

Fixes #765
Fixes #235

### DIFF
--- a/SourceryFramework/Sources/Parsing/Composer.swift
+++ b/SourceryFramework/Sources/Parsing/Composer.swift
@@ -17,15 +17,15 @@ public struct Composer {
     ///
     /// - Parameter parserResult: Result of parsing source code.
     /// - Returns: Final types and extensions of unknown types.
-    public static func uniqueTypesAndFunctions(_ parserResult: FileParserResult) -> (types: [Type], functions: [SourceryMethod]) {
+    public static func uniqueTypesAndFunctions(_ parserResult: FileParserResult) -> (types: [Type], functions: [SourceryMethod], typealiases: [Typealias]) {
         var unique = [String: Type]()
         var modules = [String: [String: Type]]()
-        let types = parserResult.types
+        let parsedTypes = parserResult.types
         let functions = parserResult.functions
-        let typealiases = self.typealiasesByNames(parserResult)
+        let (resolvedTypealiases, unresolvedTypealiases) = self.typealiases(parserResult)
 
         //set definedInType for all methods and variables
-        types
+        parsedTypes
             .forEach { type in
                 type.variables.forEach { $0.definedInType = type }
                 type.methods.forEach { $0.definedInType = type }
@@ -33,7 +33,7 @@ public struct Composer {
         }
 
         //map all known types to their names
-        types
+        parsedTypes
             .filter { $0.isExtension == false}
             .forEach {
                 unique[$0.name] = $0
@@ -46,15 +46,15 @@ public struct Composer {
 
         //replace extensions for type aliases with original types
         //extract all methods and variables from extensions
-        types
+        parsedTypes
             .filter { $0.isExtension == true }
             .forEach {
-                $0.localName = actualTypeName(for: TypeName($0.name), typealiases: typealiases) ?? $0.localName
+                $0.localName = actualTypeName(for: TypeName($0.name), typealiases: resolvedTypealiases) ?? $0.localName
             }
 
         //extend all types with their extensions
-        types.forEach { type in
-            type.inheritedTypes = type.inheritedTypes.map { actualTypeName(for: TypeName($0), typealiases: typealiases) ?? $0 }
+        parsedTypes.forEach { type in
+            type.inheritedTypes = type.inheritedTypes.map { actualTypeName(for: TypeName($0), typealiases: resolvedTypealiases) ?? $0 }
 
             let uniqueType = unique[type.name] ?? typeFromModule(type.name, modules: modules)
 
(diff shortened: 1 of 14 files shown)


SourceryRuntime/Sources/Enum.swift
```diff
@@ -111,7 +111,7 @@ import Foundation
             self.rawValue = aDecoder.decode(forKey: "rawValue")
             guard let associatedValues: [AssociatedValue] = aDecoder.decode(forKey: "associatedValues") else { NSException.raise(NSExceptionName.parseErrorException, format: "Key '%@' not found.", arguments: getVaList(["associatedValues"])); fatalError() }; self.associatedValues = associatedValues
             guard let annotations: [String: NSObject] = aDecoder.decode(forKey: "annotations") else { NSException.raise(NSExceptionName.parseErrorException, format: "Key '%@' not found.", arguments: getVaList(["annotations"])); fatalError() }; self.annotations = annotations
-            guard let indirect: Bool = aDecoder.decode(forKey: "indirect") else { NSException.raise(NSExceptionName.parseErrorException, format: "Key '%@' not found.", arguments: getVaList(["indirect"])); fatalError() }; self.indirect = indirect
+            self.indirect = aDecoder.decode(forKey: "indirect")
```
erichoracek: This change occurred when I ran `rake generate_internal_boilerplate_code` — maybe #813 didn't use this command and instead made this change by hand cc @jacobdweightman ?

SourcerySwift/Sources/SourceryRuntime.content.generated.swift
```diff
@@ -1431,7 +1432,7 @@ import Foundation
             self.rawValue = aDecoder.decode(forKey: "rawValue")
             guard let associatedValues: [AssociatedValue] = aDecoder.decode(forKey: "associatedValues") else { NSException.raise(NSExceptionName.parseErrorException, format: "Key '%@' not found.", arguments: getVaList(["associatedValues"])); fatalError() }; self.associatedValues = associatedValues
             guard let annotations: [String: NSObject] = aDecoder.decode(forKey: "annotations") else { NSException.raise(NSExceptionName.parseErrorException, format: "Key '%@' not found.", arguments: getVaList(["annotations"])); fatalError() }; self.annotations = annotations
-            guard let indirect: Bool = aDecoder.decode(forKey: "indirect") else { NSException.raise(NSExceptionName.parseErrorException, format: "Key '%@' not found.", arguments: getVaList(["indirect"])); fatalError() }; self.indirect = indirect
+            self.indirect = aDecoder.decode(forKey: "indirect")
```
erichoracek: same here re: running `rake generate_internal_boilerplate_code`

SourceryFramework/Sources/Parsing/Composer.swift
```diff
@@ -78,15 +78,12 @@ public struct Composer {
...
-            return self.resolveType(typeName: typeName, containingType: containingType, unique: unique, modules: modules, typealiases: typealiases)
+            return self.resolveType(typeName: typeName, containingType: containingType, unique: unique, modules: modules, typealiases: resolvedTypealiases)
         }
 
-        let array = Array(unique.values)
-        DispatchQueue.concurrentPerform(iterations: array.count) { (counter) in
-            let type = array[counter]
-            type.typealiases.forEach { (_, alias) in
-                alias.type = resolveType(alias.typeName, type)
-            }
```
erichoracek: This is no longer necessary as it's now handled by https://github.com/krzysztofzablocki/Sourcery/pull/814/files#diff-3b9e6e3e3ed536fd38045c732a09f562R111-R115

SourceryRuntime/Sources/Enum.swift
jacobdweightman: Yes, my mistake. I couldn't figure out how to make it regenerate this code, so I updated it by hand and assumed I wrote it correctly when CI passed. I'll keep this in mind and remember to run that `rake` command when contributing in the future, though perhaps it would be good to document that this should be done and/or make it a CI requirement.

I'd be more than happy to open up a follow-up PR to make this change—what do you think?

SourceryTests/Generating/SwiftTemplateSpecs.swift
```diff
@@ -153,7 +153,7 @@ class SwiftTemplateTests: QuickSpec {
                     try Generator.generate(Types(types: []), functions: [], template: SwiftTemplate(path: templatePath))
                     }
                     .to(throwError(closure: { (error) in
-                        expect("\(error)").to(contain("\(templatePath): Fatal error: Index out of range\n"))
+                        expect("\(error)").to(contain("\(templatePath): Fatal error: Index out of range"))
```
erichoracek: This was failing locally on my machine due there being a filename following this message, so I removed the newline requirement from the test

erichoracek:
@krzysztofzablocki could I grab a review on this? We have a use case for this at Airbnb in combination with #808 — hoping to get this into a release so we don't have to fork

krzysztofzablocki:
thanks for contributing!

PR #824, opened 2020-04-28
Support building on Xcode 11-4-1

erichoracek:
So that we can cut a new release.


SourceryTests/Parsing/ComposerSpec.swift
```diff
@@ -80,6 +80,7 @@ class ParserComposerSpec: QuickSpec {
                                                                               typeName: TypeName("String"),
                                                                               defaultValue: "\"Baz\"")],
                                                  returnTypeName: TypeName("Void"),
+                                                 accessLevel: .none,
```
erichoracek, line 83: These changes were required to get the tests to pass, it appears that SourceKit no longer reports access levels for extensions the same way that it did previously

PR #839, opened 2020-06-01
[823] Optional / Closure recognition

### LINKED ISSUE #823: TypeName.isOptional returns false positive for closures that have optional return type.
Consider this scenario:
```swift
protocol TestProtocol {
    var closure: (URL, URL) -> Bool? { get }
}
```

Sourcery for that type returns `isOptional` as `true`, even though it's not. It seems like the problem lies in this line:
https://github.com/krzysztofzablocki/Sourcery/blob/2ff5e001c68f86cdef21bdc76f6d9ccf505fffe0/SourceryRuntime/Sources/TypeName.swift#L74

since it has the suffix `?`. Currently in my templates I'm using this workaround:
```swift
func isOptional(variable: Variable) -> Bool {
    if variable.typeName.isClosure {
        return variable.typeName.name.hasSuffix(")?")
    } else {
        return variable.isOptional
    }
}
```

wasn't sure if I need to care about IUO anymore so just improved the suffix in case it's a closure. But I'm wondering about better solution for that, perhaps inside the Sourcery core?

krzysztofzablocki:
I think this should fix #823

### DIFF
--- a/SourceryRuntime/Sources/TypeName.swift
+++ b/SourceryRuntime/Sources/TypeName.swift
@@ -70,10 +70,16 @@ public protocol Typed {
         } else {
             name = name.bracketsBalancing()
             name = name.trimmingPrefix("inout ").trimmingCharacters(in: .whitespacesAndNewlines)
-            let isImplicitlyUnwrappedOptional = name.hasSuffix("!") || name.hasPrefix("ImplicitlyUnwrappedOptional<")
-            let isOptional = name.hasSuffix("?") || name.hasPrefix("Optional<") || isImplicitlyUnwrappedOptional
-            self.isImplicitlyUnwrappedOptional = isImplicitlyUnwrappedOptional
-            self.isOptional = isOptional
+
+            if name.isValidClosureName() {
+                let isImplicitlyUnwrappedOptional = name.hasPrefix("ImplicitlyUnwrappedOptional<") && name.hasSuffix(">")
+                self.isImplicitlyUnwrappedOptional = isImplicitlyUnwrappedOptional
+                self.isOptional = (name.hasPrefix("Optional<")  && name.hasSuffix(">")) || isImplicitlyUnwrappedOptional
+            } else {
+                let isImplicitlyUnwrappedOptional = name.hasSuffix("!") || name.hasPrefix("ImplicitlyUnwrappedOptional<")
+                self.isImplicitlyUnwrappedOptional = isImplicitlyUnwrappedOptional
+                self.isOptional = name.hasSuffix("?") || name.hasPrefix("Optional<") || isImplicitlyUnwrappedOptional
+            }
 
             var unwrappedTypeName: String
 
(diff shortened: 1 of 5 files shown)


krzysztofzablocki:
mind taking a look whether the logic makes sense @sunshinejr ?

SourceryTests/Models/TypeNameSpec.swift
```diff
@@ -113,6 +113,20 @@ class TypeNameSpec: QuickSpec {
                     expect(TypeName("((Int) -> (Int)) -> ()").isClosure).to(beTrue())
                     expect(TypeName("((Int, Int) -> (), Int)").isClosure).to(beFalse())
                 }
+
+                it("reports optional status correctly") {
+                    expect(TypeName("() -> ()").isOptional).to(beFalse())
+                    expect(TypeName("() -> ()?").isOptional).to(beFalse())
+                    expect(TypeName("() -> ()!").isImplicitlyUnwrappedOptional).to(beFalse())
+                    expect(TypeName("(() -> ()!)").isImplicitlyUnwrappedOptional).to(beFalse())
```
sunshinejr, line 121: how about adding one more for `isOptional`:
```swift
expect(TypeName("(() -> ()?)").isOptional).to(beFalse())
```

SourceryRuntime/Sources/TypeName.swift
```diff
@@ -70,10 +70,16 @@ public protocol Typed {
...
             name = name.trimmingPrefix("inout ").trimmingCharacters(in: .whitespacesAndNewlines)
-            let isImplicitlyUnwrappedOptional = name.hasSuffix("!") || name.hasPrefix("ImplicitlyUnwrappedOptional<")
-            let isOptional = name.hasSuffix("?") || name.hasPrefix("Optional<") || isImplicitlyUnwrappedOptional
-            self.isImplicitlyUnwrappedOptional = isImplicitlyUnwrappedOptional
-            self.isOptional = isOptional
+
+            if name.isValidClosureName() {
+                let isImplicitlyUnwrappedOptional = name.hasPrefix("ImplicitlyUnwrappedOptional<") && name.hasSuffix(">")
+                self.isImplicitlyUnwrappedOptional = isImplicitlyUnwrappedOptional
+                self.isOptional = (name.hasPrefix("Optional<")  && name.hasSuffix(">")) || isImplicitlyUnwrappedOptional
```
ilyapuchka, line 77: I think we should change the doc for this property since it now will return `true` for `!` types, not just `?` types, when before they were mutually exclusive.

SourceryRuntime/Sources/TypeName.swift
```diff
@@ -70,10 +70,16 @@ public protocol Typed {
         } else {
             name = name.bracketsBalancing()
             name = name.trimmingPrefix("inout ").trimmingCharacters(in: .whitespacesAndNewlines)
-            let isImplicitlyUnwrappedOptional = name.hasSuffix("!") || name.hasPrefix("ImplicitlyUnwrappedOptional<")
-            let isOptional = name.hasSuffix("?") || name.hasPrefix("Optional<") || isImplicitlyUnwrappedOptional
-            self.isImplicitlyUnwrappedOptional = isImplicitlyUnwrappedOptional
-            self.isOptional = isOptional
+
+            if name.isValidClosureName() {
```
ilyapuchka, line 74: how does this branch handles `(...)?` and `(...)!` cases?

SourceryRuntime/Sources/TypeName.swift
krzysztofzablocki: it worked same way before, you can see that in diff it had the `|| isImplicitlyUnwrappedOptional`

SourceryRuntime/Sources/TypeName.swift
krzysztofzablocki: covered by tests I added, reports correctly

Issue #835, opened 2020-05-15
Not able to render AutoCodable.swifttemplate


Yannmm:
Xcode 11.4.1, not sure where goes wrong, since there is not error message.

```
Using configuration file at './sourcery.yml'
Scanning sources...
	loadOrParse: 0.0028589963912963867
	resolution took 0.0006489753723144531
	combiningTypes: 0.0008379220962524414
	total: 0.0037360191345214844
Found 2 types.
Loading templates...
Loaded 1 templates.
	Loading took 0.005620002746582031
Generating code...
Rendering template /Users/ryan/Desktop/WhatIsSourcery/Templates/AutoCodable.swifttemplate
/usr/bin/env "xcrun" "swift" "build" "-Xswiftc" "-Onone" "-Xswiftc" "-suppress-warnings" "--disable-sandbox"
error: 
```

nangadiquotient:
Facing the same error here too in Xcode 11.4.1
Loading templates...
Loaded 5 templates.
	Loading took 0.04272198677062988
Generating code...
Rendering template /Users/nangadi/coupons-ios/couponsapp/Coupons/Coupons_Framework/Sourcery/Templates/AutoCodable.swifttemplate
/usr/bin/env "xcrun" "swift" "build" "-Xswiftc" "-Onone" "-Xswiftc" "-suppress-warnings" "--disable-sandbox"
error:

RenGate:
Most likely this is caused by the same issue mentioned in https://github.com/krzysztofzablocki/Sourcery/issues/817

Issue #817, opened 2020-03-25
Failed to build project with xcode 11.4


bogdangolubcr:
Hi, when i tried to build my project with new xcode 11.4 I had this issue with no error description.
```
Using configuration file at '.sourcery.platform.yml'
Scanning sources...
	loadOrParse: 0.8185429573059082
	resolution took 0.1482919454574585
	combiningTypes: 0.16124391555786133
	total: 0.9820070266723633
Found 213 types.
Loading templates...
Loaded 6 templates.
	Loading took 0.008983016014099121
Generating code...
error: 
```

hartbit:
Same here.

nicholaslythall:
I was having the same issue. If I ran Sourcery via the terminal it worked fine, but when running as a build phase it failed with `error:`

For some reason, I had the `--disableCache` flag set on the build phase script, and it worked after I removed that, but only if I ran it at least once via CLI before trying to build.

nicholaslythall:
I was able to find a temporary workaround by calling `unset SDKROOT` in the build phase script before calling sourcery.
The value for `SDKROOT` before being unset was `/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator13.4.sdk`

ilyapuchka:
Does this happen with any kind of templates or might be related just to Swift templates?

RenGate:
@ilyapuchka In my case this error is only present when swifttemplate is used. What's interesting is that even if the swifttemplate itself is empty, the error is still there. I'm using Sourcery 0.17.0.

hartbit:
For what it's worth, I also have a Swift template.

vCrespoP:
Bumping this issue. 

It's also my case with Sourcery 0.18.0 although it compiles and generates code, but throws an error:

> Command PhaseScriptExecution emitted errors but did not return a nonzero exit code to indicate failure

jonnokim17:
i'm getting the same result as @vCrespoP. do we have a fix for this yet? thanks!

krzysztofzablocki:
@vCrespoP @jonnokim17 does clearing cache change anything?

xploit42:
Same problem here, clearing the cache doesn't change anything

mgray88:
FWIW, I experienced the same issue on 11.4.1, however setting `--disableCache` on the build phase script actually eliminated the error. Subsequently removing it however, caused the error to come back

[...]

PR #849, opened 2020-08-13
Fix error at build time when using SwiftTemplate & Xcode 11.4 or higher

DavidBertet:
Fix for the issues #817 / #835

**Problem**

We obtain a weird empty error during the build when the following criteria are met
- using Xcode 11.4 and higher
- using a SwiftTemplate
- using a build phase to run Sourcery

image

**It works when we run the exact same command from a terminal**

**Solution**

As we can see, the latest command to run is "xcrun swift ..."
This command is used at only one place in Sourcery, in SwiftTemplate.swift!

By looking at that code, if an error occurred during this command execution, we throw the `output` instead of the `error`. 
```
if compilationResult.exitCode != 0 || !compilationResult.error.isEmpty {
    throw compilationResult.output
}
```

That's why we have an empty error in Xcode. by printing the error instead, we have more information

image

It looks like it is trying to generate the template by using the SDK the project is using instead of MacOSX

A quick print of the SDK path using `xcrun --show-sdk-path` confirms it isn't using the correct one
`/Applications/Xcode/.../Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator13.6.sdk`
The same command in a terminal shows it is using MacOSX. **That explains why it works in the terminal**
`/Applications/Xcode/.../Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk`

By forcing the SDK to MacOSX using the option `-sdk macosx`, it works perfectly 🎉

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Add missing documentation for the `ProtocolComposition` type.
 - Fix incorrectly taking closure optional return value as sign that whole variable is optional (#823) 
 - Fix incorrectly taking return values with closure as generic type as sign that whole variable is a closure (#845)
+- Fix empty error at build time when using SwiftTemplate on Xcode 11.4 and higher (#817)
 
 ## 0.18.0
 
--- a/SourcerySwift/Sources/SwiftTemplate.swift
+++ b/SourcerySwift/Sources/SwiftTemplate.swift
@@ -229,6 +229,7 @@ open class SwiftTemplate {
 
         let arguments = [
             "xcrun",
+            "--sdk", "macosx",
             "swift",
             "build",
             "-Xswiftc", "-Onone",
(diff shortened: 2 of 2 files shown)


ilyapuchka:
some tests broke, but nice fix, thanks!

csykes:
Thanks David! Super appreciated.

DavidBertet:
> some tests broke, but nice fix, thanks!

My bad, I ran the wrong test suite locally.

I don't know how you want to manage that though. Both `output` and `error` are useful here.
I did a simple concatenation of both if not empty. Adding a prefix `output:`/`error:` doesn't seemed necessary to me as they rarely happen together, and that wouldn't add any meaning.

Otherwise I don't know how to write a new test for that error case as it would require to make `xcrun` fail with an error.

Let me know if you have any feedback

krzysztofzablocki:
@DavidBertet Thank you so much for contributing this 🙇 We can expand tests later on if need be, I think this is good for now 👍

Issue #842, opened 2020-07-01
AutoEquatable template can yield a warning in Xcode 11.4+


marciniwanicki:
Extensions generated with https://github.com/krzysztofzablocki/Sourcery/blob/master/Templates/Templates/AutoEquatable.stencil template can yield the following warning:

```
AutoEquatable.generated.swift: warning: enum case '<case name>' has <n> associated values; matching them as a tuple is deprecated
```

It's probably due to the fact Xcode 11.4 is shipped with Swift 5.2. Checked in Xcode 12 beta1, and looks that the same warning is triggered.

**Example:**

```swift
enum MyEnum {
  case identifier(first: String, second: String)
}
```

the following switch statement will produce the warning.

```swift
switch value {
  case .identifier(let tuple): // warning
      // tuple.first, tuple.second
}
```

In the Swift 5.3 online book, the associated values are explicitly listed and it's probably "the recommended" way (https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html), i.e.

```swift
switch value {
  case .identifier(let first, let second): // no warning
      // tuple.first, tuple.second
}

```
or 

```swift
switch value {
  case let .identifier(first, second): // no warning
      // tuple.first, tuple.second
}
```

Using `()` to enforce tuple might also work, but it would be probably better to just list the values.

I think the fix should be quite straight-forward and backward compatible.

art-divin:
Closing since it was implemented in #868

PR #868, opened 2020-11-12
Tuple for associated values in enum case is deprecated

DavidBertet:
Code generated with AutoEquatable and AutoHashable templates is raising a deprecation warning

Using a tuple to access associated values in the enum case is now deprecated since Swift 5.2

Update the 2 templates accordingly to put the exhaustive list of associated values, and avoid the warning

#842


krzysztofzablocki:
thanks for contributing🙇

PR #870, opened 2020-11-21
Fix parsing method parameters without names, e.g. func x(_: ParameterWithNoName)

artyom-razinov:
This:
`func myMethod(_: Int)`

was parsed like this:
`func myMethod()`

Parameters without names were just ignored as if they didn't exist.

Feel free to ask me to improve the code if it doesn't fit codestyle of the project.

### DIFF
--- a/SourceryFramework/Sources/Parsing/FileParser.swift
+++ b/SourceryFramework/Sources/Parsing/FileParser.swift
@@ -125,45 +125,46 @@ public final class FileParser {
         var typealiases = [Typealias]()
         try walkDeclarations(source: source) { kind, name, access, inheritedTypes, source, definedIn, next in
             let type: Type
-            switch kind {
-            case .protocol:
+            switch (kind, name) {
+            case let (.protocol, name?):
                 type = Protocol(name: name, accessLevel: access, isExtension: false, inheritedTypes: inheritedTypes)
-            case .class:
+            case let (.class, name?):
                 type = Class(name: name, accessLevel: access, isExtension: false, inheritedTypes: inheritedTypes)
-            case .struct:
+            case let (.struct, name?):
                 type = Struct(name: name, accessLevel: access, isExtension: false, inheritedTypes: inheritedTypes)
-            case .enum:
+            case let (.enum, name?):
                 type = Enum(name: name, accessLevel: access, isExtension: false, inheritedTypes: inheritedTypes)
-            case .extension,
-                 .extensionClass,
-                 .extensionStruct,
-                 .extensionEnum:
+            case let (.extension, name?),
+                 let (.extensionClass, name?),
+                 let (.extensionStruct, name?),
+                 let (.extensionEnum, name?):
                 type = Type(name: name, accessLevel: access, isExtension: true, inheritedTypes: inheritedTypes)
-            case .enumelement:
+            case (.enumelement, _):
                 return parseEnumCase(source)
-            case .varInstance:
+            case (.varInstance, _):
                 return parseVariable(source, definedIn: definedIn as? Type)
-            case .varStatic, .varClass:
+            case (.varStatic, _),
+                 (.varClass, _):
                 return parseVariable(source, definedIn: definedIn as? Type, isStatic: true)
-            case .varLocal:
+            case (.varLocal, _):
                 //! Don't log local / param vars
                 return nil
-            case .functionMethodClass,
-                 .functionMethodInstance,
-                 .functionMethodStatic:
+            case (.functionMethodClass, _),
+                 (.functionMethodInstance, _),
+                 (.functionMethodStatic, _):
                 return parseMethod(source, definedIn: definedIn as? Type, nextStructure: next)
-            case .functionFree:
+            case (.functionFree, _):
                 guard let function = parseMethod(source, definedIn: definedIn as? Type, nextStructure: next) else {
                     return nil
                 }
 
                 functions.append(function)
                 return function
-            case .functionSubscript:
+            case (.functionSubscript, _):
                 return parseSubscript(source, definedIn: definedIn as? Type, nextStructure: next)
-            case .varParameter:
+            case (.varParameter, _):
                 return parseParameter(source)
-            case .typealias:
+            case (.typealias, _):
                 switch parseTypealias(source, containingType: definedIn as? Type) {
                 case nil:
                     return nil
@@ -179,10 +180,11 @@ public final class FileParser {
                         return protocolComposition
                     }
                 }
-            case .associatedtype:
+            case (.associatedtype, _):
                 return parseAssociatedType(source, definedIn: definedIn as? Type)
             default:
-                Log.verbose("\(logPrefix) Unsupported entry \"\(access) \(kind) \(name)\"")
+                let nameSuffix = name.map { " \($0)" } ?? ""
+                Log.verbose("\(logPrefix) Unsupported entry \"\(access) \(kind)\(nameSuffix)\"")
                 return nil
             }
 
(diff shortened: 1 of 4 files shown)


SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -125,45 +125,46 @@ public final class FileParser {
         var typealiases = [Typealias]()
         try walkDeclarations(source: source) { kind, name, access, inheritedTypes, source, definedIn, next in
             let type: Type
-            switch kind {
-            case .protocol:
+            switch (kind, name) {
+            case (.protocol, .some(let name)):
```
IlyaPuchkaTW, line 129: I suppose with latest versions of Swift `.some` or `?` is not needed to unwrap optional in switch, so could be just `let (.protocol, name)`?

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -310,19 +312,27 @@ public final class FileParser {
...
+        guard let kind = (dict[SwiftDocKey.kind.rawValue] as? String).flatMap({ SwiftDeclarationKind(rawValue: $0) }) else {
+            return nil
         }
 
         let accessibility = (dict["key.accessibility"] as? String).flatMap({ AccessLevel(rawValue: $0.replacingOccurrences(of: "source.lang.swift.accessibility.", with: "") ) }) ?? .none
+        
+        let name: String? = (dict[SwiftDocKey.name.rawValue] as? String).map {
+            var name: String = $0
+            
+            if case .enumelement = kind, let colon = name.firstIndex(of: "(") {
```
IlyaPuchkaTW, line 325: why does it search for first `(` while called `colon`?

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -470,8 +480,12 @@ extension FileParser {
     }
 
     internal func parseVariable(_ source: [String: SourceKitRepresentable], definedIn: Type?, isStatic: Bool = false) -> Variable? {
-        guard let (name, _, accessibility) = parseTypeRequirements(source) else { return nil }
+        guard let (nameOrNil, _, accessibility) = parseTypeRequirements(source) else { return nil }
 
+        guard let name = nameOrNil else {
+            return nil
+        }
```
IlyaPuchkaTW, line 487: why to have two separate guards if both result in the same result? shouldn't the first `guard let (name, _ accessibility) = ...` be enough? same in other methods.

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -738,7 +756,8 @@ extension FileParser {
     }
 
     private func parseTypealias(_ source: [String: SourceKitRepresentable], containingType: Type?) -> TypealiasParseOutcome? {
-        guard let (name, _, _) = parseTypeRequirements(source),
+        guard let (nameOrNil, _, _) = parseTypeRequirements(source),
+            let name = nameOrNil,
```
IlyaPuchkaTW, line 760: same here, `guard let (.some(name), _, _)` ?

IlyaPuchkaTW:
Out of curiosity, what is your use case for such methods?

artyom-razinov:
> Out of curiosity, what is your use case for such methods?

You mean, methods without names? A protocol can be defined without specifying the name:

```
protocol MyProtocol {
     func doSomething(_: Int)
}
class MyClass: MyProtocol {
     func doSomething(_ x: Int) {
          print("did something with an arg: \(x)")
     }
}
```

We had 1 such occurrence in our 1+ million lines Swift project (but only few code is used as source for code generation though, so possibly it's more than one occurrence in the whole project). I could fix the client code (our project), but I don't see changing client code as an ideal solution: if the code is valid and not an extremely rare case, why not to support it, right?

SourceryFramework/Sources/Parsing/FileParser.swift
artyom-razinov: In this case `name` is optional:
```
let (.protocol, name)
```

Valid options are:

```
let (.protocol, name?)
(.protocol, let name?)
(.protocol, .some(let name))
(.protocol, let .some(name))
let (.protocol, .some(name))
```

What should I use? I can use first one if it suits codestyle (`let (.protocol, name?)`), it's the most contemporary syntax, I mean, it was added most recently.

artyom-razinov: Good to know is whether it's working with old Xcode version. Every version I described works in Xcode 11.

SourceryFramework/Sources/Parsing/FileParser.swift
artyom-razinov: I don't know. The person who touched this code was @Liquidsoul (in 2019).

I renamed it to `openingBrace`? Does it suit?

SourceryFramework/Sources/Parsing/FileParser.swift
artyom-razinov: It's a personal preference that arised after discovery that joined guards severely impact performance of compilation. The difference is noticeable. I remember doing researches several times, each of them resulted in significantly improved performance of compilation, like 20-30%.

It's like preferences of avoiding structs, making everything final, avoiding concatenation of strings (a + b + c works much slower than "\(a)\(b)\(c)"), avoiding auto-typing here and there, splitting long functions into parts (which is a good thing anyway), etc. I'm not sure about marking all protocols as `class` for example (`protocol Something: class`), it's a standard we use, but I didn't research performance impact on it (it's possible that it affects performance, but mainly it's use to allow usage of weak referencing).

That thing about `guard` is very real, it can easily make compilation time of a function more than a second, which doesn't sound all that bad, but for big enough project it can make compilation time incredibly long.

I would join guard in this pull request though, because it will be more fitting to the codestyle of the project.

SourceryFramework/Sources/Parsing/FileParser.swift
artyom-razinov: Done

artyom-razinov:
All things are done. Also one case was added to FileParserMethodsSpec.

Before that I forgot to run tests and it turned out that I broke backward compatibility with existing templates. Now the code has only small change for small case (`.varParameter`).

I added a comment here (see `Note:`):

```
/// Describes method parameter
@objcMembers public final class MethodParameter: NSObject, SourceryModel, Typed, Annotated {
    /// Parameter external name
    public var argumentLabel: String?

    /// Parameter internal name
    // Note: although method parameter can have no name, this property is not optional,
    // this is so to maintain compatibility with existing templates.
    public let name: String
```

Is it useful or should I delete it?

SourceryFramework/Sources/Parsing/FileParser.swift
```diff
@@ -641,7 +671,7 @@ extension FileParser {
         let `inout` = type.hasPrefix("inout ")
         let typeName = TypeName(type, attributes: parseTypeAttributes(type))
         let defaultValue = extractDefaultValue(type: type, from: source)
-        let parameter = MethodParameter(argumentLabel: argumentLabel, name: name, typeName: typeName, defaultValue: defaultValue, annotations: annotations.from(source), isInout: `inout`)
+        let parameter = MethodParameter(argumentLabel: argumentLabel, name: name ?? "", typeName: typeName, defaultValue: defaultValue, annotations: annotations.from(source), isInout: `inout`)
```
artyom-razinov, line 674: Note that I used `""` instead of `nil` to maintain backward compatibility of existing templates (swift templates won't compile otherwise). Was it a good idea?

ilyapuchka (approved):
👍

Issue #856, opened 2020-09-10
Infinite recursion introduced in 1.0.0


jpsim:
The change in #845 introduced a crash in Sourcery for me.

To reproduce:

```
$ swift --version
Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53)
Target: x86_64-apple-darwin19.6.0
$ git clone https://github.com/krzysztofzablocki/Sourcery.git
$ cd Sourcery
$ git checkout 6f2c00fb4af7bc3d5022dff8336da9c40c40283c
$ swift build -c release
$ cd ..
$ git clone https://github.com/realm/SwiftLint.git
$ cd SwiftLint
$ git checkout 4083f1688556fb55c3a4972f69bb95ea0405ac96
$ ../Sourcery/.build/release/sourcery --sources Source/SwiftLintFramework/Rules --templates .sourcery/MasterRuleList.stencil --output .sourcery
No config file provided or it does not exist. Using command line arguments.
Scanning sources...
	loadOrParse: 1.1053340435028076
[1]    15167 bus error  ../Sourcery/.build/release/sourcery --sources Source/SwiftLintFramework/Rules
```

This crash doesn't happen with the previous commit on master: a441c57219fd933cd63ab49b162ad98cddfbda68

I think the crash is due to a stack overflow caused by infinite recursion.

Here's the backtrace I get when running Sourcery with Xcode:

<details>

<summary>Backtrace</summary>

```
* thread #12, queue = 'com.apple.root.user-interactive-qos', stop reason = EXC_BAD_ACCESS (code=2, address=0x70000a5a8ff8)
    frame #0: 0x00007fff713e1d36 libswiftCore.dylib`std::__1::pair<(anonymous namespace)::TupleCacheEntry*, swift::MetadataResponse> swift::LockingConcurrentMap<(anonymous namespace)::TupleCacheEntry, (anonymous namespace)::TupleCacheStorage>::getOrInsert<(anonymous namespace)::TupleCacheEntry::Key, swift::MetadataRequest&, swift::TargetValueWitnessTable<swift::InProcess> const*&>((anonymous namespace)::TupleCacheEntry::Key, swift::MetadataRequest&, swift::TargetValueWitnessTable<swift::InProcess> const*&) + 38
    frame #1: 0x00007fff713e18bc libswiftCore.dylib`swift_getTupleTypeMetadata + 140
[...]

PR #874, opened 2020-11-25
Updates Commander

hungrxyz:
This one should help resolve SPM dependencies when other dependencies, such as SwiftGen, require 0.9.0 or higher version of Commander.


hungrxyz:
@krzysztofzablocki Is there a milestone for 1.0.1? Would be good to get these dependency updates out ASAP.

krzysztofzablocki:
@hungrxyz once we fix #856, I'm going to try to figure it out this week since it's holidays in USA

PR #884, opened 2020-12-19
add .swiftpm/ to gitignore

Cyberbeni:
When opening Package.swift with Xcode, it will create a .swiftpm directory, this should be in the gitignore


krzysztofzablocki:
thanks

PR #897, opened 2021-02-05
Change how applying type extensions work

krzysztofzablocki:
I think that if Sourcery finds type extensions it should only add methods/variables/subscripts if they are considered unique, and when resolving uniqueness the order should be defined as:
- base implementation
- inheritance
- protocol conformance
- type extensions

Scenario 1.

Given a code like
```swift
protocol UrlOpening {
  func open(_ url: URL)
  func open(
    _ url: URL,
    options: [UIApplication.OpenExternalURLOptionsKey: Any],
    completionHandler completion: ((Bool) -> Void)?
  )
}

extension UrlOpening {
    func open(_ url: URL) {
        open(url, options: [:], completionHandler: nil)
    }
}
```

You'd expect `UrlOpening.methods` to contain 2 methods and not 3 like we have currently because the default implementation shouldn't matter from AST perspective right?

Scenario 2.

```swift
protocol RemoteUrlOpening {
  func open(_ url: URL)
}

protocol UrlOpening: RemoteUrlOpening {
  func open(
    _ url: URL,
    options: [UIApplication.OpenExternalURLOptionsKey: Any],
    completionHandler completion: ((Bool) -> Void)?
  )
}

extension UrlOpening {
  func open(_ url: URL) {
    open(url, options: [:], completionHandler: nil)
  }
}
```

When looking at `UrlOpening.allMethods` I should see the `open(_:options:completionHandler:)` and method from `RemoteUrlOpening`, where both are not `definedInExtension` and we should simply ignore the default implementation here. (currently we return the base method + extension one so we miss the inheritance).

Do you agree about those expectations?
/cc @Przemyslaw-Wosko @g-amichnia


IlyaPuchkaTW:
Seems sensible to me. Should we also specially consider protocol extensions? We probably don't need to distinguish it on the level of dynamic dispatch (wether method was defined in the protocol and then has a default implementation or only has a default implementation), but type should have this method whenever it is in the protocol it conforms to or in its extension. Although maybe we already do that when aggregating protocol methods, but maybe worth having as a test case.

krzysztofzablocki:
@ilyapuchka not sure if I follow what you meant here but I've [added 2 protocol inheritance + default implementation tests](https://github.com/krzysztofzablocki/Sourcery/pull/897/files#diff-f2c1ae0772c2e0160d781b2a17734df7a68ae6494c5f1fc3a11f6550f468de08R721) do they cover It?
if not could you write some sample code / expectation so I can spec it properly?

IlyaPuchkaTW:
yes, that's one of the cases I mean, I'd add another case like that to make sure we consider protocol extensions methods even if they were not defined in a protocol. Unless we already have this covered with another test, then it's ok! =)

```
protocol UrlOpening {
  func openURL(_ url: URL, parameters: ...)
}

extension UrlOpening {
  func openURL(_ url: URL) { ... }
}
```

krzysztofzablocki:
oh yeah that's covered by checking if `UrlOpening.allMethods` contains `anotherFunction(key:)` in [this test](https://github.com/krzysztofzablocki/Sourcery/pull/897/files#diff-f2c1ae0772c2e0160d781b2a17734df7a68ae6494c5f1fc3a11f6550f468de08R731)

CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 # Sourcery CHANGELOG
 
 ---
+## Master
+
+### New Features
+- [PR](https://github.com/krzysztofzablocki/Sourcery/pull/897) Methods, Variables and Subscripts will be uniqued in both `[methods]`, `all[Methods]`, new accessor is introduced that doesn't get rid of duplicates `raw[Methods]`. The deduping process works by priority order (highest to lowest):
```
ZevEisenberg, line 7:
```suggestion
- [PR](https://github.com/krzysztofzablocki/Sourcery/pull/897) Methods, Variables and Subscripts will be uniqued in both `[methods]` and `all[Methods]`. A new accessor is introduced that doesn't get rid of duplicates: `raw[Methods]`. The deduping process works by priority order (highest to lowest):
```

ZevEisenberg: Also, why are `all[Methods]` and `raw[Methods]` spelled with brackets here? I'm not familiar enough with Sourcery templating to get the reference.

SourceryRuntime/Sources/Description.generated.swift
```diff
@@ -254,9 +254,9 @@ extension Type {
         string += "name = \(String(describing: self.name)), "
         string += "isGeneric = \(String(describing: self.isGeneric)), "
         string += "localName = \(String(describing: self.localName)), "
-        string += "variables = \(String(describing: self.variables)), "
-        string += "methods = \(String(describing: self.methods)), "
-        string += "subscripts = \(String(describing: self.subscripts)), "
+        string += "rawVariables = \(String(describing: self.rawVariables)), "
+        string += "rawMethods = \(String(describing: self.rawMethods)), "
+        string += "rawSubscripts = \(String(describing: self.rawSubscripts)), "
```
ZevEisenberg, line 259: Out of scope, but I bet a multiline string would help a little with compile time and a lot with formatting here

SourceryRuntime/Sources/Type.swift
```diff
@@ -53,69 +53,141 @@ import Foundation
...
-                let all = all.array as! [T]
-                let extracted = extraction(target).filter({ filter(all, $0) })
-                return extracted
+        let allObjects = extraction(self)
+
+        /// The order of importance for properties is:
+        /// Base class
+        /// Inheritance
+        /// Protocol conformance
+        /// Extension
```
ZevEisenberg, line 150:
```suggestion
        // The order of importance for properties is:
        // Base class
        // Inheritance
        // Protocol conformance
        // Extension
```

ZevEisenberg (approved):
I'm not familiar with the use cases where this would make a difference, but the change itself seems reasonable. Would it constitute a breaking change? i.e. would existing users find that unexpected things happen when they update?

SourceryRuntime/Sources/Type.swift
ZevEisenberg: Also maybe specify whether this is from highest to lowest or lowest to highest

krzysztofzablocki: yeah I did in changelog

CHANGELOG.md
krzysztofzablocki: it just meant `allMethods` `allVariables` `allSubscripts`, better way to express that?

krzysztofzablocki:
@ZevEisenberg in most cases it shouldn't matter since there were very few tests I've had to touch and those I did were just defining wrong Swift code, but I'm going to release it as new version anyway just to be safe

CHANGELOG.md
ZevEisenberg: Since there are only three, I'd probably just list them out, or if you want to be fancy, maybe `all[Methods/Variables/Subscripts]`, but honestly I think just listing them is better.

Issue #900, opened 2021-02-10
Variables with Self type do not get resolved


erichoracek:
In [Swift 5.1](https://swift.org/blog/swift-5-1-released/), [SE-0068](https://github.com/apple/swift-evolution/blob/master/proposals/0068-universal-self.md) expanded Swift’s use of `Self` so that it refers to the containing type when used inside classes, structs, and enums. However, Sourcery doesn't seem to support this. For example:

If you have the following Swift file `test.swift`:
```swift
struct A {
  static var specificType: A { .init() }
  static var selfType: Self { .init() }
}
```

And the following template `test.stencil`:
```stencil
{% for type in types.types %}
{% for variable in type.staticVariables %}
{{ variable.name }}
- {{ variable.type.name }}
- {{ variable.typeName.name }}
{% endfor %}
{% endfor %}
```

And run `sourcery --sources test.swift --templates test.stencil --output output.txt`:

You'll get the following `output.txt`:
```
specificType
- A
- A
selfType
- 
- Self
```

Note that `variable.type.name` is `nil` for `selfType`.

My expectation would be the following output instead:
```
specificType
- A
- A
selfType
- A
- Self
```

Tested on Sourcery version 1.1.1

krzysztofzablocki:
Thanks for providing nice report @erichoracek, I'll try to add this feature to next release

Issue #899, opened 2021-02-08
Automatically resolve types across modules


krishnaagarwal1994:
I am parsing a method in a class A present in TargetA, 
Class A has a method that has a return type of Class B, which is also present in TargetA.
But sourcery is still returning false for 
```
{{ method.returnType }} -> nil
{{ method.returnTypeName }} -> ClassB

```
The configuration file that I am using

```
project:
  - file: #path to project file
    target:
      name: TargetA
templates:
    - SourceryTemplates
output: src/Generated
```

krishnaagarwal1994:
@ilyapuchka Can you please help me with this?
Thank You.

krzysztofzablocki:
@krishnaagarwal1994 can you create a sample project reproducing the issue? it would help

krishnaagarwal1994:
[sourcery0.zip](https://github.com/krzysztofzablocki/Sourcery/files/5954296/sourcery0.zip)
Hi @krzysztofzablocki , thanks for your response.
I have created a sample project, in which I am able to reproduce the problem.
Please have a look into it, and do let me know in case you have any questions regarding this Sample Project.
Thanks.

krzysztofzablocki:
thanks, seems that something is off with cross module resolution, if you add the target name explicitly to `getStudentB() -> CustomTarget.Student` then it gets all information, I'll look into it.

krishnaagarwal1994:
@krzysztofzablocki Yeah, I already tried that, if mentioning the target name explicitly then it's identifying correctly.
Also, I would like to share one more finding with you.

If I am mentioning project path, target names then the Sourcery is retuning the returnType as nil
But if you simply provide the target files path using the source attribute in the Sourcery configuration file then it's able to identify the returnType.

krishnaagarwal1994:
@krzysztofzablocki  
Changing the sourcery.yml file contents to below is giving the right results.
```
sources:
  - CustomTarget
  - CustomTargetB
templates:
    - SourceryTemplates
output: Generated

```

krzysztofzablocki:
yeah sources don't run into those kind of issues, that's original implementation and more popular, xcodeproj was contributed later and there are some kinks there to figure out

krishnaagarwal1994:
Actually, I am working on a quite complex project, where I need to support functionality on a module/target level.
At some places the physical path and logical path of files are different. and some files belong to more than one target. 
[...]

PR #901, opened 2021-02-11
Type inference and resolution improvements

krzysztofzablocki:
- `Self` reference is resolved to correct type. [Enchancement Request](https://github.com/krzysztofzablocki/Sourcery/issues/900)
- Sourcery will now attempt to resolve local type names across modules when it can be done without ambiguity. Previously we only supported fully qualified names. [Enchancement Request](https://github.com/krzysztofzablocki/Sourcery/issues/899)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Sourcery CHANGELOG
 
 ---
+## Master
+
+### New Features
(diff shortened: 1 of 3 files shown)


erichoracek (approved):
Thank you! I don't have enough context on resolving local type names across modules boundaries but the Self reference changes look great to me.

PR #902, opened 2021-02-12
Added hasPrefix and hasSuffix filters on String

sroebert:
I have used Swiftgen quite a lot and I'm quite used to having `hasPrefix` and `hasSuffix` filters when using stencil templates. I thought it would be a good addition for Sourcery as well.


krzysztofzablocki:
@sroebert we already have them, both SwiftGen and Sourcery use same stencil extension https://github.com/SwiftGen/StencilSwiftKit :)

sroebert:
Really? It was not working for me, until I added this code.

krzysztofzablocki:
definitely, we use the linked extension, which version of Sourcery were you using? might have been added recently but we should be at newest

sroebert:
Ok, I guess I was calling it wrong in the beginning, it is working now, thanks for the quick response.

krzysztofzablocki:
@sroebert thank you, you went into actually contributing and not just reporting an issue and that's awesome :)

PR #907, opened 2021-03-02
run sourcery with multiple configurations

### LINKED ISSUE #727: multiple configurations in .sourcery.yml
I want to apply several templates just to one specific folder each.

Is there an option to have multiple configurations in the .sourcery.yml file?
I think something like this would be nice:

```
configurations:
  - sources:  Library/Model/
    templates: Sourcery/a.stencil
    output: Library/Generated
  - sources: Main/Model/
    templates:  Sourcery/b.stencil
    output: Main/Generated
```

ilyapuchka:
Resolves #727 

When yaml contains "configurations" root key then we try to parse it as array of configurations and run sourcery for each of them. In watch mode we will observe all paths from all configurations.

- [x] Haven't tested that on a real config yet but based on code we run in main don't see why it wouldn't work.
- [x] After successful test will update documentation for configuration as well.


ilyapuchka:
@krzysztofzablocki I've tested this on my actual project and it woks well, also updated the docs, so it's ready for review.

krzysztofzablocki:
👌

PR #921, opened 2021-03-15
do not set raw value types on enums with no cases or cases with associated values

IlyaPuchkaTW:
We can't make raw value types detection bullet proof (for implicit values inferred by the compiler) but we can at least cut out cases that can't have associated values per language specification. Also added the note in the docs explaining the behaviour of inferring raw value type from inherited types.


SourceryRuntime/Sources/Composer.swift
```diff
@@ -576,6 +576,12 @@ public enum Composer {
             enumeration.rawTypeName = rawValueVariable.actualTypeName
             enumeration.rawType = rawValueVariable.type
         } else if let rawTypeName = enumeration.inheritedTypes.first {
+            // enums with no cases or enums with cases that contain associated values can't have raw type
```
IlyaPuchkaTW, line 579:
```suggestion
            // enums with no cases or enums with cases that contain associated values can't have raw value
```

krzysztofzablocki (approved):

PR #930, opened 2021-03-26
keep length when removing generated code before parsing the syntax tree

Cyberbeni:
Sourcery removes the code inside inline blocks for syntax tree parsing, which changes the bodyTypeRange calculated by SyntaxTreeCollector which is used when calculating ranges for new sourcery:inline:auto blocks.

If we replace the code that we don't want to parse with the same number of spaces, the rangeInFile will match with the range calculated by SyntaxTreeCollector.

Stencil:

```
{% for type in types.all %}
// sourcery:inline:auto:{{ type.name }}.init
{{ type.accessLevel }} init({% for variable in type.storedVariables %}{{variable.name}}: {{variable.typeName}}{% if variable.defaultValue %} = {{variable.defaultValue}}{% endif %}{% ifnot forloop.last %}, {% endif %}{% endfor %}) {
{% for variable in type.storedVariables %}
	self.{{variable.name}} = {{variable.name}}
{% endfor %}
}
// sourcery:end
{% endfor %}
```

Source file:

```swift
class A {
	let o: String

	// sourcery:inline:auto:A.init
	internal init(o: String) {
		self.o = o
	}
	// sourcery:end
}

class B {
	let o: String
}
```

Original output:

```swift
class A {
	let o: String

	// sourcery:inline:auto:A.init
	internal init(o: String) {
		self.o = o
	}

// sourcery:inline:auto:B.init
internal init(o: String) {
	self.o = o
}
// sourcery:end
	// sourcery:end
}

class B {
	let o: String
}
```

Output with the changes in PR:

```swift
class A {
	let o: String

	// sourcery:inline:auto:A.init
	internal init(o: String) {
		self.o = o
	}
	// sourcery:end
}

class B {
	let o: String

// sourcery:inline:auto:B.init
internal init(o: String) {
	self.o = o
}
// sourcery:end
}
```


krzysztofzablocki:
@Cyberbeni I'm not sure if Spaces are in-fact going to work in all cases, take a look at `StringView` as that's what we use, you might want to add some composed unicode characters to the tests (like emoji) and I think we'd need to match utf8 length instead of number of chars

Cyberbeni:
Emojis seem to just break inline:auto without having any Sourcery generated code to begin with.

Got an out of bounds crash when trying to run the same stencil on this file: 
[a.swift.zip](https://github.com/krzysztofzablocki/Sourcery/files/6238524/a.swift.zip)

NSString is UTF16, I don't see why we are specifying UTF8 in a lot of places. I'm gonna try turning all the utf8 range magic into utf16 on the weekend to see if that helps. (Unless there is a good reason why we need utf8 in those places @krzysztofzablocki ?)

ps: there are also composite emojis, like 👩‍🚀 (which is 3 UTF16 characters as opposed to just 1 in the attached file), IDK if that breaks anything more than regular emojis.

Cyberbeni:
It turns out, it was just 1 spot where the wrong StringView was used for range calculations, so I just fixed that instead of finding out the hard way why UTF8 ranges are needed.

krzysztofzablocki:
that's cool @Cyberbeni, thanks for contributing! 

I'll get this and other hanging pr's in this week :)

PR #940, opened 2021-04-09
Extending AutoMockable template by adding handling of "autoMockableImports" and "autoMockableTestableImports" args.

adurbalo:
This change provides the ability to setup mocks generation like:

```
    - sources: # AutoMockable
      - "./Source/"
      templates:
        - "./SourceryTemplates/Tests"
      output: "./Example/Tests/Mocks/SourceryGenerated"
      args:
        autoMockableImports:
          - CoreLocation
        autoMockableTestableImports:
          - SomeTestableFramework
```

As a result of generation, we got:

```
import Foundation
#if os(iOS) || os(tvOS) || os(watchOS)
import UIKit
#elseif os(OSX)
import AppKit
#endif

import CoreLocation

@testable import SomeTestableFramework
...
```


krzysztofzablocki:
hey @adurbalo, 
thanks for contributing 🙇 

I think we should scope this arguments to mockable template? otherwise it might collide with other templates arguments

adurbalo:
@krzysztofzablocki hmm, yep it can be not obvious for someone. In my case, I split testing and main project to separate configurations (with different paths and templates and so). But for more generic solution I would ask suggestions from your side.

krzysztofzablocki:
just prepend the argument name with automockable prefix and we should be ok 👌

adurbalo:
@krzysztofzablocki thanks for the suggestion!

krzysztofzablocki:
@adurbalo sweet, thanks!

Issue #936, opened 2021-04-03
Annotations field is empty for protocol composition starting from Sourcery 1.3.0


Roman-Andreyenko:
I use Swift version 5.3.2.
I have following source code:

// sourcery: AppContext
typealias AppContext = NetworkReachabilityServiceContainer
    & RootStylePropertyServiceContainer
    & FirebaseAppServiceContainer
    & DeeplinkHandlerServiceContainer

NetworkReachabilityServiceContainer, RootStylePropertyServiceContainer, FirebaseAppServiceContainer and DeeplinkHandlerServiceContainer are protocols.

I run sourcery with following template swifttemplate:

// sourcery:file:Factories/Test
    <%= types.protocolCompositions %>
// sourcery:end 

I received following output:

// Generated using Sourcery 1.3.0 — https://github.com/krzysztofzablocki/Sourcery
// DO NOT EDIT
    [ProtocolComposition: module = nil, imports = [], typealiases = [:], isExtension = false, kind = protocolComposition, accessLevel = internal, name = AppContext, isGeneric = false, localName = AppContext, rawVariables = [], rawMethods = [], rawSubscripts = [], initializers = [], annotations = [:], staticVariables = [], staticMethods = [], classMethods = [], instanceVariables = [], instanceMethods = [], computedVariables = [], storedVariables = [], inheritedTypes = ["NetworkReachabilityServiceContainer", "RootStylePropertyServiceContainer", "FirebaseAppServiceContainer", "DeeplinkHandlerServiceContainer"], containedTypes = [], parentName = nil, parentTypes = AnyIterator<Type>(_box: Swift._IteratorBox<Swift._ClosureBasedIterator<SourceryRuntime.Type>>), attributes = [:], kind = protocolComposition, composedTypeNames = [NetworkReachabilityServiceContainer, RootStylePropertyServiceContainer, FirebaseAppServiceContainer, DeeplinkHandlerServiceContainer]]

But I expect that "annotations = ["AppContext": 1]". Sourcery 1.2.1 generates correct output

bdunay3:
I'm seeing this exact same issue in 1.3.4. I've modified the AutoMockable template to take protocol compositions into account since we are doing dependency injection using protocol compositions. If I remove the `where type|annotated:"AutoMockable"` check, it works (but I get all protocol compositions which is not what I want).

bdunay3:
[...]

PR #942, opened 2021-04-12
Fixed: Annotations field is empty for protocol composition starting from Sourcery 1.3.0

bdunay3:
I've fixed [issue 936](https://github.com/krzysztofzablocki/Sourcery/issues/936) where the annotations dictionary was empty for any ProtocolComposition type. When support was added for that type the `AnnotationsParser` wasn't invoked and the results not passed into the constructor for `ProtocolComposition`.


krzysztofzablocki (approved):
thanks for contributing 🙇

PR #944, opened 2021-04-15
Fix broken output for files with inline annotations from multiple templates

HeMet:
**Problem**: to process annotations Sourcery uses `ParsingResult` computed before any template generated and any file modified. Problem is sources can be altered but respective ranges in `ParsingResult` never updated. That means when next template will be processed inline annotations could be generated at wrong locations.

**Solution**: update `ParsingResult` using information about changes made to files after each template processed.

**Note**: it's possible to arrange templates and annotations in such way what the issue won't occur. The first template should change the last annotation in a file, the second template should change `N-1` annotation in file and so on. This order ensured for changes made by a single template but this is not enough.

**Note #2**: one test modified since that test checked that `inline:auto` annotations inserted at beginning of the type's definition not at the end. Two test added.


krzysztofzablocki:
@HeMet wasn't this fixed by https://github.com/krzysztofzablocki/Sourcery/pull/930 ?

krzysztofzablocki:
@HeMet did you consider the changes from #930 in here?

HeMet:
> @HeMet did you consider the changes from #930 in here?

That particular change is important https://github.com/krzysztofzablocki/Sourcery/pull/930/files#diff-a8fcd586390fbf1495762ec98f9633cf0a5363ab048f73b210bc263238c9db3fR556 

I made that change too, but noticed what it is already in master (originally this patch was based on older version of master).

That change affects string generated by template, not source files so it should be fine 
https://github.com/krzysztofzablocki/Sourcery/pull/930/files#diff-7f4054a27fafedc450c7c113e7721ea126d50a26acef361f51b328649f1a3247R29

krzysztofzablocki:
Ok cool, the code looks really clean and well written, I'll do proper review on the weekend but will probably be fine as it is, great job 👏

would you mind adding changelog info?

HeMet:
@krzysztofzablocki Did you have any chance to review this PR?

krzysztofzablocki:
yeah I think it's good, I'll be adding sourcerytemplate support today and will merge this in along that and make a release

PR #945, opened 2021-04-27
Update Class.swift

robertjpayne:
Use modifiers to determine if class is final


robertjpayne:
@krzysztofzablocki sorry I don't have enough time to setup test cases and sorts here but this fixes it after inspecting some of the sources and what not.

It looks like the `final` keyword is not part of the attributes list anymore but rather part of the modifier list.

PR #947, opened 2021-04-27
Support Swift5.4

funzin:
- Updated swift-syntax versin from 0.50300.0 to 0.50400.0
- By containing this, can use Xcode12.5

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   test:
     macos:
-      xcode: "12.4.0"
+      xcode: "12.5.0"
     environment:
       - XCODE_WORKSPACE: Sourcery.xcworkspace
       - XCODE_PROJECT: Sourcery.xcodeproj
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   test:
     macos:
-      xcode: "12.5.0"
+      xcode: "12.4.0"
     environment:
       - XCODE_WORKSPACE: Sourcery.xcworkspace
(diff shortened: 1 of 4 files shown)


funzin:
Circle CI has not  supported Xcode12.5 yet🤦‍♂️
スクリーンショット 2021-04-28 8 53 11

krzysztofzablocki (changes requested):
Sourcery works on new Xcode if you use binary distribution.

This is useful for updating to newer version but you need to update bundled dylib, that's what we rely on because we rewrite rpath, we are not relying on user Xcode version

funzin:
>Sourcery works on new Xcode if you use binary distribution.

I installed Sourcery via SwiftPackageManager or Mint, not use binary distribution.

krzysztofzablocki:
@funzin I'm happy with the PR but you just need to copy the dylib from new Xcode and put it in the repo as we bundle it, once we move to SPM branch it will be using right dylib bundling regardless of Xcode version too

funzin:
I understood.

I should only replace [lib_InternalSwiftSyntaxParser.dylib](https://github.com/krzysztofzablocki/Sourcery/blob/master/lib_InternalSwiftSyntaxParser.dylib) in the repo to new dylib which Xcode12.5 has, right?

krzysztofzablocki:
no no, this PR is good, you just need to replace that file and update `changelog.md` then I can merge :)

funzin:
- Replaced `lib_InternalSwiftSyntaxParser` f1def25
- Updated ChangeLog dc1de30
  - Current version is `1.4.1` so I added `1.4.2` section

funzin:
Failed test...

PR #954, opened 2021-05-04
Remove symlinks

HeMet:
Symlinks removed since these can confuse Xcode (files showed twice in search, content in editor goes out of sync with file on storage). 

`SourceryExecutable` target introduced which contains `main.swift` only. This way `Sourcery` can be imported by test target. However, respective target has name `SourceryLib` due to another Xcode issue.

Minimal changes were made to project structure to avoid changing scripts.


krzysztofzablocki:
🙇

PR #801, opened 2019-12-30
Process inline code generated by another template if annotation's suffix is one of the templates in force-parse argument

ranvirp:
There is requirement that inline code generated from a different template can be processed. force-parse arguments provide this ability for generated code in a file. In this pull request, annotationranges is given forceparse argument and if the inline annotation ends with one of the templates then that inline code is not removed.

PR #957, opened 2021-05-11
Process inline code generated by another template if annotation's suffix is one of the templates in force-parse argument

bitomule:
This PR is based on @ranvirp work here https://github.com/krzysztofzablocki/Sourcery/pull/801

Just re-created based on his test with latest master changes


krzysztofzablocki:
🙇

Issue #906, opened 2021-02-24
configuration: Linking inside the root group


nnnax:
I've created a very basic Xcode project named "simple" 
I want to link my generated files directly inside the default group named "simple" so my configuration file is set as follow : 

<img width="853" alt="Capture d’écran 2021-02-24 à 23 20 04" src="https://user-images.githubusercontent.com/2895717/109073971-621d0200-76f7-11eb-8523-850b5ab9216c.png">

After each Sourcery generation, a new group named "simple" is created
I expect to have my files added to the default group :

<img width="323" alt="Capture d’écran 2021-02-24 à 23 21 52" src="https://user-images.githubusercontent.com/2895717/109074464-14ed6000-76f8-11eb-9ace-38c2f60fd2a8.png">

After investigation, the default group name is empty (should be set to "simple" in my opinion) but the default group path is set to "simple" 

So I've applied a temporary fix in Sourcery.swift :

<img width="891" alt="Capture d’écran 2021-02-24 à 23 33 31" src="https://user-images.githubusercontent.com/2895717/109074969-c8565480-76f8-11eb-85ad-1818032cb8f5.png">

It does the job for me but maybe it's not really a definitive solution

Kind Regards

krzysztofzablocki:
@nnnax hey, what happens if you don't specify the group name in the config? I'm looking at docs and they say that it should add the files to your root group

nnnax:
Hi @krzysztofzablocki,
If I don't specify the group, my file is added under the project node (so near my default group) exactly as we see in the second screenshot. I think it's the expected behavior because the root group is the project node.

And the code itself is correct, I mean by default the rootGroup is used :
        var fileGroup: PBXGroup = rootGroup

But, I think the real problem is that my rootGroup children are not correctly defined (maybe an issue in xcodeproj)
If I loop through my rootGroup children, the child representing my "default group" has a "name" property empty

<img width="307" alt="Capture d’écran 2021-02-25 à 09 18 08" src="https://user-images.githubusercontent.com/2895717/109124177-eb115900-774a-11eb-9cb1-b009272b2932.png">

krzysztofzablocki:
Ok, thanks for detailed write up 🙇‍♂️

I'll fix it but likely in next release after 1.3 since 1.3 is already going to be massive

nnnax:
@krzysztofzablocki  
Just to let you know, after an in-depth analysis, in fact the issue is exactly the same as the issue #904 

[...]

PR #958, opened 2021-05-12
Fix group creation

avdyushin:
Reuse groups (also created), will fix also: #906 
Related bug in dependency is: https://github.com/tuist/XcodeProj/issues/613


krzysztofzablocki:
thanks for contributing!

PR #962, opened 2021-05-19
Adds hasSetter property to Variable type

bdunay3:
I've added a new property to the `Variable` type to tell when a variable is a property that has a setter implemented. The need arose while I was implementing my custom AutoMockable template where the setter for a property in a mock implementation is only generated if `set` is specified in the protocol declaration.

Note: In the AutoMockable template included with Sourcery will generate a getter and setter implementation for any non-optional property in a protocol regardless of how it is declared.

For example if you declare the following protocol:
```
// sourcery: AutoMockable
protocol Foo {
    var bar: String { get }
}
```

I want my mock class output to be this:
```
final class MockFoo: Foo {
    var bar: String {
        get { return underlyingValueBar }
    }
    var underlyingValueBar: String!
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Sourcery CHANGELOG
 
 ---
+## 1.4.3
+
+## Fixes
+
+## New Feature
+- Added new `hasSetter` property to `Variable`.
+
+---
 ## 1.4.2
 
 ## Fixes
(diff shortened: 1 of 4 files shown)


bdunay3:
I bought Sourcery Pro (great tool BTW, I absolutely love it so far!) and noticed that when running the above template code in that tool, it works as expected. But if I run it against the command line tool, I get the issue reported above.

PR #968, opened 2021-06-06
Update swift cli detection

### LINKED ISSUE #967: Detecting build running from terminal doesn't work (in GitHub Actions)
My action was running these commands:
```bash
/usr/local/bin/git clone --depth 1 --branch 1.4.2 https://github.com/krzysztofzablocki/Sourcery /Users/runner/install-swift-tool-794b233d-2f0c-5d6d-a5a9-5ee54587f249
/usr/bin/swift build --package-path /Users/runner/install-swift-tool-794b233d-2f0c-5d6d-a5a9-5ee54587f249 --configuration release --disable-sandbox
```

The build failed because of `<unknown>:0: error: unknown argument: '-rpath'`: https://github.com/Cyberbeni/install-swift-tool/runs/2754226819?check_suite_focus=true#step:3:224

It's not supposed to add that linker flag when running from command line: https://github.com/krzysztofzablocki/Sourcery/blob/1.4.2/Package.swift#L225

Shouldn't we check if `ProcessInfo.processInfo.arguments[0] == "/usr/bin/swift"` instead of the environmental variables?

Cyberbeni:
Fixes #967 

The old logic was relying on the ["_" environemnt valiable](https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_03_02.html), which was `/usr/local/opt/runner/provisioner/provisioner` inside GH Actions.
When using Xcode/xcodebuild, it updates the PATH, which works regardless of where it was launched from.


krzysztofzablocki:
thanks 🙇

PR #969, opened 2021-06-06
Add support for Property Wrappers

### LINKED ISSUE #803: Support Swift 5.1's Property Wrappers
I have a struct where some of the fields are using some custom property wrappers. I want to be able to extract those values. Example

Using the Lazy property from this spec https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md#delayed-initialization

```swift
protocol Test {}
struct Foo: Test {
    @Lazy var bar: Int = 1234
}
```

With the template

```stencil
{% for type in types.implementing.Test %}
extension {{type.name}} {
    func test() {
        {% for variable in type.variables %}
        print("{{variable.name}}", "{{variable.typeName}}")
        {% endfor %}
    }
}
{% endfor %}
```

I get

```swift
extension Foo {
    func test() {
        print("bar", "Int")
    }
}
```

But it doesn't have the value `_bar: Lazy<Int>`. Also, these property wrappers have support for projectedValue's too. In which case, `$bar: Lazy<Int>` would also be available, but the type would have to be known in that case.

daltonclaybrook:
Resolves #803 

This PR adds support for parsing Property Wrappers into the attributes list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Master
 ## Features
 - adds support for variadic parameters in functions
+- Adds support for parsing property wrappers
 
 ## Fixes
 - correct passing `force-parse` argument to specific file parsers and renames it to `forceParse` to align with other naming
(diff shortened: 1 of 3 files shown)


krzysztofzablocki:
🙇

Issue #965, opened 2021-05-26
[Feature Request] Same-line trailing comment annotation parsing


JoGoFo-:
It would be really cool if Sourcery could identify comment annotations at the _end_ of the line and associate it with the preceding statement.

For example:
```
let number: Int // sourcery: first = 1
```
and
```
enum Foo {
    case bar /* sourcery: first = "value" */
}
```
This would be helpful for cleaner formatting of code.

In the Writing Templates guide it says:
> Using /* and */ for annotation comment you can put it on the same line with your code. 

My initial interpretation of this was as described in above `Foo` example, it wasn't particularly clear that the annotation had to come _before_ the type.

I don't know how complex this would be to implement, I love this tool and would be more than happy to have a try myself if someone could guide me on how to get started. Tried cloning the repo already - the circleci config references a 'Sourcery.xcworkspace' which doesn't seem to exist anywhere, and the Rakefile _runs_ but fails a bunch of tests.

Thanks!

krzysztofzablocki:
hey you want to open package.swift in Xcode and use that for development now, thanks for reminding me to update circle config 😅

AnnotationParser is a separate object you could try tweaking to support that logic, I'd be fine with same line being annotation like that being processed with the associated statement 👍

art-divin:
Closing this one since it was implemented by #970

PR #970, opened 2021-06-12
Support for trailing line annotations

JoGoFo-:
Add support for annotations at the end of a line of code - see #965 

For example:

```
let number: Int // sourcery: first = 1
```
or
```
enum Foo {
    case bar /* sourcery: first = "value" */
}
```

If there are multiple declarations on the same line, they will all inherit the annotation:
```
let one: Int, two: Int // sourcery: default = 1
```

I'm open to any thoughts or feedback - thanks!


krzysztofzablocki (approved):

krzysztofzablocki:
Sweet, thanks for contributing!

 I'll fix the docs things separately

PR #975, opened 2021-07-21
Fix unintentionally losing attributes of closures inside extra optional closure

### LINKED ISSUE #974: Attributes of closures inside extra closures are lost when unwrapping type name.
This is a legit piece of code:
```swift
func test(_ closure: (@convention(c) () -> ())?) {
}
```

## Losing attributes

When parsing `(@convention(c) () -> ())?`, I expected one level of type name unwrapping to become `(@convention(c) () -> ())`. However, it currently turns into `(() -> ())`, completely losing the attributes.

I understand omitting the attribute is intentional, as shown by this test case:
```swift
it("removes attributes in unwrappedTypeName") {
  expect(typeName("@escaping (@escaping ()->())->()").unwrappedTypeName).to(equal("(@escaping () -> ()) -> ()"))
}
```

But here the attribute is "inside" the extra closure, and we are only unwrapping `?`, so it should still keep the attributes.

cheeseonhead:
This PR fixes #974.

For example, when parsing this type:
`(@convention(c) () -> ())?`

The old behavior of unwrapping it turns into `(() -> ())`.

The new behavior now retains the attributes as well as follow existing closure behavior `(@convention(c) () -> ())`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - corrects `isMutable` regression on protocol variables #964
 - Added multiple targets to link
 - Fix groups creation
+- Fix unintentionally losing attributes of closures inside extra optional closure #974
 
 ---
 ## 1.4.2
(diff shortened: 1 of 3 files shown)


cheeseonhead:
I added a few more tests.

I found the behavior of this test to be quite surprising. I hope it's expected:
```swift
expect(typeName("(((@escaping (@escaping ()->())->())))").unwrappedTypeName).to(equal("(@escaping () -> ()) -> ()"))
```

cheeseonhead:
I think this fix is wrong. I ran into more edge cases in the past that this fix doesn't work on. Closing.

PR #976, opened 2021-07-21
Thin lib_InternalSwiftSyntaxParser.dylib in `rake build`

Cyberbeni:
`lipo` invalidates the signature but I think that shouldn't matter if the file doesn't have the quarantine attribute. I don't have my Mac with me ATM, so make sure to check that before merging/releasing.

~~edit: also I don't know how to get the current architecture inside the Rakefile~~

edit 2: `lib_InternalSwiftSyntaxParser.dylib` probably has the `linker-signed` flag, which should result in any operations breaking the signature to resign it.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -60,7 +60,8 @@ task :build do
   sh %Q(rm -fr #{CLI_DIR})
   sh %Q(mkdir -p "#{CLI_DIR}bin")
   sh %Q(mkdir -p "#{CLI_DIR}lib")
-  sh %Q(lipo -thin #{Hardware::CPU.arch} lib_InternalSwiftSyntaxParser.dylib -output #{CLI_DIR}lib/lib_InternalSwiftSyntaxParser.dylib)
+  ARCH = `uname -m`.chomp
+  sh %Q(lipo -thin #{ARCH} lib_InternalSwiftSyntaxParser.dylib -output #{CLI_DIR}lib/lib_InternalSwiftSyntaxParser.dylib)
   sh %Q(cp SourceryJS/Resources/ejs.js #{CLI_DIR}bin)
   `mv #{BUILD_DIR}release/sourcery #{CLI_DIR}bin/`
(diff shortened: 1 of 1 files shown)


Cyberbeni:
Okay, I checked it and lipo doesn't invalidate the signature, it will be still valid after thinning:
```console
> codesign -dvvvv ./cli/lib/lib_InternalSwiftSyntaxParser.dylib
Executable=/Users/benedek.kozma/dev/Sourcery/cli/lib/lib_InternalSwiftSyntaxParser.dylib
Identifier=com.apple.lib_InternalSwiftSyntaxParser
Format=Mach-O thin (arm64)
CodeDirectory v=20400 size=28139 flags=0x0(none) hashes=873+2 location=embedded
VersionPlatform=1
VersionMin=720896
VersionSDK=721664
Hash type=sha256 size=32
CandidateCDHash sha256=831ba25aee4caab26ed071808c78e2ac677e2613
CandidateCDHashFull sha256=831ba25aee4caab26ed071808c78e2ac677e261338271504c80e526ace1f4503
Hash choices=sha256
CMSDigest=831ba25aee4caab26ed071808c78e2ac677e261338271504c80e526ace1f4503
CMSDigestType=2
Executable Segment base=0
Executable Segment limit=2949120
Executable Segment flags=0x0
Page size=4096
CDHash=831ba25aee4caab26ed071808c78e2ac677e2613
Signature size=4412
Authority=Software Signing
Authority=Apple Code Signing Certification Authority
Authority=Apple Root CA
Info.plist=not bound
TeamIdentifier=59GAB85EFG
Sealed Resources=none
Internal requirements count=1 size=88
```

krzysztofzablocki:
awesome @Cyberbeni, thank you for contributing!

krzysztofzablocki:
I'll just change so the stripping happens only in context of brew, since we don't want our pre-built binaries to be stripped

PR #977, opened 2021-07-22
Preserve attributes and specifiers in TypeName.name

### LINKED ISSUE #974: Attributes of closures inside extra closures are lost when unwrapping type name.

cheeseonhead:
# Motivation
The usage of `TypeName.name` is inconsistent for closures, it is unclear if it should only store the "base type" or the whole thing.

When I mean "base type", I mean the part of the closure excluding attributes, specifiers and optional info. So for `(@convention(c) () -> Void)?`, the base type would be `() -> Void`. And of course, whole thing means `(@convention(c) () -> Void)?`.

In old code, parsing an optional when the wrapped type is a closure preserves the outer parentheses in `name`. So for `(() -> Void)?` the name would be `(() -> Void)?` (exactly the same), notice how the optional `?` is kept.

However, when it comes to an attributed type, all the attributes are omitted from `name`. So for `@escaping () -> Void`, it becomes `() -> Void`; and `@convention(c) () -> Void` becomes `() -> Void`.

This creates the question, is `name` "unwrapped" or not? And by "unwrapped" I'm using the definition used by Sourcery to say "name that doesn't have any attributes, specifiers or optional info".

This hasn't been an issue because it's not valid to use `@escaping` within an optional tuple:
```swift
func test(_ c: (@escaping () -> Void)?) {} // Not allowed
```

But there's also `@convention(c)` which is very rare, but makes that kind of syntax allowed:
```swift
func test(_ c: (@convention(c) () -> Void)?) {} // Allowed
```

So in this case, the old code would parse the inner component first, which is `@convention(c) () -> Void` and it's treated as an attributed type, so the `name` omits attributes and become `() -> Void`.

Then it gets passed back up the chain to the code handling Optional. Since this is a closure, it's considered `needsWrapping`, and therefore it adds `()?` around it.

So the final name ends up being `(() -> Void)?` which is completely wrong as it's not the completely unwrapped type name nor is it the entire thing including attributes.

# Proposal
Follow the behavior of `name` pretty much everywhere else, basically it's the original source code that you see, including attributes, specifiers, and optional data.

Then we can use `unwrappedTypeName` property to track the name without any of the unneeded information.

# Additional Changes
The old code was modifying `TypeName.name` specifically at `MethodParameter` and `ClosureParameter` by prepending `inout` to it. This is another proof that we expect `name` to include all those info.

And with the proposal change, `inout` is now automatically included as part of `TypeName.name` during parsing, this means `MethodParameter` and `ClosureParameter` no longer needs to do more post-processing.

This also make sense that we put `inout` property on `TypeName` itself, since it's already in its `name`. Then change `MethodParameter` and `ClosureParameter`'s `inout` to computed properties referencing `TypeName.inout`.

### DIFF
--- a/SourceryRuntime/Sources/AST/TypeName/TypeName.swift
+++ b/SourceryRuntime/Sources/AST/TypeName/TypeName.swift
@@ -133,34 +133,16 @@ import Foundation
 
     /// Prints typename as it would appear on definition
     public var asSource: String {
-        // TODO: TBR special treatment
-        let specialTreatment = isOptional && name.hasPrefix("Optional<")
-
-        var description = (
-          attributes.flatMap({ $0.value }).map({ $0.asSource }) +
-          modifiers.map({ $0.asSource }) +
-          [specialTreatment ? name : unwrappedTypeName]
-        ).joined(separator: " ")
-
-        if let _ = self.dictionary { // array and dictionary cases are covered by the unwrapped type name
-//            description.append(dictionary.asSource)
-        } else if let _ = self.array {
-//            description.append(array.asSource)
-        } else if let _ = self.generic {
-//            let arguments = generic.typeParameters
-//              .map({ $0.typeName.asSource })
-//              .joined(separator: ", ")
-//            description.append("<\(arguments)>")
-        }
-        if !specialTreatment {
-            if isImplicitlyUnwrappedOptional {
-                description.append("!")
-            } else if isOptional {
-                description.append("?")
-            }
-        }
-
-        return description
+        let result = TypeName.asSource(
+            name: name,
+            unwrappedTypeName: unwrappedTypeName,
+            isOptional: isOptional,
+            isImplicitlyUnwrappedOptional: isImplicitlyUnwrappedOptional,
+            `inout`: `inout`,
+            attributes: attributes,
+            modifiers: modifiers)
+        
+        return result
     }
 
     public override var description: String {
@@ -229,5 +211,39 @@ extension TypeName {
             Log.astWarning("Unknown type, please add type attribution")
         }
         return TypeName(name: "UnknownTypeSoAddTypeAttributionToVariable", attributes: attributes)
+    }
+    
+    public static func asSource(name: String, unwrappedTypeName: String, isOptional: Bool, isImplicitlyUnwrappedOptional: Bool, `inout`: Bool, attributes: AttributeList, modifiers: [SourceryModifier]) -> String {
+        // TODO: TBR special treatment
+        let specialTreatment = isOptional && name.hasPrefix("Optional<")
+
+        var description = [
+            `inout` ? "inout" : "",
+            attributes.asSource,
+            modifiers.asSource,
+            specialTreatment ? name : unwrappedTypeName
+        ].filter { !$0.isEmpty }.joined(separator: " ")
+
+        if !specialTreatment {
+            if isImplicitlyUnwrappedOptional {
+                description.append("!")
+            } else if isOptional {
+                description.append("?")
+            }
+        }
+
+        return description
+    }
+}
+
+public extension AttributeList {
+    var asSource: String {
+        flatMap(\.value).map(\.asSource).sorted().joined(separator: " ")
+    }
+}
+
+public extension Array where Element == SourceryModifier {
+    var asSource: String {
+        map(\.asSource).sorted().joined(separator: " ")
     }
 }
--- a/SourceryRuntime/Sources/AST/TypeName/TypeName.swift
+++ b/SourceryRuntime/Sources/AST/TypeName/TypeName.swift
@@ -130,7 +130,15 @@ import Foundation
     
     /// Whether or not this type is an inout parameter.
     public var `inout`: Bool
-
+    
+    /// Indicating whether or not this type would need to be wrapped inside parentheses
+    /// in order to have proper type resolution precedence.
+    ///
+    /// For example, `@escaping () -> Void` needs to be wrapped in order to become optional `(@escaping () -> Void)?`
+    public var needsPrecedenceWrapping: Bool {
+        return isClosure || isProtocolComposition
+    }
+    
     /// Prints typename as it would appear on definition
     public var asSource: String {
         let result = TypeName.asSource(
@@ -140,17 +148,14 @@ import Foundation
             isImplicitlyUnwrappedOptional: isImplicitlyUnwrappedOptional,
             `inout`: `inout`,
             attributes: attributes,
-            modifiers: modifiers)
+            modifiers: modifiers,
+            needsPrecedenceWrapping: needsPrecedenceWrapping)
         
         return result
     }
 
     public override var description: String {
-       (
-          attributes.flatMap({ $0.value }).map({ $0.asSource }) +
-          modifiers.map({ $0.asSource }) +
-          [name]
-        ).joined(separator: " ")
+       asSource
     }
 
 // sourcery:inline:TypeName.AutoCoding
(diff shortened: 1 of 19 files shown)


SourceryRuntime/Sources/AST/TypeName/TypeName.swift
```diff
@@ -125,45 +127,35 @@ import Foundation
...
-
+    
+    /// Whether or not this type is an inout parameter.
+    public var `inout`: Bool
+    
+    /// Indicating whether or not this type would need to be wrapped inside parentheses
+    /// in order to have proper type resolution precedence.
+    ///
+    /// For example, `@escaping () -> Void` needs to be wrapped in order to become optional `(@escaping () -> Void)?`
+    public var needsPrecedenceWrapping: Bool {
```
krzysztofzablocki, line 138: nice🔥

SourceryTests/Parsing/FileParser + MethodsSpec.swift
```diff
@@ -49,7 +49,7 @@ class FileParserMethodsSpec: QuickSpec {
                     
                     expect(method).to(equal(Method(name: "fooInOut(some: Int, anotherSome: (inout String) -> Void)", selectorName: "fooInOut(some:anotherSome:)", parameters: [
                         MethodParameter(name: "some", typeName: TypeName(name: "Int")),
-                        MethodParameter(name: "anotherSome", typeName: TypeName.buildClosure(ClosureParameter(typeName: TypeName.String, isInout: true), returnTypeName: .Void))
+                        MethodParameter(name: "anotherSome", typeName: TypeName.buildClosure(ClosureParameter(typeName: TypeName(name: "inout String", unwrappedTypeName: "String", isInout: true)), returnTypeName: .Void))
```
krzysztofzablocki, line 52: can you add a builder `inout` generator? in same place I added those String etc extensions

that way you could write it as `TypeName.String.inOut`

krzysztofzablocki (approved):
Makes sense to me, also this is great work 👏 ! 
thanks for adding all the new tests 🙇 

I think you'll need to update mockable template to use updated AST info for the changes to pass tests

SourceryTests/Parsing/FileParser + MethodsSpec.swift
cheeseonhead: Created!

cheeseonhead:
I discovered a regression with `??`.

`Method` uses `TypeName`'s `asSource` property to generate its signature, and with the new definition introduced in this PR, we can't reliably reconstruct `??`.

Consider this example `func foo(bar: String??)`. Let's focus on `bar`'s `TypeName`. What should its `unwrappedTypeName` be? `String` or `String?`?

If it's the former, it fits the consistency and fits the definition of this PR, but the problem is that `asSource` has no information to reconstruct two `?`s since we only have one `isOptional` property. We have no concept of nesting when it comes to `?`.

If it's supposed to be `String?` then the question becomes "what exactly is unwrappedTypeName"?

Wondering what are your thoughts?

krzysztofzablocki:
I think I'd expect it to unwrap one level, mostly because of the context you usually use that property, e.g. in case of mock you'd unwrap one level so that you can do force variant of the underlying optional.

Purely from language perspective `TypeName` would allow nesting recursively and that's what SwiftSyntax parser returns but when you start thinking from the use-case perspective it would be pretty hard to work with in templating context and would break a lot of existing codebases

cheeseonhead:
After really analyzing `AutoMockable` I totally agree that it should just unwrap one level. I doubt it will be useful beyond that in the real world. So I've reconsidered my initial definition and came up with better ones (in my humble opinion).

Below is my proposal, and I think I'll have to make these changes in another PR and close this one.

## Summary
1. `unwrappedTypeName` now unwraps one optional level only if there is one. Wraps the result in `()` if needed. Strips context sensitive keywords like `@autoclosure`, `inout`, `@escaping`, `where`.
2. Introduce new property `pureTypeName` that strips out context sensitive keywords from the `TypeName` just like above, but without unwrapping any optional layers.
3. `name`, `description`, `asSource` will become the same value. They will have slight clean ups from original source code like removing redundant `()`. Other than that they are pretty much the same as original source code.

## Details
UnwrappedTypeName is now a property that returns a `String` that represents this `TypeName` with one less layer of `Optional`. For example, `Int?` unwrapped would become `Int`. This property will also automatically wrap `TypeName` that require parentheses, so `() -> Void` will become `(() -> Void)`. We do this purely because of the context this is most likely used in, and it can make people's lives easier. Also adding extra parentheses has no code impact.

This property also strips any context sensitive keywords like `inout`, `@escaping`, `@autoclosure` and `where` statements. The reason is because the way this property is used usually gets moved around to different contexts, from parameter to function return type or from function return type to variable types, etc. Keeping these keywords in this property is just not that useful normally, and if they keywords are really needed, it should be easy to add with `isXXXXX` properties.

Here are some examples of context sensitive keywords being removed when calling `unwrappedTypeName`:
1. `@escaping () -> Void` -> `(() -> Void)`
2. `inout (() -> Void)?` -> `(() -> Void)`
3. `Int where T: C` -> `Int`
4. `Int? where T: C` -> `Int`
5. `@convention(c) @escaping () -> Void` -> `(@convention(c) () -> Void)`

In fact, this is SO useful that I think we need to introduce a new property just to provide this function without unwrapping an optional layer. Without this, the `AutoMockable` template can't really mock methods that have `where` clause, since it uses `name`, it dumbly adds `where` to the `implicitReturnAReturnValue`'s type:
```swift
    //MARK: - implicitReturn<T>

    var implicitReturnACallsCount = 0
    var implicitReturnACalled: Bool {
        return implicitReturnACallsCount > 0
    }
    var implicitReturnAReceivedA: T?
    var implicitReturnAReceivedInvocations: [T] = []
    var implicitReturnAReturnValue: String! where T: Collection // THIS IS WRONG!
    var implicitReturnAClosure: ((T) -> String?)?

    func implicitReturn<T>(a: T) -> String! where T: Collection {
        implicitReturnACallsCount += 1
        implicitReturnAReceivedA = a
        implicitReturnAReceivedInvocations.append(a)
        return implicitReturnAClosure.map({ $0(a) }) ?? implicitReturnAReturnValue
    }
```

Finally, I propose we unify `name`, `description`, and `asSource` to always produce the same value. The reason `name` and `description` need to be the same is because Stencil templates allow you to write `{{ typeName }}` and it gets the description, but in EJS you have to use `typeName.name`. The current implementation makes `description` prepend attributes in front of `name` again even though `name` should have everything already.

As for `asSource`, as the name implies, it should be equivalent to the source, so I don't see a reason why it should be different from `name`.

krzysztofzablocki:
@cheeseonhead hey, why did you close the PR? 

it all sounded good to me except we'd probably want separated var for this logic `This property will also automatically wrap TypeName that require parentheses, so () -> Void will become (() -> Void)` because whether you need wrapping or not is context specific

cheeseonhead:
Because I'm not exactly sure when I'll be able to finish it, I did an audit of various possible permutations `inout`, `Optional`, `closure`, `attributes`, etc, and there are a lot of changes I need to make.

I didn't want to keep a PR open for too long while I'm working on it, it feels like clutter. I will reopen this once I have something new for you to review!

PR #979, opened 2021-08-09
Corrects args name in the writing-templates doc

jcislinsky:
Previously, it was documented as "arguments" but it's implemented as "argument".


krzysztofzablocki:
thanks

PR #981, opened 2021-08-13
Update Type.uniqueMethodFilter(_:_:)

rkreutz:
I was facing an issue where I had two methods with the same name but they had different return types but `type.allMethods` would only list one of them (the first) and `Type.uniqueMethodFilter(_:_:)` seems to be the culprit, since it's not checking the return types.

An example of what I was doing:
```
protocol MyProtocol: AutoMockable {

    // Sync version
    func method() -> Element?

    // Async version
    func method() -> Observable<Element>
}
```
Then, in the stencil if you call:
```
{{ type.allMethods }}
```
You'll notice that just the first declaration is listed, because both methods have the same name, but have different return types (which Swift allows to)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Sourcery CHANGELOG
 
+## 1.5.1
+## Fixes
+- Fixing `Type.uniqueMethodFilter(_:_:)` so it compares return types of methods as well.
+
 ## 1.5.0
 ## Features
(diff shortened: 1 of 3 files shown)


krzysztofzablocki:
@rkreutz looks good, would you mind adding test / changelog for it?

rkreutz:
@krzysztofzablocki done

krzysztofzablocki:
🙇  thanks for contributing

PR #982, opened 2021-08-24
Orders TypeName attributes

juhanh:
`TypeName.attribute` is a `Dictionary` which does not guarantee an ordering.
If a `TypeName` has several different attributes the ordering of these in generated code is not guaranteed.
As an example: having both `@autoclosure` and `@escaping` arguments on a closure in a file that is used to generate code will result the generated file being regenerated with both `@autoclosure @escaping` and `@escaping @autoclosure` orderings.

This pull request adds sorting to:
- `TypeName.asSource`
- `TypeName.description`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Sourcery CHANGELOG
 
+## 1.5.2
+## Fixes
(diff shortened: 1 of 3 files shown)


krzysztofzablocki (approved):
awesome, thanks!

PR #997, opened 2021-10-05
Add subspec for Sourcery.podspec to only include bin and lib folders

fabianehlert:
### Description
* This PR adds the ability to include only the `bin` and `lib` directories of a Sourcery release zip when installing via Cocoapods.
* This is achieved using a subspec inside the main `Sourcery.podspec` which configures `preserve_paths = 'bin', 'lib'`.
* Installation via `pod 'Sourcery'` remains untouched but users can optionally append `, :subspecs => ['CLI-Only']` to omit any directories inside `$PODS_ROOT/Sourcery/` not strictly required for using `sourcery`.

### Use case
We're working with an iOS project where dependencies are managed via Cocoapods and the content of `/Pods` is checked into the repository. When migrating our Sourcery installation from Mint over to Cocoapods, we've noticed that some files and directories would get checked in that we would not like to include (e.g. the __MACOSX directory and Sourcery.docset).

We had first thought about manually removing not needed files in a post-install script but prefer to contribute to and make this available as part of Sourcery as perhaps other users might find this useful too.

### Potential issues
* When additional, required directories are added in the future, the subspec needs to be modified or otherwise the installation might fail for users of the subspec.

I'm looking forward to hearing your feedback. I'm happy to make further adjustments if desired 🙂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Sourcery CHANGELOG
 
+## Master
+## New Features
+- Added `CLI-Only` subspec to `Sourcery.podspec` [#997](https://github.com/krzysztofzablocki/Sourcery/pull/997)
+
+---
+
 ## 1.6.0
 ## Fixes
 - Update dependencies to fix build on Xcode 13 and support Swift 5.5 [#989](https://github.com/krzysztofzablocki/Sourcery/issues/989)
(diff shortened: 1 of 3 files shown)


krzysztofzablocki:
looks great to me, can you only change the name to `CLI-Only` since that's more accurate?

Issue #1024, opened 2022-02-06
how to use argument.autoMockableTestableImports and argument.autoMockableImports ??


alexvbush:
hi, I'm trying to have my project main target be and other dependencies be automatically imported into my auto generated mocks file. I see that the default template has `autoMockableTestableImports` and `autoMockableImports` argument params in it but I can't figure out how to use it?

krzysztofzablocki:
you pass them in your sourcery configuration file as arguments, [example configs are in docs](http://merowing.info/Sourcery/usage.html)

alexvbush:
thank you @krzysztofzablocki 

I created the following config but it doesn't seem to work:

```yml
sources:
  - ../MyiOSTarget
templates:
  - ../Pods/Sourcery/Templates/AutoMockable.stencil
output:
  ../MyiOSTargetTests/AutoGeneratedMocks.swift
args:
  autoMockableTestableImports: MyiOSTarget
  autoMockableImports: MyLibraryTarget
```

Nikoloutsos:
@alexvbush 
FYI at any given moment write the following in the stencil file to get what the arguments have (good for debugging)
```stencil
{{argument]]
```
So if you do that you will realise your arguments are there. But why they are not playing correctly?
That's because if you take a closer look to the implementation of autoMock.stencil **a for loop is used**.

And for loop works only for arrays. So nothing happens for your single values `MyiOSTarget` and `MyLibraryTarget`.

In order to make it work you have to do the following:

```yml
sources:
  - ../MyiOSTarget
templates:
  - ../Pods/Sourcery/Templates/AutoMockable.stencil
output:
  ../MyiOSTargetTests/AutoGeneratedMocks.swift
args:
  autoMockableTestableImports: ["MyiOSTarget"]
  autoMockableImports: ["MyLibraryTarget"]
```

alexvbush:
Thank you @Nikoloutsos ! it fixed it!

jpeckner:
@artem-alexandrov I know this is an old post, but in case it helps, this works for me on the command line:

```
$ sourcery                                                                                              \
  --sources MyAppSources                                                                                \
  --templates Path/To/AutoMockable.stencil                                                              \
  --args autoMockableImports="Combine",autoMockableImports="Foundation",autoMockableImports="UIKit"     \
  --output MyOutputDir
```

PR #1030, opened 2022-02-17
Add documentation about adding imports to AutoMockable.stencil

alexvbush:
This change adds documentation to clarify how to add imports when using AutoMockable template. It addresses the question in https://github.com/krzysztofzablocki/Sourcery/issues/1024


krzysztofzablocki:
🙇

PR #1035, opened 2022-02-28
Parsing .swiftinterface form xcframework

rehsals:
*Problem*
Issue: https://github.com/krzysztofzablocki/Sourcery/issues/87
Our project is really big so we decided to use it's dependencies as binary XCFrameworks.
The problem is to mock types declared in modules from such binary frameworks which we manage with CocoaPods.
The idea to work this limitation around was to scan `.swiftinterface` file inside the xcframework.

*Line of thought*
At first I considered I could use `sources` key to set this up. But `sources` is mutually exclusive
with `project` which is an issue in our case. I gave it a try to make `sources` usable along with `project`
but it didn't work quite well. The thing is we need to use module name to namespace types in mocks. The case of matching
type names is not so rare. So `sources` was not entirely working. Another consideration was to scan `.xcodeproj` to find a path
to xcframework and add it to sources list. But CocoaPods makes target so placeholder-ish that there is no way that I could find
to extract that path.

*Proposed solution*
Finally I introduced optional key `xcframework` to configuration.

```
project:
  - file: "Pods/Pods.xcodeproj"
    target:
      - name: "Alamofire"
        xcframework: "Pods/Alamofire/Alamofire.xcframework"
```
This path is added to target and gets scaned. Some messy fiddling with paths inside `.xcframework` file could be avoided but
for now it just works. Also the `xcframework` key could be made an array (not done yet)

This Pull Request might not be the the right way to do it, but might as well be the starting point for a discussion.
I am sure there is a better way to achieve this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Sourcery CHANGELOG
 
+## Main
+## New Features
+- Adds `xcframework` key to `target` object in configuration file to enable processing of `swiftinterface`
+
 ## 1.7.0
 
 ## New Features
--- a/Sourcery/Configuration.swift
+++ b/Sourcery/Configuration.swift
@@ -16,6 +16,7 @@ public struct Project {
         public struct XCFramework {
 
             public let path: Path
+            public let swiftInterfacePath: Path
 
             public init(rawPath: String, relativePath: Path) throws {
                 let frameworkRelativePath = Path(rawPath, relativeTo: relativePath)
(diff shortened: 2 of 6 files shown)


krzysztofzablocki:
this looks good to me, it solves the problem at hand and the code doesn't look bad or complicated. 

Let's just add support for the array case 🔥

rehsals:
Great! Thanks. Then I'll spend some time making this PR mergeable

rehsals:
So I guess the PR is ready for review

krzysztofzablocki:
thanks!

Issue #1009, opened 2021-11-04
Threading crashes in Composer.resolveType


erichoracek:
We're experiencing some random threading crashes with Sourcery in the `Composer.resolveType` implementation.

Specifically, while in the concurrent logic in the `Array.parallelPerform` sections in the `Composer.uniqueTypesAndFunctions` method we're experiencing 
nondeterministic `EXC_BAD_ACCESS`s. It would appear that there's something thread-unsafe going on in the parallel performs.

We have reverted to Sourcery 1.3.1 which doesn't seem to experience this issue.

Here's the stack trace from a sample crash log:
```
Crashed Thread:        0  Dispatch queue: com.apple.root.default-qos

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000030
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [45000]

Thread 0 Crashed:: Dispatch queue: com.apple.root.default-qos
0   libobjc.A.dylib               	0x00007fff203484af objc_release + 31
1   sourcery                      	0x0000000101594e20 @objc TypeName.__ivar_destroyer + 48
2   libobjc.A.dylib               	0x00007fff2034fafb object_cxxDestructFromClass(objc_object*, objc_class*) + 83
3   libobjc.A.dylib               	0x00007fff20348632 objc_destructInstance + 94
4   libobjc.A.dylib               	0x00007fff20348598 _objc_rootDealloc + 62
5   sourcery                      	0x00000001015a0f52 static Composer.resolveType(typeName:containingType:state:) + 8306
6   sourcery                      	0x00000001015a1b07 closure #3 in static Composer.uniqueTypesAndFunctions(_:) + 279
7   sourcery                      	0x00000001015ac004 partial apply for thunk for @callee_guaranteed (@guaranteed Type) -> () + 20
8   sourcery                      	0x00000001015ac0c7 partial apply for specialized closure #1 in Array.parallelPerform(transform:) + 87
9   libswiftDispatch.dylib        	0x00007fff315da5d1 partial apply for thunk for @callee_guaranteed (@unowned Int) -> () + 17
[...]

PR #1044, opened 2022-03-28
Resolving types of typealiases in a single thread

igor-savelev-at-bumble:
For the past several months we experienced occasional crashes of Sourcery with `EXC_BAD_ACCESS`. To get more details what might be causing it, I enabled Thread sanitizer and run all tests from the repo. This is what I found:

Screenshot 2022-03-25 at 13 42 53

It appears that instance of one typealias could be used to resolve type of another typealias. And, becase `Typealias` is a class, it's possible that some typealias is mutated at the same time when another thread reads its property. I found similar situation described in one of the issues: #1009. For us, the crash was reproducing both on 1.6.1 and 1.7.0.

To resolve the issue, I simply replaced `parallelPerform` with a regular `forEach`. This led to a slight performance degradation (about 4%), but stability seems to be more important in this case.

### DIFF
--- a/SourceryRuntime/Sources/Composer.swift
+++ b/SourceryRuntime/Sources/Composer.swift
@@ -284,7 +284,7 @@ public enum Composer {
 
         /// Resolve typealiases
         let typealiases = Array(state.unresolvedTypealiases.values)
-        typealiases.parallelPerform { alias in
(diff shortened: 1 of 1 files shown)


krzysztofzablocki:
thanks @igor-savelev-at-bumble

Issue #1045, opened 2022-03-28
Sourcery 1.8.0 isn't buildable on non-supported Xcode 13.3 Macs


AF-cgi:
Thanks a lot for the new release and the fast fix for integrating the new SwiftSyntax release. But now it's required to use Xcode 13.3 to build Sourcery. Xcode 13.3 isn't available for the same Macs like Xcode 13.2.1 did this. Do you see a chance to add this support again?

krzysztofzablocki:
we won't be adding backward compatibility, we are always moving forward since we are also integrating newer SwiftSyntax versions and people can stay on older version until they are ready to switch

AF-cgi:
I see your point. But Xcode 13.3 requires a Mac running macOS Monterey 12 or later. That means that new Sourcery features are only available for Macs with macOS Monterey 12 or later. I know, that SwiftLint solve this point.

liamnichols:
The prebuilt binary of 1.8.0 (downloaded from the releases page, or via CocoaPods/Homebrew) should work on any supported macOS version (10.12+ iirc), this is only about compiling the tool (i.e using it with Mint).

Given that pre 1.8.0 always discouraged using Mint, I don't think this would be a widespread issue. 

As a temporary workaround, can you use a precompiled version instead of Mint on your CI infrastructure that doesn't have Xcode 13.3 installed?

AF-cgi:
Our CI infrastructure doesn't support macOS 12 right now. From that point the local and CI builds are diverge. There is also some Apple hardware which doesn't supports macOS 12.
We use also Mint as CLI dependency manager, because we use many other CLIs. From that perspective it would be great to find a solution to solve it.
As a temporary workaround it's possible to use a precompiled version. But that's not the preferred way.
My Question on that point how other CLIs solve this, f.e. SwiftLint? SwiftLint is buildable over Mint with Xcode 13.3 and Xcode 13.2.1. I thing it's a also possible to install SwiftLint via Home-brew on multiple macOS versions.

liamnichols:
So I thought a bit harder about this, maybe #1049 is what we want. It's still Sourcery using SwiftSyntax 5.6 with the statically linked lib_internalSwiftSyntaxParser, but it allows for building using Xcode 13.x.

This approach would probably be good up until we want to use Swift 5.6 specific language features in the codebase, which is probably far enough away to give teams time to move to Xcode 13.3 (i.e later this year).

AF-cgi:
[...]

PR #1046, opened 2022-03-31
chore: makes sourcery buildable starting from Xcode 13.0 / Swift 5.5 (#1045)

AF-cgi:
Solves the issue of #1045. Further the size of the build artifact has been reduced dramatically. Now Sourcery is buildable starting from Xcode 13.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Sourcery CHANGELOG
 
+## 1.9.0
+## Internal Changes
+- Swift 5.5 and Xcode 13.0 is now required to build the project
+- `lib_internalSwiftSyntaxParser` has removed
+
 ## 1.8.0
 ## New Features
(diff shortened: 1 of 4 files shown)


krzysztofzablocki:
This isn't correct, you are removing dependency on libSwiftSyntax and the work that @liamnichols did to make it static instead of dynamically linked, this will break not only Mint support but all of the versions, it will fail or work depending on what Xcode people have on their machines.

liamnichols:
If we do want to go ahead with this, we'd want to keep using StaticInternalSwiftSyntaxParser, which can be done relatively simply ... I had a go in #1049 but there are a few things to think about (see the PR description)

liamnichols:
If we do want to go ahead with this, we'd want to keep using StaticInternalSwiftSyntaxParser, which can be done relatively simply ... I had a go in #1049 but there are a few things to think about (see the PR description)

PR #1075, opened 2022-07-12
Build .swifttemplate in release mode

jimmya:
Makes sure `assert`'s don't go off when generating code using a .swifttemplate


SourcerySwift/Sources/SwiftTemplate.swift
```diff
@@ -232,7 +232,7 @@ open class SwiftTemplate {
             "--sdk", "macosx",
             "swift",
             "build",
-            "-Xswiftc", "-Onone",
```
jimmya, line 235: I wasn't quite sure why this was specifically stated here

krzysztofzablocki (approved):

Issue #1023, opened 2022-02-04
[Suggestion] Add support for the new SPM build plugins


BrentMifsud:
Apple just announced that the new SPM build commands is going to be available in the next version of xcode. Its currently in beta, but adding support for this feature would be an awesome addition to Sourcery

```
Swift Package Manager
New Features
Swift Packages now support build tool plugins, as defined in [SE-0303](https://github.com/apple/swift-evolution/blob/main/proposals/0303-swiftpm-extensible-build-tools.md) and [SE-0325](https://github.com/apple/swift-evolution/blob/main/proposals/0325-swiftpm-additional-plugin-apis.md). This allows packages to define plugins that can specify tools that should run during a build operation, for example to generate source code. This is supported in both swift package and in Xcode’s support for packages. (79876749)

The swift package command now supports command plugins, as defined in [SE-0332](https://github.com/apple/swift-evolution/blob/main/proposals/0332-swiftpm-command-plugins.md). This allows Packages to define commands that can be invoked using the swift package command line to perform custom actions on the package. (82895553)
```

krzysztofzablocki:
@BrentMifsud until they support R-Path rewrites you can't build Sourcery properly with SPM only

pchmelar:
Is this possible now (after changes in https://github.com/krzysztofzablocki/Sourcery/pull/1037) ? 🤔

krzysztofzablocki:
yeah it should be possible now, anyone wants to contribute that integration? 😉 
I'm pretty low on having personal time between work and conferences

cal-os:
+1 would love to see this as a feature with apple's new plugin build tool system

doozMen:
I would love to work on this and might be very interesting in my project at work. It would be my first time contributing. Any tips before I start?

krzysztofzablocki:
@doozMen 🙇 shouldn't require anything special given Sourcery now builds purely with SPM. The 2 WWDC sessions about plugins cover everything I think, but if you run into issues you can report here and I can get Apple folks to look at it if we can't figure it out ourselves

pkurzok:
I was trying out building a Command Plugin using Sourcery as a dependency. But I get the following error when the Packages are resolved:

`error: multiple targets named 'lib_InternalSwiftSyntaxParser' in: 'sourcery', 'swiftlint'`

My Package File looks like this:

```
let package = Package(
    ...
    dependencies: [
[...]

PR #1084, opened 2022-08-03
Update StencilSwiftKit

redryerye:
Updated StencilSwiftKit to v2.10.1 to fix the issue where building Sourcery as a Command Plugin fails. https://github.com/krzysztofzablocki/Sourcery/issues/1023#issuecomment-1180284670
The issue was caused by [Komondor v1.1.3](https://github.com/shibapm/Komondor), which StencilSwiftKit used to rely on, [added extra dependencies that are unnecessary](https://github.com/shibapm/Komondor/compare/1.1.3...1.1.4).
v2.10.1 release for StencilSwiftKit now pins Komondor to the old version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Sourcery CHANGELOG
 
+## 1.8.3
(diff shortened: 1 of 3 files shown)


krzysztofzablocki (approved):
thank you!

PR #1096, opened 2022-09-22
SPM Plugin: Display target name when .sourcery.yml is not found

grennis:
When running the plugin on multiple targets, if the yml file was not found, it's helpful to which target name has the missing file. And then, continue with the next target instead of exiting out completely.

I don't think this should have a changelog entry because I'm updating code that hasn't been released yet and already has a changelog.


krzysztofzablocki:
sweet, thanks Greg!

Issue #664, opened 2018-08-17
Add option to store output in a file (json)


ilyapuchka:
Sourcekitten has the option to store structure output in the file so that it can be then used by other tools. We can do the same to enable other tools to act on types metadata.
@krzysztofzablocki @Antondomashnev thoughts?

krzysztofzablocki:
fine with me, we already have the data and could just as well generate the boilerplate for json generation

Antondomashnev:
Sorry for the late reply. If that isn't much of a work + could be useful for other tools (do we have any request for it btw?) I think it's a good enhancement.

1is10:
Hi guys,
I just want to realise json output to std, i read contribution guide and found this topic 🙌 

I have a POC here https://github.com/1is10/Sourcery/pull/1

is it what are you talking about or should I create another topic to discuss that one?

krzysztofzablocki:
@1is10 what's your use-case for json output nowadays? we moved Sourcery to work as framework so one can build tools on top of Sourcery without using intermediate formats like json

1is10:
My first thoughts about json stdout was
- dry run, it's useful for many cases for example to count autogenerated code lines, or by CI to check errors during code generation, without write on disk
- post processing output without modifying templates in any script language
- building tools in other languages, without linking sourcery framework directly

krzysztofzablocki:
@1is10 ok, I'd be fine with adding feature like that it we can isolate the functionality not to add too much complexity to the tool

1is10:
i implemented it with `--dry` flag, and made a `pull request` here + ready to fix any issues related about isolation and overall logic, tested locally + added some tests for `--dry` option.

pull request link: 
https://github.com/krzysztofzablocki/Sourcery/pull/1097

PR #1097, opened 2022-09-26
Dry run support (with json output)

1is10:
https://github.com/krzysztofzablocki/Sourcery/issues/664


SourceryTests/Models/TypedSpec.generated.swift
```diff
@@ -315,4 +315,4 @@ class TypedSpec: QuickSpec {
             }
         }
     }
-}
+}
```
krzysztofzablocki, line 318: where is the lack of empty EOF line coming from? do you have some post-processing that happens outside of Sourcery?

krzysztofzablocki (commented):
looking good

SourceryTests/Models/TypedSpec.generated.swift
1is10: Sorry, my bad, i will be more careful in future. Cause of this, is that i run temporary version of sourcery by mistake inside repo. 

That was early version that trims end lines, that bug don't exist in current version.

i rechecked code, removed changes to package.resolved & regenerated SourceryRuntime code with sourcery binary.

krzysztofzablocki:
thanks for contributing 🙇

1is10:
Thank you for sourcery 🙌

Issue #850, opened 2020-08-18
Custom Automockable annotation doesn't work for public protocols


Sieniek:
Command: sourcery --config ./.api_sourcery.yml

Does work:
`
// sourcery:file: AutoMockableMyTarget
protocol A { 
 func b()
}
`

Doesn't work – no mock is generated:
`
// sourcery:file: AutoMockableMyTarget
public protocol A { 
 func b()
}
`
.api_sourcery.yml
```
sources:
  - API
  - src
templates:
  - Tests/Mocks/API/Generated/CustomAutoMockable.stencil
output:
  Tests/Mocks/API/Generated/
```

CustomAutoMockable.stencil
```
{% macro swiftifyMethodName name %}{{ name | replace:"(","_" | replace:")","" | replace:":","_" | replace:"`","" | snakeToCamelCase | lowerFirstWord }}{% endmacro %}
{% macro methodThrowableErrorDeclaration method %}
    var {% call swiftifyMethodName method.selectorName %}ThrowableError: Error?
{% endmacro %}
{% macro methodThrowableErrorUsage method %}
        if let error = {% call swiftifyMethodName method.selectorName %}ThrowableError {
            throw error
        }
{% endmacro %}
{% macro methodReceivedParameters method %}
    {%if method.parameters.count == 1 %}
        {% call swiftifyMethodName method.selectorName %}Received{% for param in method.parameters %}{{ param.name|upperFirstLetter }} = {{ param.name }}{% endfor %}
        {% call swiftifyMethodName method.selectorName %}ReceivedInvocations.append({% for param in method.parameters %}{{ param.name }}){% endfor %}
    {% else %}
    {% if not method.parameters.count == 0 %}
        {% call swiftifyMethodName method.selectorName %}ReceivedArguments = ({% for param in method.parameters %}{{ param.name }}: {{ param.name }}{% if not forloop.last%}, {% endif %}{% endfor %})
        {% call swiftifyMethodName method.selectorName %}ReceivedInvocations.append(({% for param in method.parameters %}{{ param.name }}: {{ param.name }}{% if not forloop.last%}, {% endif %}{% endfor %}))
    {% endif %}
    {% endif %}
{% endmacro %}
{% macro methodClosureName method %}{% call swiftifyMethodName method.selectorName %}Closure{% endmacro %}
[...]

PR #1100, opened 2022-10-07
Add support for public protocols in AutoMockable template

gspiers:
## Description
In a project I'm working on we have some test helper modules that contain the mocks for protocols that are public. These test helper modules are imported into other unit test modules.

The built-in AutoMockable template didn't take into account access level of a protocol when generating the mocks.

I see from searching past issues that someone else had to create a custom template in #850.

I wanted to contribute back our improvements. The changes shouldn't create too much noise for someone updating to this template as the `internal` access level is continued to not be added to mocks.

Please let me know if this is in good shape to contribute back and if not what I can do to help.

By the way, really appreciate Sourcery, thanks for all the work on it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Sourcery CHANGELOG
 
+## Main
+## New Features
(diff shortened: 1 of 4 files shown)


krzysztofzablocki (approved):

krzysztofzablocki:
Thanks!

PR #1113, opened 2022-11-08
Fixed parsing of extensions and nested types in swiftinterface files

till0xff:


krzysztofzablocki (approved):

krzysztofzablocki:
🙇

krzysztofzablocki:
@till0xff had to reverse part of it in 1.9.2 since it doesn't work properly, I added additional test that surfaces the issue with your implementation, happy to merge a new implementation that works with both

Issue #1102, opened 2022-10-08
SPM usage of 1.9.0


pchmelar:
- I'm trying to add the 1.9.0 version via SPM and keep getting this error.
- Is it needed to use the exact commit of [swift-syntax](https://github.com/apple/swift-syntax) or can we switch to some of its releases?
```
error project: Failed to resolve dependencies Dependencies could not be resolved because 'shareddomain' depends on 'sourcery' 1.8.3..<2.0.0. 'sourcery' >= 1.8.3 cannot be used because package 'sourcery' is required using a stable-version but 'sourcery' depends on an unstable-version package 'swift-syntax' and no versions of 'sourcery' match the requirement {1.8.3..<1.9.0, 1.9.1..<2.0.0}.
```

krzysztofzablocki:
@pchmelar it needed an exact commit since they didn't officially release the new parser yet, you should be able to just point at Sourcery's exact commit of 1.9.0 same way I did with SwiftSyntax

pchmelar:
Thanks for the tip! I was using `from: "1.8.3"` as stated in the [documentation](https://github.com/krzysztofzablocki/Sourcery#installation). Now it's working as expected 🙂 

So for anyone else experiencing this issue, you have to add this to your `Package.swift` for now:
```
.package(url: "https://github.com/krzysztofzablocki/Sourcery.git", .revision("8aab07cbda5206a2625ff9736c065262d2804364"))
```


BTW wouldn't it be better to provide a configuration for the plugin in `sourcery.yml` instead of `.sourcery.yml`? Swift Packages in Xcode don't display hidden files, so it's not possible to edit the configuration directly from the Xcode.

pchmelar:
@krzysztofzablocki Hey 👋 We are experiencing issues in projects that are using both Sourcery and SwiftLint via SPM. Both libraries have a dependency on a new parser from swift-syntax, but each is pointing to a different source, resulting in SPM resolution error. Sourcery is using the exact commit `.revision("093e5ee151d206454e2c1950d81333c4d4a4472e")`, while SwiftLint is pointing to the latest pre-release `exact: "0.50800.0-SNAPSHOT-2022-12-29-a"`. Do you think we can point Sourcery to the same pre-release in order to mitigate this issue until the 0.50800.0 version is released? 🙏

krzysztofzablocki:
@pchmelar should be ok, mind sending over the PR with that?

pchmelar:
[...]

PR #1128, opened 2023-01-18
Bump swift-syntax to 0.50800.0-SNAPSHOT-2022-12-29-a

pchmelar:
As discussed in: https://github.com/krzysztofzablocki/Sourcery/issues/1102#issuecomment-1386498064


krzysztofzablocki:
I had to fix parser logic here since they changed it but pushing to master, thanks for contributing 🙇

PR #1130, opened 2023-01-20
Handle UTF-8 offsets as such

### LINKED ISSUE #1129: Crash when non-ASCII characters in comment
When parsing a source file with the following code, Sourcery 2.0 crashes (whatever the template is):
```swift
func foo() -> Int // ああ
```

vincentisambart:
Closes #1129

UTF-8 offsets were incorrectly used as character offsets so I fixed that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Sourcery CHANGELOG
 
+## x.x.x
+## Internal Changes
+- Fixed non-ASCII characters handling in source code parsing [#1130](https://github.com/krzysztofzablocki/Sourcery/pull/1130)
+
 ## 2.0.0
 ## Changes
 - sourcery:auto inline fragments will appear on the body definition level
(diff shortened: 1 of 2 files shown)


vincentisambart:
The failure does not seem to be due to my change, but to `SourceryTests/SourcerySpec.swift` referencing an absolute path not writable for most people.
https://github.com/krzysztofzablocki/Sourcery/blob/bf06f37585c37138e6f374902d8959a5c9be2001/SourceryTests/SourcerySpec.swift#L368
Changing it to something like `/tmp/sourcery-test.txt` seems to fix it. Should I add that change to this PR?

krzysztofzablocki:
Ah that's some left over test code 🙈 we can remove that

krzysztofzablocki:
brought to master, thanks for contributing 🙇

vincentisambart:
Thanks for reviewing and merging 😄

PR #1132, opened 2023-01-22
Support multibyte in annotation parsing

muukii:
I got a minor issue that fails to parse when codes are using multibyte, like following

```swift
enum Foo {
  case こんにちは(Int)
}
```

It crashes by out-of-bounds in a string due to making a substring from the offset.
Offset is technically bytes offset, it's not UTF8's offset.

I did only work on processing string with offset in a correct way.

I'm not an expert on string processing.
Please correct me in a better way.

Thank you for your super helpful tool.

### DIFF
--- a/SourceryFramework/Sources/Parsing/Utils/AnnotationsParser.swift
+++ b/SourceryFramework/Sources/Parsing/Utils/AnnotationsParser.swift
@@ -152,11 +152,13 @@ public struct AnnotationsParser {
         return documentation.reversed()
     }
 
-
+    
     func inlineFrom(line lineInfo: (line: Int, character: Int), stop: inout Bool) -> Annotations {
         let sourceLine = lines[lineInfo.line - 1]
-        var prefix = sourceLine.content.bridge()
-            .substring(to: max(0, lineInfo.character - 1))
+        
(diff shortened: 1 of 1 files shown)


krzysztofzablocki:
can you add a test for it?

muukii:
@krzysztofzablocki 
got it. but I'm not sure how the test managed in this project.
for this case, where should I add a test into?

krzysztofzablocki:
@muukii probably to `AnnotationsParserSpec`

art-divin:
Hello @muukii 🙌,

thank you very much for your contribution.

I have investigated this issue, and it seems like the reason of the crash is not about multibyte handling, which is fully supported, but about handing associated type with multibyte character identifier.

For example, the following code is parsed without the crash:

```swift
enum Foo {
  case こんにちは = 33
}
```

I'll rework your change set and create a new MR, but it is definitely inspired by your contribution.

Thank you 🙏

art-divin:
For observability: minimal reproducible test sample is this:
`enum Foo {case こんに(Int)}`

art-divin:
I have resolved this issue in #1178

PR #1143, opened 2023-02-23
Fix configuration file documentation link in ReadMe

ajevans99:
I was getting a 404 link. I think this new link is the proper location of the documentation.


krzysztofzablocki:
Ghost redirects are not great apparently

PR #1144, opened 2023-02-24
Fixed properties with composite type in mock template

pavel-trafimuk:
I found a small issue with composite types in mocks.

Example
```swift
protocol Prot1 {}
protocol Prot2 {}

// sourcery: AutoMockable
protocol Abc {
    var property: Prot2 & Prot1 { get }
}
```

Sourcery generates:
```swift
protocol Prot1 {}
protocol Prot2 {}

// sourcery: AutoMockable
class AbcMock: Abc {
    var property: Prot2 & Prot1 {
        get { return underlyingProperty }
        set(value) { underlyingProperty = value }
    }
    var underlyingProperty: Prot2 & Prot1!
}
```

But compiler doesn't like ``` var underlyingProperty: Prot2 & Prot1!```
Should be ```var underlyingProperty: (Prot2 & Prot1)!```

This PR fixes it

### DIFF
--- a/Templates/Templates/AutoMockable.stencil
+++ b/Templates/Templates/AutoMockable.stencil
@@ -154,7 +154,8 @@ import {{ import }}
         get { return {% call underlyingMockedVariableName variable %} }
         set(value) { {% call underlyingMockedVariableName variable %} = value }
     }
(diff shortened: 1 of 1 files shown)


krzysztofzablocki (approved):

krzysztofzablocki:
🙇

PR #1145, opened 2023-02-24
Added ability to wrap mocks under define

pavel-trafimuk:
Currently mocks don't have ability to be hidden for some purposes, e.g. release/production.
Yes, we can generate them only in unit-test targets, but this approach does not work in multi-modules real app:
when I want mocks from moduleA in tests of moduleB and moduleC.

So it's better to put generated mocks in moduleA, but wrap them under define, so they will not expose in release code.
To do that, I add new optional argument ```mockWrapDefine```

### DIFF
--- a/Templates/Templates/AutoMockable.stencil
+++ b/Templates/Templates/AutoMockable.stencil
@@ -1,6 +1,7 @@
 // swiftlint:disable line_length
 // swiftlint:disable variable_name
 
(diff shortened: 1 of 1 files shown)


krzysztofzablocki:
@pavel-trafimuk mind adding info to the template docs?

EnesKaraosman (approved):

pavel-trafimuk:
Please let me few days to update this MR

Issue #1142, opened 2023-02-22
running sorcery in buildtool plugin on multiple packages within a project results in error


swwol:
I'm trying to make a buildtool plugin which uses the sourcery binary to generate mocks. When using this plugin on a single package, it works fine. But if compiling a project where the plugin is used in more than one package I get this error:

accessing build database "/private/var/folders/bl/dpyystbj1szdr_rzp6p1fwq40000gn/T/SwiftTemplate/2.0.1/.build/arm64-apple-macosx/build.db": database is locked Possibly there are two concurrent builds running in the same filesystem location.

Any suggestions for a solution appreciated!

swwol:
A bit more info on this: sourcery seems to be generating some .swift files and writing them to the system temp directory when it runs.. I think the build is seeing these as input files which is causing issues. Are these files required for sorcery to run and if not is there a way to prevent them from being generated?

swwol:
Specifically these files are appearing in temp directory:

AccessLevel.swift		BytesRange.swift		Diffable.generated.swift	GenericRequirement.swift	Protocol.swift			Typealias.swift
Actor.swift			Class.swift			Diffable.swift			Import.swift			ProtocolComposition.swift	Typed.generated.swift
Annotations.swift		Closure.swift			Documentation.swift		JSExport.generated.swift	Struct.swift			Typed.swift
Array+Parallel.swift		Coding.generated.swift		Enum.swift			Log.swift			Subscript.swift			Variable.swift
Array.swift			Composer.swift			Equality.generated.swift	Method.swift			TemplateContext.swift
AssociatedType.swift		Definition.swift		Extensions.swift		Modifier.swift			Tuple.swift
Attribute.swift			Description.generated.swift	FileParserResult.swift		ParserResultsComposed.swift	Type.swift
AutoHashable.generated.swift	Dictionary.swift		Generic.swift			PhantomProtocols.swift		TypeName.swift

krzysztofzablocki:
do you have .swifttemplates?

swwol:
I do - (it's the .swifttemplate used by swiftymocky) I delved into the source code a little and kinda understand what is going on now.  Thinking of making a fork which allows an argument to set a custom build directory, which could be the plugins packageWorkDirectory. This should be different for every target so I think will resolve the issue I am having.  Do you think that would be a good way forward?

krzysztofzablocki:
a PR with optional build directory sounds good to me

vykut:
[...]

PR #1147, opened 2023-02-25
Add a new command line argument to allow setting an optional build directory for .swifttemplates

swwol:
what:

This adds a new command line argument to allow setting an optional build directory for .swifttemplates. When this is set the custom directory is used. If it's not set, then the system temp directory is used as before. 

why: 

As documented in [1142](https://github.com/krzysztofzablocki/Sourcery/issues/1142), when running several concurrent instances of sourcery there can be conflicts if they are all trying to write to the same temporary directory. 

This scenario in which this happened for me is after creating a swift package manager plugin to generate mocks, and running that plugin on multiple targets within the same project or package file.  This was causing build failures. 

Allowing the plugin to set this new command line argument to its own work directory has resolved this issue as each target the plugin operates on has its own work directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,7 @@ Inside a project/package that uses this command plugin, right-click the project 
 - `--version` - Display the current version of Sourcery
 - `--help` - Display help information
 - `--cacheBasePath` - Base path to the cache directory. Can be overriden by the config file.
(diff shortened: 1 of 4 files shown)


swwol:
Not sure why the validate_docs step is failing on the CI - same command is ok locally..

krzysztofzablocki:
I'll update it with next release, thanks!

PR #1159, opened 2023-04-03
[SPM] Update to latest stable swift-syntax release

bdbergeron:
A [stable version of swift-syntax for Swift 5.8](https://github.com/apple/swift-syntax/releases/tag/508.0.0) has now been released. This PR updates Sourcery's `Package.swift` file to help fix the issues around SPM complaining about the source-unstable version when resolving dependencies.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,7 +26,7 @@ let package = Package(
         .package(url: "https://github.com/kylef/PathKit.git", exact: "1.0.1"),
         .package(url: "https://github.com/SwiftGen/StencilSwiftKit.git", exact: "2.10.1"),
         .package(url: "https://github.com/tuist/XcodeProj.git", exact: "8.3.1"),
-        .package(name: "SwiftSyntax", url: "https://github.com/apple/swift-syntax.git", from: "508.0.0"),
+        .package(url: "https://github.com/apple/swift-syntax.git", from: "508.0.0"),
         .package(url: "https://github.com/Quick/Quick.git", from: "3.0.0"),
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -26,7 +26,7 @@ let package = Package(
         .package(url: "https://github.com/kylef/PathKit.git", exact: "1.0.1"),
         .package(url: "https://github.com/SwiftGen/StencilSwiftKit.git", exact: "2.10.1"),
         .package(url: "https://github.com/tuist/XcodeProj.git", exact: "8.3.1"),
-        .package(name: "SwiftSyntax", url: "https://github.com/apple/swift-syntax.git", branch: "0.50800.0-SNAPSHOT-2022-12-29-a"),
+        .package(name: "SwiftSyntax", url: "https://github.com/apple/swift-syntax.git", from: "508.0.0"),
```
jsorge, line 29: Thanks for doing this!

Running `swift package resolve` with this branch gives me the warning:
```warning: 'package(name:url:from:)' is deprecated: use package(url:from:) instead```
It looks like we can drop the `name: "SwiftSyntax",` from the beginning of the call there.

bdbergeron: Since this is just a deprecation error, I didn't feel the need to update this, but I'm happy to do so. 👍

krzysztofzablocki:
thanks!

bdbergeron:
@krzysztofzablocki would it be possible to get a new 2.0.2 release tagged with this change?

PR #1163, opened 2023-04-19
Update SwiftTemplate to make possible to edit Package.swift

k-o-d-e-n:
### Motivation:
Existing template engines are limited and inconvenient, I would like to use the full power of Swift when creating templates, in particular to connect third-party libraries.
### Solution:
All you need now is to disable overwriting of `Package.swift` to be able to edit it
### Impact on existing behavior:
None.

---------------

Crash with invalid URL in `task.currentDirectoryURL` has also been fixed and manifest has been updated to version 5.7.

An example of using a third-party library to generate code in Sourcery can be viewed [here](https://github.com/Everything-as-UI/SourcerySupport )


krzysztofzablocki:
nice 🙇

PR #1176, opened 2023-07-03
Fix enum case with rawValue includes trailing comment

### LINKED ISSUE #1154: Enum with rawValue parses comment
The rawValue of enum cases is parsed wrong in sourcey 2.0.1.
As you can see in the output, the comment "some comment" is part of the rawValue.
I would expect the comment not to be part of the rawValue.
Currently wrong: `zero = 0 // some comment`
Expected correct: `zero = 0`
Workaround until fixed: put the comment on an extra line

source.swift
```
public enum SomeEnum: Int {
    case zero = 0 // some comment
    case positive = 1
    case negative = -1
}
```

template.ejs
```
<%
function asdf() {
    const typesEnums = types.all.filter(type => type.kind === "enum")
    return typesEnums
        .map((theEnum) =>
            theEnum.cases.map((enumCase) =>
                `${enumCase.name} = ${enumCase.rawValue}`
            )
            .join("\n")
        )
}
%>
<%-
asdf()
%>
```

output
```
// Generated using Sourcery 2.0.1 — https://github.com/krzysztofzablocki/Sourcery
// DO NOT EDIT

zero = 0 // some comment
positive = 1
negative = -1
```

art-divin:
## Context

When `SyntaxTreeCollector` visits `EnumDeclSyntax`, `EnumDeclSyntax` is being parsed from `EnumCase` extension (file `EnumCase+SwiftSyntax.swift`).

In there there's a logic which extracts the rawValue from the `EnumDeclSyntax` by assigning `equal` sign to `nil`:

    var value = node.rawValue?.withEqual(nil).description.trimmed

## Problem Statement

Using `node.rawValue` does not eliminate contents of trailing trivia:

```swift
po node.rawValue?.withEqual(nil)
...
    ▿ value : IntegerLiteralExprSyntax
      - unexpectedBeforeDigits : nil
      ▿ digits : integerLiteral("0")
        - text : "0"
        ▿ leadingTrivia : []
          - pieces : 0 elements
        ▿ trailingTrivia : [spaces(1), lineComment("// some comment")]
          ▿ pieces : 2 elements
            ▿ 0 : spaces(1)
              - spaces : 1
            ▿ 1 : lineComment("// some comment")
              - lineComment : "// some comment"
        ▿ tokenKind : TokenKind
          - integerLiteral : "0"
...
```

By adding `withTrailingTrivia(.zero).` it is possible to eliminate comments before and after defined rawValue.

## Notes

Resolves #1154

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Modifications to included files of Swift Templates are now detected by hashing instead of using the modification date when invalidating the cache ([#1161](https://github.com/krzysztofzablocki/Sourcery/pull/1161))
 - Fixes incorrectly parsed /r/n newline sequences ([#1165](https://github.com/krzysztofzablocki/Sourcery/issues/1165) and [#1138](https://github.com/krzysztofzablocki/Sourcery/issues/1138))
 - Fixes incorrect parsing of annotations if there are attributes on lines preceeding declaration ([#1141](https://github.com/krzysztofzablocki/Sourcery/issues/1141))
+- Fixes incorrect parsing of trailing inline comments following enum case' rawValue ([#1154](https://github.com/krzysztofzablocki/Sourcery/issues/1154))
 
 ## 2.0.2
 - Fixes incorrectly parsed variable names that include property wrapper and comments, closes #1140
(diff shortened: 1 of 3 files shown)


rocketnik:
💜Thank you

PR #1179, opened 2023-07-04
Support contained types as variable type name

### LINKED ISSUE #1091: Invalid type lookup
Hi!

I ran into a problem in my `.swifttemplate` template. It seems that the `type` of a `Variable` is resolved incorrectly if there is a nested and global type with the same name.

In the following example, the `type` of a `Variable` `a` is resolved to the global type `A` instead of the expected nested type `X.A`.

```
struct X {
  struct A {}
  struct B {}

  let a: A // variable.type?.globalName == "A"
  let b: B // variable.type?.globalName == "X.B"
}

struct A {}
```

[Example (zip)](https://github.com/krzysztofzablocki/Sourcery/files/9491614/example.zip)

> swift-driver version: 1.45.2 Apple Swift version 5.6.1 (swiftlang-5.6.0.323.66 clang-1316.0.20.12)
> Sourcery version 1.8.2

art-divin:
## Context

Given the following declaration:

```swift
struct X {
  struct A {}

  let a: A
}

struct A {}
```

code that is generated annotates variable `a` with type `A`, however it should automatically infer type for this variable as `X.A` based on the contained type.

## Problem Statement

In file `Variable+SwiftSyntax` there's this code which decides what type should be used for the given variable:
```swift
var typeName: TypeName? = node.typeAnnotation.map { TypeName($0.type) } ??
        node.initializer.flatMap { Self.inferType($0.value.description.trimmed) }
```

However, it does not take into account `containedTypes` array of `visitingType` variable.

## Solution

By iterating over all of the `containedTypes` in the `visitingType` and picking the one which name matches the one used in the declaration. 

## Notes

Resolves #1091

This MR does not cover `Method` and `MethodArgument` cases, which have the same issue for arguments and return values. Same should be true for typealiases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fixes incorrect parsing of annotations if there are attributes on lines preceeding declaration ([#1141](https://github.com/krzysztofzablocki/Sourcery/issues/1141))
 - Fixes incorrect parsing of trailing inline comments following enum case' rawValue ([#1154](https://github.com/krzysztofzablocki/Sourcery/issues/1154))
 - Fixes incorrect parsing of multibyte enum case identifiers with associated values
+- Improves type inference when using contained types for variables ([#1091](https://github.com/krzysztofzablocki/Sourcery/issues/1091))
 
 ## 2.0.2
 - Fixes incorrectly parsed variable names that include property wrapper and comments, closes #1140
--- a/SourceryFramework/Sources/Parsing/SwiftSyntax/AST/Variable+SwiftSyntax.swift
+++ b/SourceryFramework/Sources/Parsing/SwiftSyntax/AST/Variable+SwiftSyntax.swift
@@ -60,13 +60,24 @@ extension Variable {
             hadGetter = true
         }
 
-        let isComputed = node.initializer == nil && hadGetter && !(visitingType is SourceryProtocol)
+        let isVisitingTypeSourceryProtocol = visitingType is SourceryProtocol
+        let isComputed = node.initializer == nil && hadGetter && !isVisitingTypeSourceryProtocol
         let isAsync = hadAsync
         let `throws` = hadThrowable
         let isWritable = variableNode.letOrVarKeyword.tokens(viewMode: .fixedUp).contains { $0.tokenKind == .varKeyword } && (!isComputed || hadSetter)
 
-        let typeName = node.typeAnnotation.map { TypeName($0.type) } ??
-          node.initializer.flatMap { Self.inferType($0.value.description.trimmed) }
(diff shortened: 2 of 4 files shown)


art-divin:
This MR breaks generics annotations, needs to be improved prior to merge.

krzysztofzablocki (approved):
I normally would use `resolveVariableTypes` but if this special case handling for containers makes things simple that's fine by me

art-divin:
I have found out that `Composer.uniqueTypesAndFunctions` function does not set `globalName` of variables to `generic` types, which are calculated correctly. 

I think that this MR needs a bit of change in logic, and the actual bug fix would cover all of type resolutions (methods, variables etc.).

Investigating...

art-divin:
🦀 🧽 ... Couple of weeks later ... 🧽 🦀 

I have discovered that `typeName.actualTypeName` is nill-ified somewhere between 
`Composer.swift`'s
```swift
return (
            types: composed.types.sorted { $0.globalName < $1.globalName },
            functions: composed.functions.sorted { $0.name < $1.name },
            typealiases: composed.unresolvedTypealiases.values.sorted(by: { $0.name < $1.name })
        )
```

and `Sourcery.swift`'s

```swift
let result = try Generator.generate(parsingResult.parserResult, types: parsingResult.types, functions: parsingResult.functions, template: template, arguments: self.arguments)
```

That is the reason why generics are missing when template specifies `variable.typeName.actualTypeName?.name`.

I'll fins the place soon, I think it is somewhere where ParserResult is constructed.

art-divin:
Alright, really hard to grasp this issue, not giving up though 💪

At least now I know that:

1) ParserResults are passed into `TemplateContext` 
2) `TemplateContext ` is serialized as Plist and stored in a temporary file encoded
3) `SwiftTemplate` is loaded from stencil file and compiled via `swift` CLI (apparently, this is what takes the most time while running `sourcery`)
4) when `TemplateContext` is deserialized, `actualTypeName` loses resolved absolute types like they were before serialization 🤔

Continuing investigation 🕵🏻

art-divin:
A bit more investigation has shown that `actualTypeName` is not `nil`, and it works as expected.
`SwiftTemplate` which was compiled, but was not replaced when running "example" app I have created to test `sourcery`, for the reason I do not understand. Once removed manually, compiled binary of `SwiftTemplate` started to get updated along with updates in stencil file. Thus, `actualTypeName` was inserted.

This roundtrip; however, helped me out to understand how to fix failing unit tests properly, and that the solution I have proposed is "ok"-ish.

What a journey 🕯️

artyom-stv:
@art-divin Thank you so much! Great work 💪🏻

PR #1182, opened 2023-07-17
Fix relative path of artifactbundle

erichoracek:
The `artifactbundle.info.json.template` references the `sourcery/bin` subdirectory but the binary is not moved to that path so the `artifactbundle` does not work correctly. We need to actually move the file there and then we get the right structure

Artifact bundle support was originally added in #1155 by @InkyTsar

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -364,11 +364,12 @@ namespace :release do
 
   desc 'Create a zip containing all the prebuilt binaries in the artifact bundle format (for SwiftPM Package Plugins)'
   task :artifactbundle => :zip do
-    bundle_dir = "build/sourcery.artifactbundle"
+    bundle_dir = 'build/sourcery.artifactbundle'
+    bin_dir = "#{bundle_dir}/sourcery/bin"
 
     # Copy the built product to an artifact bundle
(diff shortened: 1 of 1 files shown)


erichoracek:
[Here's the docs](https://github.com/apple/swift-evolution/blob/main/proposals/0305-swiftpm-binary-target-improvements.md#artifact-bundle) on the structure of the artifact bundle if it's helpful

krzysztofzablocki:
thanks!

utahwithak:
This still appears to be broken in the 2.1.1 release. I'm seeing the structure be: 
bundle
 - info.json
 - sourcery
   -  bin
      -  sourcery
         - bin 
the info.json is specifying `sourcery/bin/sourcery` which is incorrect. 

It should be pointing directly to the executable which would be `sourcery/bin/sourcery/bin/sourcery` I feel like this is a bit too nested and the whole nested sourcery folder should be brought up so it's just `sourcery/bin/sourcery`

art-divin:
👋🏻  Thank you @utahwithak ,

I'll have a look once I have a minute 👍🏻

PR #1184, opened 2023-07-26
Add SPM config support

jimmya:
This PR adds support for Swift Packages in the config. This allows Sourcery to determine the module name of the scanned sources.

It depends on https://github.com/apple/swift-package-manager which is quite a big dependency (binary size also almost doubled after adding it). It is however the easiest way to read the contents of the package. Other method could be use the JSON output from `swift package dump-package` and parsing that ourselves. The upside to this is that we're not dependent on this big dependency and we're less attached to local Xcode version vs shipped SPM version. Downside is we have to parse stuff ourselves.

I'd love to hear your thoughts on that.

### DIFF
--- a/guides/Usage.md
+++ b/guides/Usage.md
@@ -100,6 +100,22 @@ project:
         - <path to xcframework file>
 ```
 
+You can also provide a Swift Package which will be scanned. Source files will be scanned based on the package's `path` and `exclude` options.
+
+```yaml
+package:
+  path: <path to to the Package.swift root directory>
+  target: <target name>
+```
+Multiple targets:
+```yaml
+package:
+  path: <path to to the Package.swift root directory>
+  target:
+    - <target name>
+    - <target name>
+```
+
 #### Excluding sources or templates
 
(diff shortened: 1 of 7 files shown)


krzysztofzablocki:
This is great, could you add some docs for the new config?

jimmya:
I've added some docs in `Usage.md` and added a changelog entry.

guides/Usage.md
```diff
@@ -100,6 +100,22 @@ project:
         - <path to xcframework file>
 ```
 
+You can also provide a Swift Package which will be scanned. Source files will be scanned based on the package's `path` and `exclude` options.
```
art-divin, line 103: I must say, this would be such a nice feature 👍🏻

krzysztofzablocki:
thanks for contributing 🙇

jimmya:
@krzysztofzablocki  @art-divin I'm sorry to bother you but do you have any plans for a new release in the near future? Having these changes would help me massively.

art-divin:
Thanks for asking, @jimmya 👋 

I am all for the new release, @krzysztofzablocki what do you say? How to proceed?

krzysztofzablocki:
@art-divin there is `rake release` workflow which you can use to make a new release

Issue #244, opened 2017-04-20
SPM installation


abeintopalo:
Hi,

First, I'd like to thank you @krzysztofzablocki for your awesome work. Sourcery is a very useful tool. :)

I'm using Sourcery via SPM. Just updated to 0.6.0 version.
I've built it with ```swift build``` an it built just fine.

My ```Package.swift``` file:

```
import PackageDescription

let package = Package(
    name: "Package",
    dependencies: [
        .Package(url: "https://github.com/krzysztofzablocki/Sourcery.git", Version(0,6,0)),
    ]
)
```

I use the tool in an Xcode build step:

```
"${SRCROOT}/Package/.build/debug/sourcery" --sources "${SRCROOT}/MyProject" --templates "${SRCROOT}/Package/.build/checkouts/Sourcery.git-2310539879416764236/Templates" --output "${SRCROOT}/generated/code"
```

When I build my project I get the following error:

```
Error Domain=NSCocoaErrorDomain Code=260 "The folder “SwiftTemplateRuntime” doesn’t exist." UserInfo={NSFilePath=<MyPath>/Package/.build/debug/SwiftTemplateRuntime, NSUserStringVariant=(
    Folder
), NSUnderlyingError=0x7fba9a43e580 {Error Domain=NSOSStatusErrorDomain Code=-43 "fnfErr: File not found"}}
Command /bin/sh failed with exit code 4
```

Any clue what could be the problem?

Any help is appreciated.

Br,
Attila

ilyapuchka:
Hi @abeintopalo !
Thanks for reporting that. I'm afraid SPM setup was never working with Swift templates. We recently added one template example (Decorator.swifttemplate), so that's why it is now failing. If you point it to your own templates (not swift ofc), it will work.

I think it will be fixed with #240 that requires the same changes as this issue (extracting runtime files into a module)

abeintopalo:
Hi @ilyapuchka,

Thanks for the help. By creating my own template folder and removing Decorator.swifttemplate did the trick.

It is a good workaround for now.

ilyapuchka:
[...]

PR #1188, opened 2023-07-30
Linux Support

### LINKED ISSUE #306: Linux Support
I've tried getting Sourcery working on Linux as a test. SourceKit and SourceKitten can be compiled for Linux, I've already managed that, but the FileWatcher used in Sourcery seems to be the only library that doesn't support Linux. If Sourcery moves to SPM you could fetch all dependencies except the FileWatcher using SPM and compile it. `--watch` wouldn't work, but Sourcery would be working on Linux which is good for Server Side Swift development. Anywhere a feature isn't supported/supportable on Linux you can place a `#if !os(Linux)` or preferably `#if os(macOS) os(iOS) os(watchOS)`.

Just as an added note: I'm not suggesting this as a must for 1.0. But I'd definitely like this to be supported in the future.

art-divin:
Resolves #306 

## Description

This MR enables support on Linux (particularly was compiled under Ubuntu 22.04, but should work on other version as well if all dependencies are met).

## Technical Details

A couple of things were disabled (#1198):

1. Due to the fact that `FileWatcher` relies on `macOS` SDK and uses `FSEvents` framework, it was disabled, i.e. folder watching flag (`--watch`) is not working under Linux
2. All `JavaScriptCore` functionality (`ejs` templates etc.) were disabled under Linux only
3. `NSObject` available under Linux does not conform to KVC, thus tests in `TypedSpec.generated.swift` were moved to class definitions.

## Notes

1. `getVaList` was replaced with recommended `withVaList() {}` (used mostly for `Codable`-related code generation)
2. `NSException` was replaced with `fatalError` only under Linux
4. `CFAbsoluteTimeGetCurrent` was replaced with `Date().timeIntervalSince1970` for all platforms
5. `FolderWatcher` needs to be implemented probably using [approach similar to this one](https://github.com/apple/swift-package-manager/blob/a15833c8a4c172be6cd7f7626cbbbaf94af08de5/Sources/Workspace/ResolvedFileWatcher.swift#L25)
6. `sha256()` of `Data` extension is implemented via [swift-crypto package](https://github.com/apple/swift-crypto)

## Environment Setup

I have installed ubuntu VM through [tart](https://github.com/cirruslabs/tart/issues/62#issuecomment-1225956540) and updated to 22.04 according to [this guide](https://www.linuxtechi.com/upgrade-ubuntu-20-04-to-ubuntu-22-04/).

I had to run the following commands prior to being able to run `bundle install` in Sourcery:

1.  `sudo apt install libffi-dev`
2. `sudo apt install build-essential`
3. `sudo apt install libsqlite3-dev`
4. `sudo apt-get install libncurses5-dev`


art-divin:
I was able to make tests green on macOS while preserving successful compilation under Linux 🔥 

However, `swift test` does not run tests under Linux.

I also need support how to setup CircleCI to run `linux` and also compile `Sourcery` for releases @krzysztofzablocki do you have an insight? Maybe a simple hint, anything would be appreciated 👍🏻 

🙏🏻

krzysztofzablocki:
I don't really have experience with circle but you could also move us to Github actions if you prefer, this project predates them but we could easily use GH actions now

art-divin:
Thanks @krzysztofzablocki . 

Apart from all Linux-related stuff, I have faced an issue when running tests with `swift test`. Seems like bundle is different when `swift test` is running tests (creating `xctest` file) and when Xcode does the same (generates `xctest` file) ->

When `swift test` is executed, the same project generates `SourceryPackageTests.xctest` bundle, which does _not_ contain resources,
while when Xcode runs the same tests, bundle `SourceryLibTests.xctest` is generated, and that bundle contains described resources as per `Package.swift` spec.

Looking into it, but this is closely related to why `.ejs` are disabled for SPM-compiled sourcery (related to #244)

art-divin:
Tried to rename `SourceryLibTests` to `SourceryPackageTests` just to make `swift test` "think" that it's OK to put resources there, and yet it did not help. Here's the difference:

Screenshot 2023-08-08 at 1 13 25 AM

My suspicion is that `swift test` does not support copying resources just yet, but of course, google shows no results... 🤷🏻

art-divin:
My current plan is:

1. [ ] Make `swift test` work
2. [ ] Enable `GitHub Actions` to compile for macOS and Linux
3. [ ] Make tests executable under Linux
4. [ ] Merge this MR

art-divin:
👋🏻  Update:

1. GH actions are working, and Linux build was built
2. macOS is now tested with `swift test`

This means that:

1. Rakefile can be removed once all "lanes" would be implemented with GH actions
2. CircleCI can be disabled after this MR will be merged (and green on both CIs)

art-divin:
I am facing a weird issue on Linux, when trying to compile for tests with `swift test`.

Error: 
```
error: couldn't build /home/tart/Documents/projects/Sourcery/.build/aarch64-unknown-linux-gnu/release/SourceryPackageTests.xctest because of missing inputs: /home/tart/Documents/projects/Sourcery/.build/aarch64-unknown-linux-gnu/release/SourceryPackageTests.build/runner.swift.o
error: fatalError
```

Same issue happens either in `release` or `debug` configuration.

Apart from this, I was only able to enable test discovery under Linux by adding the swift compiler flag [as suggested here](https://forums.swift.org/t/test-discovery-on-linux/26203/22) to `SourceryPackageTests` target (SourceryLibTests in the past) - `swiftSettings: [.unsafeFlags(["-enable-testing"])]`

art-divin:
👋🏻 Update:

1. issue with [missing input file is resolved](https://forums.swift.org/t/bringing-sourcery-to-linux/66735/2).
3. tests are not discovered on Linux still, investigating 🕵🏻

art-divin:
👋🏻 hey there, 

A little update: 

1. Tests are running under both macOS and linux ✅ 
2. UTs are failing under Linux (makes sense huh 🧐)

Once all UTs will be green, this MR will be ready for merging 😮

krzysztofzablocki:
Amazing progress 🙇

art-divin:
👋🏻 A little scoop of the progress so far:

I have found a blocking issue for DryOutputSpec tests under Linux:

when `GenerationWays.stencil` is used for parsing `Base.swift` file in test `supports different ways for code generation` of `DryOutputSpec`, the code generation fails to process 4 types, due to the following code:

```diff
private func resolve(bit: String, context: Any?) -> Any? {
    let context = normalize(context)

    if let context = context as? Context {
      return context[bit]
    } else if let dictionary = context as? [String: Any] {
      return resolve(bit: bit, dictionary: dictionary)
    } else if let array = context as? [Any] {
      return resolve(bit: bit, collection: array)
    } else if let string = context as? String {
      return resolve(bit: bit, collection: string)
    } else if let object = context as? NSObject {  // NSKeyValueCoding
+      #if os(Linux)
+       return nil
+     #else
        if object.responds(to: Selector(bit)) {
          return object.value(forKey: bit)
        }
      #endif
    } else if let value = context as? DynamicMemberLookup {
      return value[dynamicMember: bit]
    } else if let value = context {
      return Mirror(reflecting: value).getValue(for: bit)
    }

    return nil
  }
```

Where `context` is `SourceryRuntime.Types` containing the 4 types from `Base.swift`. 

Due to the fact that KVC is not supported by `swift-corelibs-foundation`, it is not possible to inspect a subclass of `NSObject` through a dynamically constructed selector.

This is related to https://github.com/stencilproject/Stencil/pull/341/files created by @compnerd for windows port of Stencil.

The only idea I have is to use Mirror API to introspect `context` object in that function and get the value dynamically. But I have 5 minute experience with Mirrors in Swift, and 10 years with objc/runtime. So, let's dive 🤿

art-divin:
5 minutes was enough to understand that `Mirror` is a no-go. `DynamicMemberLookup` is the way to go. Can be automated in future with Sorucery to generate memberwise access. 

I'll do it manually for now, and create MR to Stencil to change order in that method, first to check for `DynamicMemberLookup` conformance, and then to NSObjectProtocol.

My next step would be to point Sourcery Package.swift to that MR, since Stencil repository is stale and it looks like it is not supported since a year. I wonder what would it take to become collaborator there as well 🤔

art-divin:
👋🏻  A little update:

1. Created a fork of Stencil and SwiftStencilKit repositories to include new preference logic for `DynamicMemberLookup` over attempting `KVC` first
2. Facing minor issue where there's a missing `\n` in the very end of generated result, meanwhile all the other contents of the generated file for `DryOutputSpec` tests matches between macOS and Linux.

🌧️ 🌧️ 🌧️ 
🌌🛣️🚶🏻

art-divin:
👋🏻  Update:

Almost all tests are green for both platforms. I hope to close this today. 

  🌤️ 
🛣️ 🚶🏻

art-divin:
👋🏻 

I have enabled almost all unit tests, Sourcery is almost ready for an experimental support. 🔥 

Faced another crash running `TemplatesTests`, [reported here](https://github.com/apple/swift/issues/68296#issuecomment-1704336802). I'll do a workaround for this, referencing that bug report.

🌌🌤️     
🛣️🏃🏻‍♂️🐛

art-divin:
👋🏻  Hello @krzysztofzablocki ,

Since I am very close to finishing, which I am to do tonight, I would like to ask you to briefly go through the code and see if there's something really wrong that you won't agree with.

As for what I know is not done correctly is:

1. Code generation for Sourcery codebase - it generates `Equatable.generated` etc. -> which were moved to class definitions due to the fact that under Linux it is not possible to override a method from an extension (`@objc` is unavailable). Even in Darwin variant, it looks odd, though, and I want to address this in a future PR, where I would inject generated code into specific areas of the class definitions like it is now. For that, I need to research template system a little bit more. I see this is done already for Codable methods.
2. Some tests are skipped that should not be - on Linux there's a number of issues with swift compiler. [I have reported](https://github.com/apple/swift/issues/68296#issuecomment-1704336802) one crash from swift compiler which breaks a number of tests. In future, I would remove `#if canImport(ObjectiveC)` piece by piece. As of now, I did maximum to limit the amount of disabled tests down to JavaScriptCore and 3-5 more tests throughout the codebase.
3. I'll add a proper LINUX.md and update README.md, which would describe how to set things up for contributions under Linux - it will be helpful when fixing failing tests on Linux, for example. I am glad it is not hard to do.


Thank you, I'd appreciate any feedback from your side. As for changeset, I won't change much more, I am fixing the only tests left - TemplatesTests. Then, I'll adjust paddings and deliver.

🙏

Sourcery/Sourcery.swift
```diff
@@ -235,41 +240,52 @@ public class Sourcery {
...
+#if canImport(ObjectiveC)
+private extension Sourcery {
+    func templates(from: Paths) throws -> [Template] {
         return try templatePaths(from: from).compactMap {
             if $0.extension == "sourcerytemplate" {
                 let template = try JSONDecoder().decode(SourceryTemplate.self, from: $0.read())
                 switch template.instance.kind {
                 case .ejs:
-                    guard EJSTemplate.ejsPath != nil else {
-                        Log.warning("Skipping template \($0). JavaScript templates require EJS path to be set manually when using Sourcery built with Swift Package Manager. Use `--ejsPath` command line argument to set it.")
```
art-divin, line 245: @krzysztofzablocki here I was really unsure about this guard statements. Due to changes I have introduced in [EJSTemplate.swift](SourceryJS/Sources/EJSTemplate.swift) file in this MR, `swift test` started to work, but for tests, not for a release build which I am unsure how to verify.

Could you please share some insight on this, shall I leave these guard statements as they were, or would it actually work due to this change I have mentioned? 

Thank you 🙏🏻

SourceryRuntime/Sources/AST/AssociatedValue_Linux.swift
```diff
@@ -0,0 +1,148 @@
...
+import Foundation
+// For DynamicMemberLookup we need to import Stencil,
+// however, this is different from SourceryRuntime.content.generated.swift, because
+// it cannot reference Stencil
+import Stencil
+
+/// Defines enum case associated value
+public final class AssociatedValue: NSObject, SourceryModel, AutoDescription, Typed, Annotated, Diffable, DynamicMemberLookup {
+    public subscript(dynamicMember member: String) -> Any? {
+        switch member {
```
art-divin, line 15: This list is not exhaustive, but just to make all unit tests work. It must be autogenerated like `init(coder:)` methods, and I'll address it in #1198

SourceryTests/Generating/StencilTemplateSpec.swift
```diff
@@ -70,12 +70,21 @@ class StencilTemplateSpec: QuickSpec {
...
                     it("doesnt modify the value") {
                         let result = generate("{% for key,value in type.MyClass.variables.2.annotations %}{{ value | toArray }}{% endfor %}")
                         expect(result).to(equal("[Hello, beautiful, World]"))
                     }
                 }
+                #else
+                context("given array") {
+                    it("doesnt modify the value") {
+                        let result = generate("{% for key,value in type.MyClass.variables.2.annotations %}{{ value | toArray }}{% endfor %}")
+                        expect(result).to(equal("[\"Hello\", \"beautiful\", \"World\"]"))
```
art-divin, line 84: On linux, `stringify` from Stencil provides quotes around string values. I could not workaround it, and most likely this is due to differences in how `NSObject` works on two platforms.

SourceryTests/GeneratorSpec.swift
```diff
@@ -256,10 +256,15 @@ class GeneratorSpec: QuickSpec {
...
                     expect(generate("{% for variable in type.Complex.variables %}{{ variable.type.name }}{% endfor %}")).to(equal("FooBar"))
                 }
-
+                #if canImport(ObjectiveC)
                 it("can render variable isOptional") {
                     expect(generate("{{ type.Complex.variables.first.isOptional }}")).to(equal("0"))
                 }
+                #else
+                it("can render variable isOptional") {
+                    expect(generate("{{ type.Complex.variables.first.isOptional }}")).to(equal("false"))
```
art-divin, line 265: 🤷🏻 Don't ask me why 🤷🏻 Really weird difference

SourceryTests/SourcerySpec.swift
```diff
@@ -1271,14 +1294,15 @@ class SourcerySpecTests: QuickSpec {
                         }.toNot(throwError())
                 }
 
+#if canImport(ObjectiveC)
```
art-divin, line 1297: This is related to crash in Swift compiler [reported here](https://github.com/apple/swift/issues/68296).

Templates/Tests/TemplatesTests.swift
```diff
@@ -82,6 +90,12 @@ class TemplatesTests: QuickSpec {
             expect(generatedFileFilteredLines).to(equal(expectedFileFilteredLines))
         }
 
+#if !canImport(ObjectiveC)
```
art-divin, line 93: On Linux, `static func setUp()` is not called, so I needed to manually add this `beforeSuite` to generate files before tests are run.

Templates/Tests/TemplatesTests.swift
```diff
@@ -32,14 +37,18 @@ class TemplatesTests: QuickSpec {
                 print(error)
             }
         }
-
+        #if canImport(ObjectiveC)
+        let contextSources = "\(resources)/Context"
+        #else
+        let contextSources = "\(resources)/Context_Linux"
```
art-divin, line 43: Different file folders are used for two reasons:

1. SPM does not have a method to specify where to put the copied resources (`Resource.copy(path:locatization:)`). 
3. Because of the [crash in Swift compiler](https://github.com/apple/swift/issues/68296), `AutoCodable.swift` file needs to be omitted when tests are run on Linux

Joannis:
AMAZING. Looking forward to seeing this hopefully merged

SourceryFramework/Sources/Parsing/Utils/Time.swift
```diff
@@ -9,5 +9,5 @@ import Foundation
 
 /// Returns current timestamp interval
 public func currentTimestamp() -> TimeInterval {
-    return CFAbsoluteTimeGetCurrent()
+    return Date().timeIntervalSince1970
```
krzysztofzablocki, line 12: could we use currentTimestamp() everywhere we used `Date().timeInterval..` to avoid repetition?

SourceryRuntime/Sources/AST/TypeName/Tuple_Linux.swift
```diff
@@ -0,0 +1,201 @@
...
+// For DynamicMemberLookup we need to import Stencil,
+// however, this is different from SourceryRuntime.content.generated.swift, because
+// it cannot reference Stencil
+import Stencil
+
+/// Describes tuple type
+public final class TupleType: NSObject, SourceryModel, Diffable, DynamicMemberLookup {
+    public subscript(dynamicMember member: String) -> Any? {
+        switch member {
+            case "elements":
```
krzysztofzablocki, line 12: what about name? we def need to codegen those things

krzysztofzablocki (approved):
Some small comments but amazing work 👏

SourceryRuntime/Sources/AST/TypeName/Tuple_Linux.swift
art-divin: Sure, I have mentioned it in #1198 which would take care of it. I also marked this Linux support release as "experimental" (and I'd like to do the same in the next release notes as well), to make it clear that this is not 100% working as of now. But maintaining large MR is cumbersome, I'd like to give some traction now. 

Apart from that, `DynamicMemberLookup` protocol conformance is isolated to non-Darwin platforms. It is done now to make all tests green.

SourceryFramework/Sources/Parsing/Utils/Time.swift
art-divin: What a miss, thank you for noticing! Done ✅

Sourcery/Sourcery.swift
art-divin: Nice one, thanks. I'll revert these changes to throw a more precise error as before.

Newer SPM (I think 5.9) supports "resources embedded into binary executables" (see `.embedInCode`) - maybe we could include EJS template contents into code (encoded, say, with base16/64 encoding or as array of bytes) - I'd need to dig deeper into this, but sounds like some form of convenience for Sourcery. I'll revisit this later.

PR #1213, opened 2023-10-23
Bump SPM version to support Swift 5.9

jimmya:
In order to read Packages when Swift 5.9 is installed we need SPM bundled that supports Swift 5.9.

Note: I'm going to work on a fix to make it more agnostic from the installed Swift version, and less heavy due to not having to bundle SPM with Sourcery. 

Until that time it would be great if we can ship this asap as this is blocking my team at the moment.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -19,12 +19,21 @@
       }
     },
     {
-      "identity" : "komondor",
+      "identity" : "cwlcatchexception",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/shibapm/Komondor.git",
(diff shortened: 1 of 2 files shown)


krzysztofzablocki (approved):

art-divin:
Hey @jimmya , 

thank you for your PR! Nice job 🤝 

Do you need a new release to unblock your team? Thank you

jimmya:
> Hey @jimmya ,
> 
> thank you for your PR! Nice job 🤝
> 
> Do you need a new release to unblock your team? Thank you

@art-divin yes that would be great. We’re depenending on some Xcode 15 stuff so currently we can’t generate without temporarily downgrading

Issue #1218, opened 2023-11-09
Automockable Protocol with a property of type AnyPublisher with an extistential type as Output


pocketal:
I have a protocol which is annotated as AutoMockable.
This protocol has a property of type publisher  `AnyPublisher<any PersonProtocol, Never>?`

This is how the protocol looks like : 

``` swift
protocol PersonProtocol {}

// sourcery: AutoMockable
protocol HouseProtocol {
    var ownerPublisher: AnyPublisher<any PersonProtocol, Never>? { get }
}
```

The generated mock looks like this : 

``` swift
class HouseProtocolMock: HouseProtocol {
    var ownerPublisher: AnyPublisher<(any PersonProtocol, Never>)?
}
```

The parentheses are misplaced

I use the latest version of sourcery version `2.1.2`

I believe there is something wrong with the `AutoMockable.stencil`

paul1893:
This should address the issue: #1220

art-divin:
One we'd push the new release, I'll close this issue and inform OP, so we'd have a confirmation that the issue is resolved. Thank you very much @paul1893 for your contribution 🤝

art-divin:
Closing the issue, please re-open if this is not yet working in 2.1.3

PR #1220, opened 2023-11-09
[Bugfix] - Automockable Protocol with a property of type AnyPublisher

paul1893:
This review to address #1218


paul1893:
A teammate pointed out to me that we should handle `CurrentValueSubject`, `PassthroughSubject`, `GenericType<ExistentialType, ...>`

art-divin:
Thank you very much @paul1893 for this bug fix 🤝🙌🏻

PR #1231, opened 2023-11-29
Reduced counts of collisions for cache srf filenames

### LINKED ISSUE #1230: Broken Cache Logic, as hash of string is not unique
Hello, we're using Sourcery on our project (dozens of difficult templates), so thank you for your hard job

Currently our all generations took up 80 seconds and I'm trying to find how to improve it, checking cache I found that it sometimes it doesn't work for the same 1-5 files in each module.
After research I found that Sourcery use 'hash' for strings, but hash logic for strings is not unique.
(I replaced NDA words with XXXXX)

E.g. 
        let line1 = "/Users/pavel.trafimuk/Developer/Viber_iOS/Classes/Assembly/Modules/ViberXXXXManager/Dependencies/LastXXXXXXNotificationProcessorDependency.swift"
        let line2 = "/Users/pavel.trafimuk/Developer/Viber_iOS/Classes/Assembly/Modules/ViberXXXXManager/Dependencies/NewXXXXXXeNotificationProcessorDependency.swift"

In both cases we receive 5181768815062482058

In results Sourcery doesn't find valid cache (unarchived file shows another modification date) and replace it with another classes. On the next run will be the same

pavel-trafimuk:
Resolves #1230 issue

### DIFF
--- a/Sourcery/Sourcery.swift
+++ b/Sourcery/Sourcery.swift
@@ -429,7 +429,7 @@ extension Sourcery {
         }
 
         let path = parser.path
-        let artifactsPath = cachesPath + "\(path.string.hash).srf"
+        let artifactsPath = cachesPath + "\(path.string.sha256()?.addingPercentEncoding(withAllowedCharacters: .alphanumerics) ?? "\(path.string.hash)").srf"
 
         guard
             artifactsPath.exists,


art-divin:
Hello @pavel-trafimuk ,

thank you very much for your contribution 🤝

PR #1237, opened 2023-12-12
Fixed missing attributes in optional closure type name

### LINKED ISSUE #1004: Closures with @convention(c) attributes lost it in func.name.
It seems like `@convention(c)` is not being retained, here's a simple example:
```swift
struct Test {
    func test(c: (@convention(c) (NSException) -> Void)?) -> Void {
        fatalError()
    }
}
```

If you access the function's `name` property, you will see this:
```
test(c: ((NSException) -> Void)?)
```

I'm not sure if it's due to the closure being in a singular optional tuple?

leonspok:
We've recently encountered an issue where optional closure type with concurrency attributes wasn't properly parsed.

Example:

```
var closure: (@MainActor(unsafe) @Sendable (Int) -> Void)?
```

In Stencil template this type is returned as:

```
variable.typeName.name -> ((Int) -> Void)?
variable.typeName.unwrappedTypeName -> ((Int) -> Void)
```

This PR should fix #1004.


krzysztofzablocki:
🙇

art-divin:
Thank you very much @leonspok 🤝 I'll try to make a new release this weekend

leonspok:
Thanks @art-divin!

PR #1241, opened 2023-12-17
Support for variadic types as method arguments

### LINKED ISSUE #1222: Protocol method with variadic parameter
Hello 👋

The `AutoMockable` stencil does not properly handle methods with variadic parameters.`...` is being omitted from argument's type.

Example:

```swift
protocol SomeProtocol {
    func string(key: String, args: CVarArg...) -> String
}
```

The equivalent method inside the mock looks like this: 

```swift
func string(key: String, args: CVarArg) -> String
```

Thanks.

art-divin:
Resolves #1222 

## Context

Currently it is not possible to mock a method which has a VarArg argument type. Neither it is possible to mock a method or variable, whose return type contains a closure with a VarArg argument.

For example, the following code would not consider `...` after the type:

```swift
func string(key: String, args: CVarArg...) -> String
```

## Solution

Check if the MethodParameter is `variadic`, and if so, append `...` after the `TypeName.name`.


rokridi:
Hi @art-divin 👋 

Is it possible to add `isVariadic` property to `ClosureParameter` type ?

PR #1249, opened 2024-01-04
Added isAsync and throws to Subscript

leonspok:
Added two more fields to Subscript to handle cases when subscript is async or throws. For example,

```
protocol Subscript: AnyObject {
    subscript(arg1: Int) -> Int? { get async }
    subscript(arg2: Int) -> Int? { get throws }
    subscript(arg3: Int) -> Int? { get async throws }
}
```

This now produces:

```
Subscript: ... isAsync = true, `throws` = false ...
Subscript: ... isAsync = false, `throws` = true ...
Subscript: ... isAsync = true, `throws` = true ...
```


art-divin:
👋🏻  Thank you @leonspok for yet another contribution 🤝

PR #1250, opened 2024-01-05
Initialise Subscript's returnTypeName with TypeSyntax, not String

leonspok:
The current implementation doesn't handle optional return types in subscripts well.

For example:

```
protocol Subscript: AnyObject {
    subscript(arg1: Int) -> Int? { get set }
    subscript(arg2: Int) -> Int! { get }
}
```

In this case `subscript.returnTypeName.unwrappedTypeName` will be the same as `subscript.returnTypeName.name`. This should be fixed now.


art-divin (approved):
Looks amazing, thank you @leonspok 🤝

PR #1253, opened 2024-01-09
Update AutoMockable.expected

art-divin:
## Context

[#1252](https://github.com/krzysztofzablocki/Sourcery/issues/1252) introduced support for generic arguments to be mockable after a regression in 2.1.3 release (particularly related to #1240).

This PR updates the entire `AutoMockable.expected` to resolve unit test failures on both Linux and macOS.

## Details

To get currently generated `AutoMockable.expected` to update it, one needs to:

1. put breakpoint in `TemplatesTests.swift` on **line 90** (`expect(generatedFileFilteredLines).to(equal(expectedFileFilteredLines))`)
2. print value of `generatedFilePath`
3. open file located at the printed path
4. copy contents and paste into `AutoMockable.expected` file, removing redundant new lines (there'd be a few).

PR #1261, opened 2024-01-22
Add support for inout parameter for AutoMockable protocols

pocketal:


art-divin:
👋Hello @pocketal and thank you for your MR! 🎉

Could you please check if you have updated files for Linux as well to make CI happy?

See Details in https://github.com/krzysztofzablocki/Sourcery/pull/1253

Templates/Tests/Expected/AutoMockable.expected
```diff
@@ -1459,7 +1484,7 @@ class StaticMethodProtocolMock: StaticMethodProtocol {
 
     static func staticFunction(_ : String) -> String {
         staticFunctionStringStringCallsCount += 1
-        staticFunctionStringStringReceived =
```
art-divin, line 1462: This change needs to be reverted, because the equality check would fail during tests.

Problem occurs if to open this file in Xcode, which truncates trailing whitespace.

art-divin (changes requested):

Templates/Tests/Expected/AutoMockable.expected
pocketal: Fixed

PR #1263, opened 2024-01-23
Fixed generated AutoMockable compilation issue due to generated variable names containing & character.  Added support for existential any for throwable errors.

pocketal:
1. fixed generated AutoMockable compilation issue due to generated variable names containing & character.
2. added support for existential any for throwable errors.

### DIFF
--- a/Templates/Tests/Expected/AutoMockable.expected
+++ b/Templates/Tests/Expected/AutoMockable.expected
@@ -404,19 +404,19 @@ class AnyProtocolMock: AnyProtocol {
 
     //MARK: - z
 
-    var zAnyStubProtocol&CustomStringConvertibleCallsCount = 0
-    var zAnyStubProtocol&CustomStringConvertibleCalled: Bool {
-        return zAnyStubProtocol&CustomStringConvertibleCallsCount > 0
+    var zAnyStubProtocolCustomStringConvertibleCallsCount = 0
+    var zAnyStubProtocolCustomStringConvertibleCalled: Bool {
+        return zAnyStubProtocolCustomStringConvertibleCallsCount > 0
     }
-    var zAnyStubProtocol&CustomStringConvertibleReturnValue: (any StubProtocol & CustomStringConvertible)!
-    var zAnyStubProtocol&CustomStringConvertibleClosure: (() -> any StubProtocol & CustomStringConvertible)?
+    var zAnyStubProtocolCustomStringConvertibleReturnValue: (any StubProtocol & CustomStringConvertible)!
+    var zAnyStubProtocolCustomStringConvertibleClosure: (() -> any StubProtocol & CustomStringConvertible)?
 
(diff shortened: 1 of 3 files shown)


Templates/Tests/Expected/AutoMockable.expected
```diff
@@ -1484,7 +1484,7 @@ class StaticMethodProtocolMock: StaticMethodProtocol {
 
     static func staticFunction(_ : String) -> String {
         staticFunctionStringStringCallsCount += 1
-        staticFunctionStringStringReceived =
```
art-divin, line 1487: This change is redundant, please revert

Templates/Tests/Generated/AutoMockable.generated.swift
```diff
@@ -1090,7 +1090,7 @@ public class ProtocolWithMethodWithInoutParameterMock: ProtocolWithMethodWithIno
         executeParamInoutStringVoidCallsCount += 1
         executeParamInoutStringVoidReceivedParam = param
         executeParamInoutStringVoidReceivedInvocations.append(param)
-        executeParamInoutStringVoidClosure?(&param)
```
art-divin, line 1093: Is this change intended?

art-divin:
Thank you for finding the issue and creating the PR.

Don't forget to update Linux-specific files as well 👍🏻 - details in https://github.com/krzysztofzablocki/Sourcery/pull/1253

Templates/Tests/Generated/AutoMockable.generated.swift
pocketal: It's fixed

Templates/Tests/Expected/AutoMockable.expected
pocketal: Done

art-divin:
There are test failures, both on macOS and Linux, on macOS - `AutoMockable_template__generates_expected_code` - most likely TemplateSpec of some sort has a test which uses files you've modified (cannot be more specific at the moment).
On linux - `QuickSpec.AutoMockable template, generates expected code`. Seems to be the same test which fails.

For information, you could run unit tests locally in Xcode on macOS to see the failure and debug, if needed.

That being said, I see the change:
```
internal extension AutoCasesEnum {
  static let count: Int = 4
  static let allCases: [AutoCasesEnum] = [
    .north,
    .south,
    .east,
    .west
  ]
}
public extension AutoCasesHasAssociatedValuesEnum {
  static let count: Int = 2
}
internal extension AutoCasesOneValueEnum {
  static let count: Int = 1
  static let allCases: [AutoCasesOneValueEnum] = [
    .one
  ]
}
```

which  does not make much sense to be frank. Could you please double check and maybe revert this part?
Aside that, again, thank you for your contributions 👍🏻

pocketal:
It's strange indeed I have the save test result with or without the changes.
I revert

PR #1290, opened 2024-03-05
[Bugfix][AutoMockable.stencil]- inout with multiple parameters & optional any parameter in completion handler

paul1893:
# Context
I noticed 2 bugs in mock generation. If a user have these kind of APIs:
```swift
public protocol ProtocolWithMethodWithInoutParameter {
    func execute(param: inout String, bar: Int) // 👈 `inout` parameter in a function that have more that one parameter.
}

protocol FunctionWithNullableCompletionThatHasNullableAnyParameterProtocol {
    func add(
        _ request: Int,
        withCompletionHandler completionHandler: (((any Error)?) -> Void)? // 👈 a completion with an optional that is annotated with `any` keyword.
    )
}
```
then it will produce:
```swift
public var executeParamInoutStringBarIntVoidReceivedArguments: (param: inout String, bar: Int)? // ❌ 'inout' may only be used on parameters
public var executeParamInoutStringBarIntVoidReceivedInvocations: [(param: inout String, bar: Int)] = [] // ❌ 'inout' may only be used on parameters

...

var addRequestIntWithCompletionHandlerCompletionHandlerAnyErrorVoidVoidClosure: ((Int, (((any Error)?) -> Void))?) -> Void)? // ❌ Does not compile because parenthesis syntax is wrong 

func add(_ request: Int, withCompletionHandler completionHandler: (((any Error)?) -> Void))?) { // ❌ Does not compile because parenthesis syntax is wrong
     addRequestIntWithCompletionHandlerCompletionHandlerAnyErrorVoidVoidCallsCount += 1
     addRequestIntWithCompletionHandlerCompletionHandlerAnyErrorVoidVoidClosure?(request, completionHandler)
}
```
# In this PR
I've updated the AutoMockable.stencil to support these 2 identified cases.

now it will produce:
```swift
public var executeParamInoutStringBarIntVoidReceivedArguments: (param: String, bar: Int)? // ✅
public var executeParamInoutStringBarIntVoidReceivedInvocations: [(param: String, bar: Int)] = [] // ✅

...
var addRequestIntWithCompletionHandlerCompletionHandlerAnyErrorVoidVoidClosure: ((Int, ((((any Error)?) -> Void))?) -> Void)? // ✅

func add(_ request: Int, withCompletionHandler completionHandler: ((((any Error)?) -> Void))?) { // ✅
    addRequestIntWithCompletionHandlerCompletionHandlerAnyErrorVoidVoidCallsCount += 1
    addRequestIntWithCompletionHandlerCompletionHandlerAnyErrorVoidVoidClosure?(request, completionHandler)
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 # Sourcery CHANGELOG
+## 2.1.8
+## Changes
+- Fix: Function with completion as parameter that contains itself an optional any parameter produces wrong mock ([#1290](https://github.com/krzysztofzablocki/Sourcery/pull/1290))
+- Fix: Function with inout parameter when function has more than one parameter produces wrong mock ([#1290](https://github.com/krzysztofzablocki/Sourcery/pull/1290))
+
 ## 2.1.7
 ## Changes
 - Podspec updates - set correct filepath for Sourcery
--- a/Templates/Templates/AutoMockable.stencil
+++ b/Templates/Templates/AutoMockable.stencil
@@ -85,8 +85,8 @@ import {{ import }}
     {% call accessLevel method.accessLevel %}{% call staticSpecifier method %}var {% call swiftifyMethodName method %}Received{% for param in method.parameters %}{{ param.name|upperFirstLetter }}: {{ '(' if param.isClosure }}({% call existentialClosureVariableTypeName param.typeName.unwrappedTypeName param.isVariadic %}{{ ')' if param.isClosure }})?{% endfor %}
     {% call accessLevel method.accessLevel %}{% call staticSpecifier method %}var {% call swiftifyMethodName method %}ReceivedInvocations{% for param in method.parameters %}: [{{ '(' if param.isClosure }}({% call existentialClosureVariableTypeName param.typeName.unwrappedTypeName param.isVariadic %}){{ ')' if param.isClosure }}{%if param.typeName.isOptional%}?{%endif%}]{% endfor %} = []
     {% elif not method.parameters.count == 0 and not hasNonEscapingClosures %}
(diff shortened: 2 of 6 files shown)


art-divin (approved):
Thanks, great find @paul1893 🤝

vakhidbetrakhmadov:
Hi, 

Why the fix is updating the template? Wouldn't the correct solution be to fix the parser or whatever so that `typeName` has correct syntax?

I have a custom template, and i am facing the same error 

```
// sourcery: MyAutoMockable
protocol UserNotificationCenterProtocol {
    func add(
        _ request: UNNotificationRequest,
        withCompletionHandler completionHandler: (((any Error)?, Void) -> Void)?
    )
}
```

```
final class UserNotificationCenterMock: UserNotificationCenterProtocol {
    func add(
        _ request: UNNotificationRequest,
        withCompletionHandler completionHandler: ((any Error?) -> Void)?    // error: optional 'any' type must be written '(any Error)?'
    ) {
        ...
    }
}
```

`{{ method.name }}` value

```
add(_ request: UNNotificationRequest, withCompletionHandler completionHandler: ((any Error?) -> Void)?)
```

`{{ method }}` value

```
<<Method: name = add(_ request: UNNotificationRequest, withCompletionHandler completionHandler: ((any Error?) -> Void)?), selectorName = add(_:withCompletionHandler:), parameters = [MethodParameter: argumentLabel = nil, name = request, typeName = UNNotificationRequest, `inout` = false, isVariadic = false, typeAttributes = [:], defaultValue = nil, annotations = [:], asSource = _ request: UNNotificationRequest, index = 0, MethodParameter: argumentLabel = Optional("withCompletionHandler"), name = completionHandler, typeName = ((any Error?) -> Void)?, `inout` = false, isVariadic = false, typeAttributes = [:], defaultValue = nil, annotations = [:], asSource = withCompletionHandler completionHandler: ((any Error?) -> Void)?, index = 1], returnTypeName = Void, isAsync = false, `throws` = false, `rethrows` = false, accessLevel = internal, isStatic = false, isClass = false, isFailableInitializer = false, annotations = [:], documentation = [], definedInTypeName = Optional(UserNotificationCenterProtocol), attributes = [:], modifiers = [], genericRequirements = [], genericRequirements = []>>
```

PR #1294, opened 2024-03-07
Update SwiftSyntax dependency to 510.0.0

calda:
This PR updates the SwiftSyntax dependency from 508.0.1 to 510.0.0. This should add support for parsing syntax that was added in Swift 5.9 / Swift 5.10.

For example, within Airbnb we've sometimes seen parsing failures in Sourcery when trying to use if / switch expressions. These were added in Swift 5.9 so are not supported by SwiftSyntax 508.0.1. I believe updating to a more recent version of SwiftSyntax should fix this issue.

SwiftSyntax 509.0.0 included a lot of breaking changes. This PR fixes all the breaking changes, so everything compiles. 

As far as I can tell locally, the tests still pass after these changes.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -167,8 +167,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-syntax.git",
       "state" : {
-        "revision" : "2c49d66d34dfd6f8130afdba889de77504b58ec0",
-        "version" : "508.0.1"
+        "revision" : "08a2f0a9a30e0f705f79c9cfaca1f68b71bdc775",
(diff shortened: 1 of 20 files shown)


art-divin:
Hello @calda ,

firstly, welcome to The Contributors club 🤝 🥳 🎈 🎉 

secondly, thank you so much for tackling this! I have had this in mind, but postponed it until after 2.1.8 release. 
And with your tremendous help, now 2.1.8 would include this improvement!

CI passed, so this PR is verified by more than 600 tests, in actuality there are much more _unit tests_ within those 600+ behavioural tests.

Merging 🎉

PR #1063, opened 2022-05-06
Add a flag to parse serially

erichoracek:
Parsing in parallel is leading to unexpected `EXC_BAD_ACCESS`es in our codebase.  By switching to `compactMap`, the crashes go away.

The specific crash is deep in the internals of `SwiftSyntax` mid-parse, which leads me to believe that there's some internal global thread-unsafe state that's being shared between parser instances. I'm not sure if there's any guarantees about thread safety in `SwiftSyntax`, but I couldn't find any. Here's an example (abbreviated) stacktrace:
```
#0	0x00000001001209e4 in protocol witness for SyntaxProtocol._syntaxNode.getter in conformance Syntax ()
#1	0x0000000100121ec3 in SyntaxProtocol.raw.getter at /Users/eric_horacek/Library/Developer/Xcode/DerivedData/Sourcery-hibchsafsxkdcjaxxaylrbeyoydh/SourcePackages/checkouts/swift-syntax/Sources/SwiftSyntax/Syntax.swift:129
#2	0x0000000100b85c32 in SimpleTypeIdentifierSyntax.init(_:) at /Users/eric_horacek/Library/Developer/Xcode/DerivedData/Sourcery-hibchsafsxkdcjaxxaylrbeyoydh/SourcePackages/checkouts/swift-syntax/Sources/SwiftSyntax/gyb_generated/syntax_nodes/SyntaxTypeNodes.swift:68
#3	0x0000000100b89979 in protocol witness for SyntaxProtocol.init(_:) in conformance SimpleTypeIdentifierSyntax ()
#4	0x00000001001acab7 in TypeSyntax.as<τ_0_0>(_:) at /Users/eric_horacek/Library/Developer/Xcode/DerivedData/Sourcery-hibchsafsxkdcjaxxaylrbeyoydh/SourcePackages/checkouts/swift-syntax/Sources/SwiftSyntax/gyb_generated/SyntaxBaseNodes.swift:421
…
#165	0x0000000100566152 in SyntaxVisitor.visitImplCodeBlockItemSyntax(_:) at /Users/eric_horacek/Library/Developer/Xcode/DerivedData/Sourcery-hibchsafsxkdcjaxxaylrbeyoydh/SourcePackages/checkouts/swift-syntax/Sources/SwiftSyntax/gyb_generated/SyntaxVisitor.swift:2584
[...]

PR #1301, opened 2024-03-13
Expand --serialParse flag to also apply to Composer.uniqueTypesAndFunctions

calda:
This PR expands the `--serialParse` flag to also apply to `Composer.uniqueTypesAndFunctions`, so it also performs work serially instead of concurrently.

This is a follow-up to https://github.com/krzysztofzablocki/Sourcery/pull/1063. 

Eric posted that PR to work around a concurrency-related crash we hit in Sourcery when running it on Airbnb's app codebase. Since then, we've internally been using a custom build of Sourcery with that change _plus_ an additional change that updates `Composer.uniqueTypesAndFunctions` to also be serial. I've confirmed that we still occasionally see concurrency-related crashes without that additional change.

The `--serialParse` flag was introduced as a way to work around concurrency-related crashes. Since it is possible to hit concurrency-related crashes in `Composer.uniqueTypesAndFunctions`, it seems reasonable to expand the `--serialParse` option to also disable concurrency within that method. This resolves the crashes we hit internally when trying to adopt Sourcery 2.1.7.

### DIFF
--- a/SourceryRuntime/Sources/Common/TemplateContext.swift
+++ b/SourceryRuntime/Sources/Common/TemplateContext.swift
@@ -48,7 +48,7 @@ public final class TemplateContext: NSObject, SourceryModel, NSCoding, Diffable 
         let fileParserResultCopy: FileParserResult? = nil
 //      fileParserResultCopy = try? NSKeyedUnarchiver.unarchiveTopLevelObjectWithData(NSKeyedArchiver.archivedData(withRootObject: parserResult)) as? FileParserResult
 
-        let composed = Composer.uniqueTypesAndFunctions(parserResult)
+        let composed = Composer.uniqueTypesAndFunctions(parserResult, serial: false)
         self.types = .init(types: composed.types, typealiases: composed.typealiases)
         self.functions = composed.functions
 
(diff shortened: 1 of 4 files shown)


SourceryRuntime/Sources/Common/TemplateContext.swift
```diff
@@ -48,7 +48,7 @@ public final class TemplateContext: NSObject, SourceryModel, NSCoding, Diffable
         let fileParserResultCopy: FileParserResult? = nil
 //      fileParserResultCopy = try? NSKeyedUnarchiver.unarchiveTopLevelObjectWithData(NSKeyedArchiver.archivedData(withRootObject: parserResult)) as? FileParserResult
```
art-divin, line 50: I would like to put a comment here for future reference, that `SwiftSyntax` may misbehave when used from background threads (see a [relevant comment here](https://github.com/krzysztofzablocki/Sourcery/pull/1063#issuecomment-1121337077)), thus `false` is hardcoded.

art-divin (approved):
Amazing work @calda ❤️

calda:
Thanks for the spppeeeeeeedddyyy review @art-divin 🙇🏻

art-divin:
@calda wanna see the real speed 🏎️ - checkout [actions tab](https://github.com/krzysztofzablocki/Sourcery/actions) 🤓

Issue #399, opened 2017-08-22
Option to disable output of version number in generated files header


danielr:
For those people who like committing the generated files to version control (yes, I'm aware of the pros and cons), it would be great to have the option to exclude the version number from the header of the generated files. Otherwise every upgrade of Sourcery leads to meaningless changes in the working copy:

```
-// Generated using Sourcery 0.7.2 — https://github.com/krzysztofzablocki/Sourcery
+// Generated using Sourcery 0.8.0 — https://github.com/krzysztofzablocki/Sourcery
```

I guess this could be either a command line option, or an option in the config file (or both).

I'd be happy to contribute this myself in case it's deemed useful.

Antondomashnev:
Hi @danielr I see your point, but from my perspective the version is something that may help eventually, as a reference I see that CocoaPods, for example, also writes version that generated `Podfile.lock`. 

@ilyapuchka @krzysztofzablocki what do you think about it?

krzysztofzablocki:
I agree with @Antondomashnev, its helpful to have version there

Antondomashnev:
@danielr I think we should keep the version number, and I really think it may help your team eventually 😄

ilyapuchka:
Agree too, having version is valuable.

ilyapuchka:
What we can do though is to provide an option to set custom header (with an option to set a version with a placeholder). It can

dcacenabes:
Hi!
I know this ticket has been closed for many years now, but we're currently struggling with this. We use a cache mechanism to re-use precompiled versions of our modules. The tool computes a SHA based on the contents of the source files within a module. Since Sourcery includes the version in the autogenerated files, it causes the SHA to be different triggering a cache miss.

I understand that versioning can be of use to some people. Is there any chance you reconsider adding an option to make this configurable?

art-divin:
Hello @dcacenabes ,

due to the importance of speeding up CI time and improving developer experience, which was rarely the case in the past,
I agree that we really should disable this part of the header from the generated files. It does not serve purpose these days anyways, as far as for the last year there was not a single bug fix, where this information helped in its resolution.

Let's reopen and quickly fix 👍🏻

dcacenabes:
Hey @art-divin, if you want to go for removing it, I can contribute with a PR myself if it is ok with you :)

art-divin:
Hey 👋

Would be greatly appreciated, in fact, just in time before 2.1.9 release!

[...]

PR #1309, opened 2024-03-25
Remove Sourcery version from header

dcacenabes:
As discussed in issue #399 , including the Sourcery version in the generated files headers has side effects, like higher CI times due to unnecessary cache invalidation.

This PR removes any references to the version as part of the file headers

### DIFF
--- a/Sourcery/Sourcery.swift
+++ b/Sourcery/Sourcery.swift
@@ -622,7 +622,7 @@ extension Sourcery {
         let resultIsEmpty = result.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
         var result = result
         if !resultIsEmpty, outputPath.extension == "swift" {
-            result = self.generationHeader + result
+            result = generationHeader + result
         }
 
         if isDryRun {
--- a/Sourcery/Sourcery.swift
+++ b/Sourcery/Sourcery.swift
@@ -19,7 +19,7 @@ import XcodeProj
 public class Sourcery {
     public static let version: String = SourceryVersion.current.value
     public static let generationMarker: String = "// Generated using Sourcery"
-    public static let generationHeader = "\(Sourcery.generationMarker) — https://github.com/krzysztofzablocki/Sourcery\n"
+    public static let generationHeader = "\(Sourcery.generationMarker) \(Sourcery.version) — https://github.com/krzysztofzablocki/Sourcery\n"
         + "// DO NOT EDIT\n"
 
     enum Error: Swift.Error {
--- a/Sourcery/Sourcery.swift
+++ b/Sourcery/Sourcery.swift
@@ -19,15 +19,7 @@ import XcodeProj
 public class Sourcery {
     public static let version: String = SourceryVersion.current.value
     public static let generationMarker: String = "// Generated using Sourcery"
-    private (set) public lazy var generationHeader = {
-        var prefix = Sourcery.generationMarker
-        if !hideVersionHeader {
-            prefix += " \(Sourcery.version)"
-        }
-        var header = "\(prefix) — https://github.com/krzysztofzablocki/Sourcery\n"
-            + "// DO NOT EDIT\n"
-        return header
-    }()
+    public let generationHeader: String
 
     enum Error: Swift.Error {
         case containsMergeConflictMarkers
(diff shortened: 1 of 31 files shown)


art-divin:
@dcacenabes  Hello 👋🏻 

Thank you very and very much for being initiative and creating this PR so quick!

I wonder, can we have such a strategy to have a runtime flag, which would do what you did here?
I do not want to trigger such a huge diff for all consumers of Sourcery 2.1.9 due to this change, rather, we can provide this feature optionally via a flag, also adding a test covering this option.

What do you think?

dcacenabes:
> @dcacenabes Hello 👋🏻
> 
> Thank you very and very much for being initiative and creating this PR so quick!
> 
> I wonder, can we have such a strategy to have a runtime flag, which would do what you did here? I do not want to trigger such a huge diff for all consumers of Sourcery 2.1.9 due to this change, rather, we can provide this feature optionally via a flag, also adding a test covering this option.
> 
> What do you think?

Makes total sense! It will take me a little bit longer, since I am not familiar with the codebase and I need to understand it a little bit better.

dcacenabes:
Just to make sure we're on the same page, you are suggesting to pass a command line option, something like [force-parse](https://github.com/krzysztofzablocki/Sourcery/blob/7a12709fcf07b4039bdf9e186d17aa9a9ca56274/SourceryExecutable/main.swift#L109) or [cacheBasePath](https://github.com/krzysztofzablocki/Sourcery/blob/7a12709fcf07b4039bdf9e186d17aa9a9ca56274/SourceryExecutable/main.swift#L119) ?

art-divin:
> Just to make sure we're on the same page, you are suggesting to pass a command line option, something like [force-parse](https://github.com/krzysztofzablocki/Sourcery/blob/7a12709fcf07b4039bdf9e186d17aa9a9ca56274/SourceryExecutable/main.swift#L109) or [cacheBasePath](https://github.com/krzysztofzablocki/Sourcery/blob/7a12709fcf07b4039bdf9e186d17aa9a9ca56274/SourceryExecutable/main.swift#L119) ?

Yes, that would be perfect.
To make things easier for you, ask all questions you have, sooner you'll ask, sooner I'll be able to accommodate you with an answer 🤝

Sourcery/Sourcery.swift
```diff
@@ -19,8 +19,15 @@ import XcodeProj
 public class Sourcery {
     public static let version: String = SourceryVersion.current.value
     public static let generationMarker: String = "// Generated using Sourcery"
-    public static let generationHeader = "\(Sourcery.generationMarker) \(Sourcery.version) — https://github.com/krzysztofzablocki/Sourcery\n"
-        + "// DO NOT EDIT\n"
+    private (set) public lazy var generationHeader = {
```
dcacenabes, line 22: I had to change this from static since now it changes based on a flag for Sourcery. Happy to discuss alternative approaches.

dcacenabes:
> > Just to make sure we're on the same page, you are suggesting to pass a command line option, something like [force-parse](https://github.com/krzysztofzablocki/Sourcery/blob/7a12709fcf07b4039bdf9e186d17aa9a9ca56274/SourceryExecutable/main.swift#L109) or [cacheBasePath](https://github.com/krzysztofzablocki/Sourcery/blob/7a12709fcf07b4039bdf9e186d17aa9a9ca56274/SourceryExecutable/main.swift#L119) ?
> 
> Yes, that would be perfect. To make things easier for you, ask all questions you have, sooner you'll ask, sooner I'll be able to accommodate you with an answer 🤝

I pushed some changes.
I also added a test although not sure it is the best `context` to add it.
Happy to apply any feedback you have to the implementation.

art-divin:
> I also added a test although not sure it is the best context to add it.

Oh well, one can never be sure, there are 600 behavioural tests and one can only guess where to put the test. But once you write it, if it becomes too much of a hassle to put it into that specific Spec file, or everything around looks too "unrelated" - then you might want to consider other Spec files for its placement - my approach is like this to this day.

art-divin:
I will merge this, and before the new release, will test with the automation if some regex needs to be adjusted in the scripting.

Thank you very much 🤝  @dcacenabes and welcome to the Contributor club 🥳 🔥

dcacenabes:
> I will merge this, and before the new release, will test with the automation if some regex needs to be adjusted in the scripting.
> 
> Thank you very much 🤝 @dcacenabes and welcome to the Contributor club 🥳 🔥

Very happy to be a contributor! 🎉 Thanks to you for your support and fast response!

visma-ios-if:
@dcacenabes I know this Pr is closed for some time now. But I am not able to figure it out how can I actually make use of this new flag `hideVersionHeader` in my sorcery .yml file.
I tried to use it like this:
```
sources:
  - {path to my sources}
templates:
  - {path to stencil file template}
output:
  {output path}
hideVersionHeader: true // it do not seems to work
```
But it do not matter how many times I am trying to generate the file, the header comments including the version are in there.

art-divin:
Hello @visma-ios-if ,

thank you very much for raising this! 

I've found out that this was not supported, and so I have added this feature in #1328. Please see PR's description for the format.

dcacenabes:
Thanks for taking care of this!

PR #1320, opened 2024-03-30
Improved Logging/Error Handling during SwiftTemplate Processing

art-divin:
## Context

This PR addresses the fact that when executing Sourcery from multiple processes (i.e. running `sourcery` executable from other scripts), mock generation fails if `cacheBasePath` is the same between these concurrent processes.
In actuality, the only reason the `swifttemplate` processing fails is because one process is trying to cache `SwiftTemplate` binary in parallel with another process doing the same for the same SwiftTemplate, i.e. when `cacheKey` is exactly the same between these executables.

## Description
This PR does the following:
- Adds more logs for time measurement in SwiftTemplate
- Adds a note how exactly cacheKey is calculated and its intended usage
- Adds possibility to pass Log configuration when using SourceryLib instead of executable for logging setup


SourcerySwift/Sources/SwiftTemplate.swift
```diff
@@ -186,22 +187,32 @@ open class SwiftTemplate {
...
+                Log.verbose("Reusing built SwiftTemplate binary for SwiftTemplate with cache key: \(hash)...")
+            } else {
+                Log.verbose("Building new SwiftTemplate binary for SwiftTemplate...")
                 try? cachePath.delete() // clear old cache
                 try cachePath.mkdir()
-                try build().move(binaryPath)
+                do {
+                    try build().move(binaryPath)
+                } catch let error as NSError {
+                    if error.domain == "NSCocoaErrorDomain", error.code == 516 {
```
art-divin, line 210: This is what now allows to execute Sourcery from multiple processes in parallel, generating code into different directories when `swifttemplate` is shared across these executions (i.e. cached and reused).

PR #1325, opened 2024-04-30
Fixed typealias resolution breaking resolution of real types.

fabianmuecke:
This PR fixes a regression introduced by https://github.com/krzysztofzablocki/Sourcery/pull/1288. If a type name contained the name of a typealias, part of the type name would be replaced with the resolved typealias, which ultimately led to the lookup of a wrong (in our case non-existent) type.

I simply added an early return, if the exact type name could be found and added a test case to make sure, that both cases work as expected now.

Please let me know, if there is anything I can do, to speed up getting this merged, as this regression has been keeping us from upgrading to the latest Sourcery version for a while (didn't get to having a closer look sooner).

### DIFF
--- a/SourceryRuntime/Sources/Common/Composer/ParserResultsComposed.swift
+++ b/SourceryRuntime/Sources/Common/Composer/ParserResultsComposed.swift
@@ -533,6 +533,10 @@ internal struct ParserResultsComposed {
         // let variable: Module.ID.ID // should be resolved as MyView.ID type
         let finalLookup = typeName.actualTypeName ?? typeName
         var resolvedIdentifier = finalLookup.generic?.name ?? finalLookup.unwrappedTypeName
+        if let type = unique[resolvedIdentifier] {
+            return type
+        }
+        
         for alias in resolvedTypealiases {
             /// iteratively replace all typealiases from the resolvedIdentifier to get to the actual type name requested
             if resolvedIdentifier.contains(alias.value.name), let range = resolvedIdentifier.range(of: alias.value.name) {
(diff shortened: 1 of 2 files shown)


art-divin:
Thank you for the investigation and PR, @fabianmuecke 🤝

fabianmuecke:
Just to keep you posted on this, @art-divin. I just identified another issue, which is probably related and does not get fixed by this PR. We're generating some mock stuff based on generic parameters and the typealias resolution / replacement, which already broke the thing described above also seems to be responsible for this:

We have a typealias named `Value` in some place of our codebase.
We have a function in a protocol like this:
`func foo<Value>(bar: Foo<Value>)`
`typeName` of `parameters.first` is `Foo<Value>` as expected, while `typeName.generic` is `Foo<Element>`, because our totally unrelated typealias of `Value` is pointed to `Element`. I could track this down to that `forEach` block after line `457` in `ParserResultsComposed.swift`, where the type parameter's `typeName` is replaced by the `typeName.actualTypeName`. Right now I'm at a loss how to fix this without breaking the typealias resolution you introduced in that PR linked above, though.

art-divin:
Hey @fabianmuecke ;

thank you very much for your message.

I wonder, does it make sense to merge this PR because it is backed up by a test, and proceed with a fix for the second issue you found?

Because CI checks are green, I see a banafit of having an additional test.
---
Regarding the second issue, Would you be so kind creating an issue with a minimal reproducible with "given vs. expected" input/output?

That would be 50% of thr bug fix, a good minimal reproducible.

_Knowing you have already provided a glympse of it here, it is a bit bureaucracy, yes._

PR #1347, opened 2024-07-15
Method/Initializer parameter types now resolve to the local type if it exists

liamnichols:
# Background

- https://github.com/krzysztofzablocki/Sourcery/pull/1345#discussion_r1678077591

In the latest 2.2.5 release, our templates seemed to have hit a regression where an associated type within an unrelated protocol would show up as the `type` for a `MethodParameter` of an unrelated type. This led me to discover a subtle inconsistency between how types of method parameters are resolved compared to variables.

When resolving a variable type from the syntax tree, it will be sure to check for the type name within the `containedTypes` because this type should take priority over any global type:

https://github.com/krzysztofzablocki/Sourcery/blob/070d0b1f6c839b85a5bf1330a6c3ed49362ef96c/SourceryFramework/Sources/Parsing/SwiftSyntax/AST/Variable%2BSwiftSyntax.swift#L77-L86

This handling was missing for `MethodParameter` which means that the `typeName` (and later `type`) would always be `Foo` instead of `Bar.Foo`.

# Changes

This change addresses the immediate issue of Method Parameter types not being correct by passing the parent `Type` into the `FunctionParameterSyntax` initialiser which is used when walking the syntax tree.

This however isn't quite a complete fix. While looking at the code, I spotted a few other gotchas:

1. ~Return types also aren't resolved with local nested types in mind~
    - I fixed this.. But the same consideration will need to be made for typed throws later on
2. It only works for the immediate parent type and doesn't account for multiple levels of nesting - this flaw also exists with the variable type resolution

But I think that it's a beneficial one for now until some further work can be done to align type resolution. Anyway.. Feedback is welcome 🙏

### DIFF
--- a/SourceryFramework/Sources/Parsing/SwiftSyntax/AST/Signature.swift
+++ b/SourceryFramework/Sources/Parsing/SwiftSyntax/AST/Signature.swift
@@ -15,8 +15,22 @@ public struct Signature {
     public let throwsOrRethrowsKeyword: String?
 
     public init(_ node: FunctionSignatureSyntax, annotationsParser: AnnotationsParser, parent: Type?) {
+        let isVisitingTypeSourceryProtocol = parent is SourceryProtocol
+
+        // NOTE: This matches implementation in Variable+SwiftSyntax.swift
+        // TODO: Walk up the `parent` in the event that there are multiple levels of nested types
+        var returnTypeName = node.returnClause.map { TypeName($0.type) }
+        if !isVisitingTypeSourceryProtocol {
+            // we are in a custom type, which may contain other types
+            // in order to assign correct type to the variable, we need to match
+            // all of the contained types against the variable type
+            if let matchingContainedType = parent?.containedTypes.first(where: { $0.localName == returnTypeName?.name }) {
+                returnTypeName = TypeName(matchingContainedType.name)
+            }
+        }
+
         self.init(parameters: node.parameterClause.parameters,
-                  output: node.returnClause.map { TypeName($0.type) },
+                  output: returnTypeName,
                   asyncKeyword: node.effectSpecifiers?.asyncSpecifier?.text,
                   throwsOrRethrowsKeyword: node.effectSpecifiers?.throwsSpecifier?.description.trimmed,
                   annotationsParser: annotationsParser,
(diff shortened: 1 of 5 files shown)


SourceryFramework/Sources/Parsing/SwiftSyntax/AST/Signature.swift
```diff
@@ -14,20 +14,42 @@ public struct Signature {
...
                   asyncKeyword: node.effectSpecifiers?.asyncSpecifier?.text,
                   throwsOrRethrowsKeyword: node.effectSpecifiers?.throwsSpecifier?.description.trimmed,
-                  annotationsParser: annotationsParser
+                  annotationsParser: annotationsParser,
+                  parent: parent
         )
     }
 
-    public init(parameters: FunctionParameterListSyntax?, output: TypeName?, asyncKeyword: String?, throwsOrRethrowsKeyword: String?, annotationsParser: AnnotationsParser) {
+    public init(
```
art-divin, line 41: 👍🏻

SourceryFramework/Sources/Parsing/SwiftSyntax/AST/Signature.swift
```diff
@@ -14,20 +14,42 @@ public struct Signature {
     /// The `throws` or `rethrows` keyword, if any.
     public let throwsOrRethrowsKeyword: String?
 
-    public init(_ node: FunctionSignatureSyntax, annotationsParser: AnnotationsParser) {
+    public init(_ node: FunctionSignatureSyntax, annotationsParser: AnnotationsParser, parent: Type?) {
+        let isVisitingTypeSourceryProtocol = parent is SourceryProtocol
+
+        // NOTE: This matches implementation in Variable+SwiftSyntax.swift
+        // TODO: Walk up the `parent` in the event that there are multiple levels of nested types
```
art-divin, line 21: Thank you for leaving NOTE and TODO 👍🏻

art-divin (approved):

art-divin:
I'll try to release sooner than later, thank you for your contribution 🤝

liamnichols:
Thanks @art-divin! No rush from my side, we’re fine on 2.2.4 for the time being. 

I just felt like I needed to solve the issue for some closure from my debugging session today 😂

Issue #1349, opened 2024-07-17
Failed to generate code with SwiftTemplate with (at least) custom build path / custom cache path


pavel-trafimuk:
After updating up to 2.2.5 my ymls with swifttemplate start failing.
Sourcery argument: '--config tools/sourcery/generated/phase_02/Settings/SettingsContainer.yml  -v -p --cacheBasePath tools/sourcery/customCache --buildPath tools/sourcery/customSwiftTemplateBuilds/5811391726280627442'

Error:
Generating code...
Rendering template /Users/pavel.trafimuk/Developer/_censored_/tools/sourcery/templates/settings-gen/SettingsContainer.swifttemplate
Reusing built SwiftTemplate binary for SwiftTemplate with cache key: MI6b0Z6LKDJ0eIbHkx7BmAltVogLlLDJK/KJLgPAPLY=...
Binary file location: /Users/pavel.trafimuk/Developer/_censored_/tools/sourcery/customSwiftTemplateBuilds/5811391726280627442/SwiftTemplate/F5970073-49DA-483D-A452-98C9960123B9/2.2.5/MI6b0Z6LKDJ0eIbHkx7BmAltVogLlLDJK%2FKJLgPAPLY%3D
Reusing built SwiftTemplate binary for SwiftTemplate with cache key: MI6b0Z6LKDJ0eIbHkx7BmAltVogLlLDJK/KJLgPAPLY=...
[...]

PR #1350, opened 2024-07-17
Fixed wrong relative path in symbolic link

pavel-trafimuk:
Fixing: https://github.com/krzysztofzablocki/Sourcery/issues/1349

`destinationBinaryPath` is based on cachePath, which could be relative (related to sourcery binary) and we are creating sym link (link at `originalBinaryPath` to `destinationBinaryPath`) which will be also relative, but relative sym link resolves itself not from sourcery binary but from itself.

E.g. 
**sourcery bin**: `~/sourcery`
**destinationBinaryPath**: `/destinationBuilds`
**originalBinaryPath**: `~/sourcery/customCache`

**symlink**: originalBinaryPath-> destinationBinaryPath (`~/sourcery/customCache` will redirect to `/destinationBuilds`),
it resolves to `~/sourcery/customCache/destinationBuilds ` but should be `~/sourcery/destinationBuilds`

### DIFF
--- a/SourcerySwift/Sources/SwiftTemplate.swift
+++ b/SourcerySwift/Sources/SwiftTemplate.swift
@@ -225,6 +225,7 @@ open class SwiftTemplate {
                 try buildDir.mkpath()
             }
             originalBinaryPath = buildDir + hashPath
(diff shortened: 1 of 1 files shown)


pavel-trafimuk:
@art-divin @krzysztofzablocki

PR #1352, opened 2024-07-22
Prevent race condition issues when parsing types

PaulTaykalo:


pavel-trafimuk:
1) maybe move lock inside the function, not outside, as it could happen everywhere in project. 
2) I don't like NSLock cause it's too slow, but I don't know of unfair lock availability on different OSes.

art-divin:
Hey @PaulTaykalo , 

while this is a reasonable fix and workaround, it can hit pretty bad in terms of performance.
I am not saying that it will, but I would avoid a workaround.

Instead, I would implement a different mechanism for storing shared state, or even better - avoid the shared state completely.
Though I do not recall how Composer is invoked or structured to the level of precision allowing do the right choice in the assumption if it's possible or not. Needs an investigation and POC.

But generally speaking ,there's nothing which stops Sourcery from using async/await instead of this custom parallelization mechanism. A little re-write, yes, but modern and robust.

PaulTaykalo:
@krzysztofzablocki Yep, sure. Totally understand. This PR is more like `issue` with a hot-fix.
Feel free to close it.

P.S. Would propably dig deeper, but for some reason Xcode doesn't show `live issues` on the `Sourcery` project, which hugely slows down development process. Are you experiencing this as well? Or Are you using other editors/IDEs
image

Issue #1362, opened 2024-08-27
error: mutable capture of 'inout' parameter 'buffer' is not allowed in concurrently-executing code


mapierce:
## Description
Running Sourcery with Xcode 16 throws out the error:
```
error: mutable capture of 'inout' parameter 'buffer' is not allowed in concurrently-executing code
```
with regards to [this line](https://github.com/krzysztofzablocki/Sourcery/blob/f1974a5f35939112d7952cee52e68beeaf316118/SourceryRuntime/Sources/Common/Array%2BParallel.swift#L20)

mapierce:
Closing as it's fixed with the above referenced PR

PR #1363, opened 2024-08-27
Fixes issue around mutable capture of 'inout' parameter 'buffer' is not allowed in concurrently-executing code

mapierce:
## Description
This PR fixes an error thrown with Xcode 16 around mutable capture of `buffer` in `Array+Parallel`

## Related issues
* https://github.com/krzysztofzablocki/Sourcery/issues/1362

### DIFF
--- a/SourceryRuntime/Sources/Common/Array+Parallel.swift
+++ b/SourceryRuntime/Sources/Common/Array+Parallel.swift
@@ -17,6 +17,7 @@ public extension Array {
     func parallelMap<T>(transform: (Element) -> T) -> [T] {
         var result = ContiguousArray<T?>(repeating: nil, count: count)
         return result.withUnsafeMutableBufferPointer { buffer in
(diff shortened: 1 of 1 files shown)


mapierce:
@art-divin would you be able to give me a hand here? 99% certain tests are failing because CI is running on Xcode 15 and not 16, so it doesn't recognise the `nonisolated` keyword. However, just bumping the version to `16.0.0` also doesn't work, presumable as it's still in beta, so it might need to be something like `16.0.0.beta`, but unsure how to proceed here. All help appreciated.

mapierce:
@art-divin @krzysztofzablocki

art-divin:
Thank you for dealing with this. I am really busy and cannot invest time into even responding to questions 👷

mapierce:
> Thank you for dealing with this. I am really busy and cannot invest time into even responding to questions 👷

No problem @art-divin any chance this could be released soon?

PR #1367, opened 2024-09-13
fix: AutoEquatable Stencil to use `any` for protocols

iDevid:


krzysztofzablocki:
hey, you need to update the generated expectation files too otherwise tests will fail

iDevid:
Hello @krzysztofzablocki, do you have any idea on why tests are failing on linux?

krzysztofzablocki:
@art-divin set-up linux side of things, I don't really know what's off there

PR #1385, opened 2024-11-22
Ability to hide generation header

ilia3546:
This PR adds ability to disable header in generated files

**Problem**

If you use SwiftLint rule "file_name", then you will catch "File Name Violation" errors/warnings for each generated file. Even if you have "swiftlint:disable all" after header.

Also we have swiftlint rule for file header pattern "file_header" which also throws a warning :(

Screenshot 2024-11-22 at 17 46 51

**Solution**

I found a similar problem in the SwiftLint repo - https://github.com/realm/SwiftLint/issues/2277. Users suggest to add this line at the beginning of the file (at 0 line) `// swiftlint:disable:this file_name`.

But at the moment this line contains header comment `// Generated using Sourcery ...`. And I thought that if we be able to hide header and add to template `// swiftlint:disable:this file_name`, then we could solve this problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,7 @@ Inside a project/package that uses this command plugin, right-click the project 
 - `--help` - Display help information
 - `--cacheBasePath` - Base path to the cache directory. Can be overriden by the config file.
 - `--buildPath` - Path to directory used when building from .swifttemplate files. This defaults to system temp directory
+- `--hideHeader` [default: false] - Stop adding the Sourcery header to the generated files.
 - `--hideVersionHeader` [default: false] - Stop adding the Sourcery version to the generated files headers.
 
 ### Configuration file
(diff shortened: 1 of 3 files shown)


krzysztofzablocki:
I don't think this is a good idea, you need to avoid chicken - egg situation so generated code SHOULD NOT be scanned

ilia3546:
> I don't think this is a good idea, you need to avoid chicken - egg situation so generated code SHOULD NOT be scanned

We can add excluded paths to .swiftlint.yml, but then we would see very poor performance of linting :( Example of such issue - https://github.com/realm/SwiftLint/issues/5018

krzysztofzablocki:
you can't have chicken and egg scenario, it will lead to arbitrary failures that will confuse everyone.

What if you add support for additional prefix to headers instead? then you can inject ignore rules

ilia3546:
> you can't have chicken and egg scenario, it will lead to arbitrary failures that will confuse everyone.
> 
> What if you add support for additional prefix to headers instead? then you can inject ignore rules

Perfect, thank you! I've created another PR with this feature. Could you check it please? 
https://github.com/krzysztofzablocki/Sourcery/pull/1389

PR #1389, opened 2024-12-10
Ability to use custom header prefix

ilia3546:
This PR adds ability to add additional prefix before header in generated files

**Problem**

If you use SwiftLint rule "file_name", then you will catch "File Name Violation" errors/warnings for each generated file. Even if you have "swiftlint:disable all" after header.

Also we have swiftlint rule for file header pattern "file_header" which also throws a warning :(

Screenshot 2024-11-22 at 17 46 51

**Solution**

I found a similar problem in the SwiftLint repo - https://github.com/realm/SwiftLint/issues/2277. Users suggest to add this line at the beginning of the file (at 0 line) `// swiftlint:disable:this file_name`.

But at the moment this line contains header comment `// Generated using Sourcery ...`. And I thought that if we be able to add additional custom prefix before the header with the following line `// swiftlint:disable:this file_name`, then we could solve this problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ Inside a project/package that uses this command plugin, right-click the project 
 - `--cacheBasePath` - Base path to the cache directory. Can be overriden by the config file.
 - `--buildPath` - Path to directory used when building from .swifttemplate files. This defaults to system temp directory
 - `--hideVersionHeader` [default: false] - Stop adding the Sourcery version to the generated files headers.
+- `--headerPrefix` - Additional prefix for headers.
 
 ### Configuration file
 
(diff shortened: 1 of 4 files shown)


krzysztofzablocki:
can you update the param names etc? it using old that don't match what it's doing since we are no longer hiding

ilia3546:
> can you update the param names etc? it using old that don't match what it's doing since we are no longer hiding

In this PR I created param `--headerPrefix <custom-prefix>"`. In the old PR it was called `--hideHeader`. I checked this PR carefully for old params and it seems clear.

Do you mean that I should rename param `--headerPrefix` to something else?

krzysztofzablocki:
I marked the old code that doesn't seem should be part of this pr

ilia3546:
Oh, the old parameter `hideVersionHeader` is still unchanged, I just added a comma at the end to be able to add the new parameter :)

krzysztofzablocki:
Oh I misread it, forgot it existed ¯\_(ツ)_/¯

PR #1413, opened 2025-05-08
chore: keep alive properly

krzysztofzablocki:


r3econ:
@krzysztofzablocki I suspect the changes from the PR might be responsible for the errors I see while trying to release

image

I will try to fix it, beginning by reverting the commit, at least so I can release. Let's see how it goes 👷🏾

r3econ:
FYI, reverting the change makes `bundle install && rake release:update_metadata` finish without errors. I'm gonna release 2.2.7 without the breaking commit 🚀 🌔

PR #1420, opened 2025-05-25
Fix AutoMockable Generation for a couple of cases with existential any

musiienko:
### Problem

Let's say we have a protocol:
```swift
protocol AnyProtocolWithOptionals: AutoMockable {
    var a: [any StubProtocol]? { get }
    var b: [Result<Void, any Error>] { get }
    var c: (Int, [(any StubProtocol)?])? { get }
    var d: (Int, (any StubProtocol)?) { get }
    var e: (Int, (any StubProtocol)?)? { get }
    var f: (Int, [any StubProtocol]?)? { get }
    func g(_ g: String, handler: @escaping ([any StubProtocol]?) -> Void) -> Bool
    func h(_ h: String, handler: @escaping ([StubProtocol]) -> Void) -> Bool
    func i(_ i: String, handler: @escaping ([(any StubProtocol)?]) -> Void) -> Bool
    var j: (anyInteger: Int, anyArray: [any StubProtocol]?)? { get }
}
```

Current implementation of `AutoMockable` stencil generates broken Swift Code - too many parentheses or they are inserted in a wrong place. The full generated code can be seen in ee823f0, here is a snippet with the issues mentioned:

```swift
class AnyProtocolWithOptionalsMock: AnyProtocolWithOptionals {


    var a: [(any StubProtocol])? // <-- should  be `[any StubProtocol]?
    var b: [Result<Void, any Error>] = [] // <-- correct
    var c: (Int, [((any StubProtocol))?]))? // <--  should be (Int, [(any StubProtocol)?])?
    var d: (Int, (any StubProtocol)?) { // <-- correct
        get { return underlyingD }
        set(value) { underlyingD = value }
    }
    var underlyingD: ((Int, (any StubProtocol)?))! // <-- correct
    var e: (Int, ((any StubProtocol))?))? // <--  should be (Int, (any StubProtocol)?)?
    var f: (Int, [(any StubProtocol])?))? // <-- should be (Int, [any StubProtocol]?)?
    var j: ((anyInteger: Int, (anyArray: [(any StubProtocol])?))? // <-- should be (anyInteger: Int, anyArray: [any StubProtocol])?)?
    ...
}
```

### Solution

Pretty (maybe too) straightforward:
* skip `any` replacement when type name contains both opening and closing brackets
* in case the type is optional, proceed with `any` replacement only if the type name doesn't contain parenthesis

Looking forward to your feedback!

### DIFF
--- a/Templates/Tests/Generated/AutoMockable.generated.swift
+++ b/Templates/Tests/Generated/AutoMockable.generated.swift
@@ -427,17 +427,17 @@ class AnyProtocolMock: AnyProtocol {
 class AnyProtocolWithOptionalsMock: AnyProtocolWithOptionals {
 
 
-    var a: [(any StubProtocol])?
+    var a: [any StubProtocol]?
     var b: [Result<Void, any Error>] = []
-    var c: (Int, [((any StubProtocol))?]))?
+    var c: (Int, [(any StubProtocol)?])?
     var d: (Int, (any StubProtocol)?) {
         get { return underlyingD }
         set(value) { underlyingD = value }
     }
     var underlyingD: ((Int, (any StubProtocol)?))!
-    var e: (Int, ((any StubProtocol))?))?
-    var f: (Int, [(any StubProtocol])?))?
-    var j: ((anyInteger: Int, (anyArray: [(any StubProtocol])?))?
+    var e: (Int, (any StubProtocol)?)?
+    var f: (Int, [any StubProtocol]?)?
+    var j: (anyInteger: Int, anyArray: [any StubProtocol]?)?
 
 
     //MARK: - g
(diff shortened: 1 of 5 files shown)


Templates/Tests/Generated/AutoMockable.generated.swift
```diff
@@ -423,6 +423,90 @@ class AnyProtocolMock: AnyProtocol {
     }
 
 
+}
+class AnyProtocolWithOptionalsMock: AnyProtocolWithOptionals {
+
+
+    var a: [(any StubProtocol])?
```
musiienko: a closing parenthesis is added at the wrong place. Similar problem is seen in other cases in this class

Templates/Tests/Generated/AutoMockable.generated.swift
```diff
@@ -423,6 +423,90 @@ class AnyProtocolMock: AnyProtocol {
...
+    var b: [Result<Void, any Error>] = []
+    var c: (Int, [((any StubProtocol))?]))?
```
musiienko: `any StubProtocol` has two sets of parentheses + there's one too many at the property definition elvel

PR #1423, opened 2025-06-02
Fix PBXGroup duplication while linking

markmax12:
The issue was previously addressed [here](https://github.com/krzysztofzablocki/Sourcery/pull/1017). However, there we check for either the path property or name property in each individual traversal (there are two of them). This means that if we traverse only looking for the path property in a PBXGroup's child, we would inevitably create duplicate child group if that child only has the name property.

I don't see any need to traverse twice, so in this implementation, I check for both path and name in a single pass.


Ernest0-Production:
@krzysztofzablocki Are there any deadlines for the release date with this fix? I faced the same issue.

PR #1428, opened 2025-07-11
potential fix compiling with Xcode 26

bo2themax:
potential fix for #1425

PR #1434, opened 2025-08-22
Magic fix with SwiftBuild on Xcode 26 Beta 6

bo2themax:
Switched to the commit from the main branch to release/6.2

> To be honest, I’m not sure why the Swift package version affects `ProcessInfo.processInfo.environment`, but this commit appears to work and removes the SwiftBuild dependency.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -264,7 +264,7 @@ var dependencies: [Package.Dependency] = [
 ]
 
 #if compiler(>=6.2)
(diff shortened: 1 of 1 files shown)


awBSH:
thanks @bo2themax 👍 

I can confirm, that this works now with my example from https://github.com/krzysztofzablocki/Sourcery/pull/1428#issuecomment-3213387129

PR #1439, opened 2025-09-16
swift-6.2-RELEASE

bo2themax:


bo2themax:
@krzysztofzablocki I’m not sure how much version needs to be bumped up, but it’s better you do the GitHub release.🚀

krzysztofzablocki:
You can do 2.3.0 @bo2themax so that it doesn't auto update for folks

